Issue #17, opened 2024-06-19
Refactoring: refining process() as API "entrypoint"


truth-quark:
Tasks:

* [ ] Change `args` argument to `kwargs` (simplifies calling `um2nc-standalone` as a module)
* [ ] Specify defaults

PR #33, opened 2024-06-24
Add driving script for automated post-processing during ESM1.5 simulations

### LINKED ISSUE #30: Add driving script for automated post-processing during ESM1.5 simulations
We want to automatically convert ESM1.5's atmospheric output to netcdf format during a simulation ([see here](https://github.com/ACCESS-NRI/access-esm1.5-configs/issues/3)). We can do this using `um2nc-standalone`plus payu's inbuilt userscript functionality, however require a wrapper script which points `um2nc-standalone` to the correct files for conversion. 

One way to do this would be to have a wrapper script which:
1.  Receives a path to the current `output` directory as a command line argument,
2.  Finds all UM output files in the `output/atmosphere` subdirectory,
3.  Runs `um2nc-standalone` on each of the above files.

To start with, we can set the wrapper to use existing `um2netcdf4` module, and swap it with `um2nc-standalone` when it is ready.

blimlim:
Closes #30 

This adds a wrapper script for automatically running the `um2nc` converter during esm1.5 simulations. It takes a single command line argument for the current output directory, finds the UM outputs in the atmosphere subdirectory, and runs the older `um2netcdf4` converter on each of them.

The wrapper is mostly a paired down version of the [one used for CM2 simulations](https://github.com/ACCESS-NRI/access-cm2-drivers/blob/main/src/run_um2netcdf.py).

Once the standalone `um2nc` is ready, we can swap it in. 

Let me know if you have any suggested changes!


src/conversion_driver_esm1p5.py
```diff
@@ -0,0 +1,107 @@
...
+        if re.match(FF_name_pattern, filename)
+    ]
+
+    for FF_name in output_fields_files:
+        FF_file_path = FF_dir / FF_name
+        NC_name = FF_name + ".nc"
+        NC_file_path = NC_dir / NC_name
+
+        # Named tuple to hold the argument list
+        Args = collections.namedtuple(
```
truth-quark, line 58: Do you mind adding a `TODO` here, referencing issue #17?

For the switchover, the replacement `um2nc` should implement `kwargs` & automatically default this arg list.

src/conversion_driver_esm1p5.py
```diff
@@ -0,0 +1,107 @@
...
+        print("Converting file " + FF_name)
+
+        try:
+            um2netcdf4.process(FF_file_path, NC_file_path, args)
+
+        except Exception as exc:
+            # Not ideal here - um2netcdf4 raises generic exception when missing coordinates
+            if exc.args[0] == "Variable can not be processed":
+                warnings.warn("Unable to convert file: " + FF_name)
+                pass
```
truth-quark, line 73: This `pass` isn't required.

truth-quark (commented):
I've done a first pass review & found a few modifications.

We can consider a few unit tests afterwards too. Another change is moving from `src` to the `umpost` package dir once #31 is merged.

truth-quark:
> I've done a first pass review & found a few modifications.
> 
> We can consider a few unit tests afterwards too. Another change is moving from `src` to the `umpost` package dir once #31 is merged.

#31 is merged, so this fix can switch to the new `umpost` package :-)

src/conversion_driver_esm1p5.py
```diff
@@ -0,0 +1,107 @@
...
+ESM1.5 conversion driver
+
+Wrapper script for automated fields file to NetCDF conversion
+during ESM1.5 simulations. Runs conversion module (currently 
+um2netcdf4) on each atmospheric output in a specified directory.
+
+Adapted from Martin Dix's conversion driver for CM2: 
+https://github.com/ACCESS-NRI/access-cm2-drivers/blob/main/src/run_um2netcdf.py
+"""
+
```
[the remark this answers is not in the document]
blimlim: Good catch – should be fixed now. My VS cide doesn't currently. Do you use an extension which highlights formatting issues?

src/conversion_driver_esm1p5.py
```diff
@@ -0,0 +1,107 @@
...
+import os
+import collections
+import um2netcdf4
+import re
+import f90nml
+import warnings
+import argparse
+import errno
+from pathlib import Path
+from collections.abc import Sequence
```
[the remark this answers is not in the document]
blimlim: Good point - I've removed it now.

src/conversion_driver_esm1p5.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def convert_esm1p5_dir(FF_dir, NC_dir, FF_name_pattern):
```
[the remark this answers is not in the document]
blimlim: Good point again – swapped to lower case.

src/conversion_driver_esm1p5.py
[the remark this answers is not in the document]
truth-quark: I'm using PyCharm community edition which constantly runs a live formatting checker. I _think_ VS code has an extension for something similar... ?

umpost/conversion_driver_esm1p5.py
```diff
@@ -0,0 +1,109 @@
...
+    output_fields_files = [
+        filename
+        for filename in os.listdir(ff_dir)
+        if re.match(FF_name_pattern, filename)
+    ]
+
+    for FF_name in output_fields_files:
+        FF_file_path = ff_dir / FF_name
+        NC_name = FF_name + ".nc"
+        NC_file_path = nc_dir / NC_name
```
truth-quark, line 53: This lock is a good candidate for a standalone input/output only function for unit testing.

umpost/conversion_driver_esm1p5.py
```diff
@@ -0,0 +1,109 @@
...
+
+        # TODO: um2netcdf will update the way arguments are fed to `process`.
+        # https://github.com/ACCESS-NRI/um2nc-standalone/issues/17
+        # Update the below arguments once the changes are added.
+
+        # Named tuple to hold the argument list
+        Args = collections.namedtuple(
```
truth-quark, line 60: I didn't notice the `Args` named tuple is defined inside a loop, which raises the question - does it get redefined at every iteration?

umpost/conversion_driver_esm1p5.py
```diff
@@ -0,0 +1,109 @@
...
+            "Args",
+            "nckind compression simple nomask hcrit verbose include_list exclude_list nohist use64bit",
+        )
+        # TODO: Confirm with Martin the below arguments are appropriate defaults.
+        args = Args(3, 4, True, False, 0.5, False, None, None, False, False)
```
truth-quark, line 65: I also realised the `Args` instantiation can also be extracted to a constant outside the convert function.

blimlim: Good idea – have added this in.

blimlim:
I've had a go at making the structure more modular, splitting the code that was previously under `if __name__ == "__main__"` and  `convert_esm1p5_dir` into several smaller functions. There are quite a few changes, so apologies for there being a bit to review. Any ideas for improving the new structure are very welcome, as are any other comments/suggestions!

truth-quark:
> I've had a go at making the structure more modular, splitting the code that was previously under `if __name__ == "__main__"` and `convert_esm1p5_dir` into several smaller functions. There are quite a few changes, so apologies for there being a bit to review. Any ideas for improving the new structure are very welcome, as are any other comments/suggestions!

Nice! This module looks a really good candidate for an initial, contained foray into unit testing.

Do you want to try implementing the test module for the PR?
* If you're using `conda` type tools, we can see how those work for an environment manager
* I default to `pytest` as a test library/test runner https://docs.pytest.org/en/latest/
* Then see if `pytest`'s docs make sense for writing a few initial tests & running them

Is that enough to go on for the next step?

blimlim:
Sounds like a good plan! I'll have a go at putting some tests together.

blimlim:
I've put together some small tests of the different functions. I couldn't think of how to best test the actual conversion functions e.g. `convert_esm1p5_output_dir`, and so have only tested the error handling. I was wondering if it would be best to wait until we swap to the new package to test those functions?

Any feedback and suggestions for the tests would be very welcome!

truth-quark:
> I've put together some small tests of the different functions. I couldn't think of how to best test the actual conversion functions e.g. `convert_esm1p5_output_dir`, and so have only tested the error handling. I was wondering if it would be best to wait until we swap to the new package to test those functions?
> 
> Any feedback and suggestions for the tests would be very welcome!

Testing the conversion function is trickier as it's the file workflow core that ties multiple parts together.  As it calls the other `um2netcdf4` module, we also want to treat that as a boundary, limiting testing here to the driving script.

One approach we can take is reviewing some tests individually, concentrating on design practices for code & tests. Then we can work up to the more complex tests.

If that sounds OK, we can start with the 1st test:
https://github.com/ACCESS-NRI/um2nc-standalone/pull/33/files#diff-275adce7802070a75673ca6a80a5a7bcf54f796f3a8874c6b60e321a1a05c476R8-R15

A few things I spotted are:
* https://github.com/ACCESS-NRI/um2nc-standalone/pull/33/files#diff-275adce7802070a75673ca6a80a5a7bcf54f796f3a8874c6b60e321a1a05c476R8: is a good use of the `tmp_path` fixture, although I found it leaves temp test files behind
* Test setup includes writing a file, reading it to obtain data & returning a single value:
  * There's "buried" I/O in a lower level function (writing & reading the namelist file)
  * `xhist_nml` goes out of scope when the func returns

Ideally, tests avoid I/O & work with simple data as much as possible. For this test, unless there's a need to guard against missing lookups in `xhist_nml` & customise some errors, I'd move these 2 lines:
https://github.com/ACCESS-NRI/um2nc-standalone/pull/33/files#diff-1c8c701fc2ecbb543f5a4261dfacede102fa9f692517fd0c9ea1e6baaf678218R61-R62

... to here:
https://github.com/ACCESS-NRI/um2nc-standalone/pull/33/files#diff-1c8c701fc2ecbb543f5a4261dfacede102fa9f692517fd0c9ea1e6baaf678218R210

Then remove the `get_um_run_id()` func & test (and the I/O & setup vanishes). Testing the workflow can be deferred while sorting out the sub-parts.

If you have 40 mins spare, you might want to watch:
https://www.youtube.com/watch?v=PBQN62oUnN8  (Hoist Your I/O)
This should serve as a useful primer for design/testing & I/O for this block of work.

blimlim:
Sounds good!
I've added in those changes for the `xhist` namelist. Thanks for sending through the video, I'll try and run through that tonight.

In a similar vein, would it be best to modify the [`find_matching_fields_files`](https://github.com/ACCESS-NRI/um2nc-standalone/pull/33/files#diff-1c8c701fc2ecbb543f5a4261dfacede102fa9f692517fd0c9ea1e6baaf678218R87-R110) function so that the testing doesn't involve [writing all those dummy files](https://github.com/ACCESS-NRI/um2nc-standalone/pull/33/files#diff-275adce7802070a75673ca6a80a5a7bcf54f796f3a8874c6b60e321a1a05c476R23-R67)? One idea I had was to change it to accept a list of file paths as an argument, instead of the current `fields_file_dir` argument. Let me know if this would be sensible approach or if you had any other suggestions!

truth-quark:
> In a similar vein, would it be best to modify the [`find_matching_fields_files`](https://github.com/ACCESS-NRI/um2nc-standalone/pull/33/files#diff-1c8c701fc2ecbb543f5a4261dfacede102fa9f692517fd0c9ea1e6baaf678218R87-R110) function so that the testing doesn't involve [writing all those dummy files](https://github.com/ACCESS-NRI/um2nc-standalone/pull/33/files#diff-275adce7802070a75673ca6a80a5a7bcf54f796f3a8874c6b60e321a1a05c476R23-R67)? One idea I had was to change it to accept a list of file paths as an argument, instead of the current `fields_file_dir` argument. Let me know if this would be sensible approach or if you had any other suggestions!

Yep, working with file paths (as `str` or `pathlib.Path`) is the way to go. If you treat a list of file names as the equivalent output of something like `os.listdir()`, one can "ignore" the underlying OS/file system (unless the code under test contains I/O). Ideally, the goal is to split I/O calls from the data they can generate, aiming to work with data as much as we can. Also, having _non-test code_ functions accept basic data structures args allows you to pass in all sorts of different values for testing.

On that, the [dir_contents var](https://github.com/ACCESS-NRI/um2nc-standalone/pull/33/files#diff-275adce7802070a75673ca6a80a5a7bcf54f796f3a8874c6b60e321a1a05c476R27) is great as you've documented an expected file structure *in* the tests. This saves other contributors from having to chase down the info.

blimlim:
Ah that makes sense, thanks! I've modified `find_matching_fields_files` to work with lists.

Attempting to continue with that approach, I've tried changing the `convert_fields_file_dir` function. Currently it takes in a directory path as an argument, searches for the fields files in that directory, and then converts them:

https://github.com/ACCESS-NRI/um2nc-standalone/blob/2e97bef880bb0cabd2062175d77f4306cd430a14/umpost/conversion_driver_esm1p5.py#L112-L147


In a test branch I've swapped it to just take that list of filepaths as an argument:

https://github.com/ACCESS-NRI/um2nc-standalone/blob/416ceb4f90b370fdc5f08e2e0428ab2afc06db13/umpost/conversion_driver_esm1p5.py#L112-L128

and moved that filesystem interaction up into the `convert_esm1p5_output_dir` function:


https://github.com/ACCESS-NRI/um2nc-standalone/blob/416ceb4f90b370fdc5f08e2e0428ab2afc06db13/umpost/conversion_driver_esm1p5.py#L168-L175

Would this be a sensible change to make? Any other suggestions are good too!

truth-quark:
> Attempting to continue with that approach, I've tried changing the `convert_fields_file_dir` function. Currently it takes in a directory path as an argument, searches for the fields files in that directory, and then converts them:
> 
> [snip]
> 
> Would this be a sensible change to make? Any other suggestions are good too!

Yep, that approach works well as you can pass different args into `convert_fields_file_list()` & there's only a single external dependency to `um2netcdf4.process()`. This makes writing the next test simpler.

For simplicity, the test needs to prevent the full `um2netcdf4.process()` from running and substitute some dummy code to return test values in its place. This is handled with the `mock` library: https://docs.python.org/3/library/unittest.mock.html (I recall the I/O hoisting video discusses this briefly)

Do you want to try:
1. Using `mock.patch` to substitute `process()` with a func returning `None`
2. Writing a test for `convert_fields_file_dir()` to check if a typical case runs end to end
3. Expanding the convert fields tests, covering the good cases with sequences of
   - one file path
   - a few file paths
   - all file paths
   - an empty sequence

Afterwards, we can look at using mocks to simulate errors from `process()`.

blimlim:
Sounds good! Would these tests be similar to what you had in mind?

https://github.com/ACCESS-NRI/um2nc-standalone/pull/33/commits/a4f518b1b31b8fdc6b645f2f8764c5d4c83e6411#diff-275adce7802070a75673ca6a80a5a7bcf54f796f3a8874c6b60e321a1a05c476R78-R101


> * an empty sequence

Good point – I guess I should think about what should happen when no fields files are found in the target directory...



> Afterwards, we can look at using mocks to simulate errors from `process()`.

Would this mainly be testing around the error that `convert_fields_file_list` tries to handle:

https://github.com/ACCESS-NRI/um2nc-standalone/blob/a4f518b1b31b8fdc6b645f2f8764c5d4c83e6411/umpost/conversion_driver_esm1p5.py#L138-L146

or other errors that `process` might raise too?

truth-quark:
> Sounds good! Would these tests be similar to what you had in mind?
> 
> [a4f518b#diff-275adce7802070a75673ca6a80a5a7bcf54f796f3a8874c6b60e321a1a05c476R78-R101](https://github.com/ACCESS-NRI/um2nc-standalone/commit/a4f518b1b31b8fdc6b645f2f8764c5d4c83e6411#diff-275adce7802070a75673ca6a80a5a7bcf54f796f3a8874c6b60e321a1a05c476R78-R101)

Yep, those tests cover the basis of typical inputs. This workflow testing is somewhat contrived, but the practices make more sense when applied for algorithmic code.

For the new tests:
* Testing with `None` for either arg in `convert_fields_file_list()` is possible, but probably simpler to skip them here & let things fail as an indicator of buggy or incomplete code. Testing with `None` args can be useful to test if failure(s) occur and is needs to be handle in a specific fashion.
* In `test_convert_fields_file_list_several()`, do you want to try:
  * Fixing a setup bug
  * Restructuring the test slightly to avoid testing with a _magic number_?
* In ` test_convert_fields_file_list_empty()`, can you spot a small bug?

Are you running the tests locally or on `gadi`?

truth-quark:
> > * an empty sequence
> 
> Good point – I guess I should think about what should happen when no fields files are found in the target directory...

Let's defer this part until the tests are in place. The tests will help with an iterative process to add features, adapt tests & refactor.

> > Afterwards, we can look at using mocks to simulate errors from `process()`.
> 
> Would this mainly be testing around the error that `convert_fields_file_list` tries to handle:
> 
> https://github.com/ACCESS-NRI/um2nc-standalone/blob/a4f518b1b31b8fdc6b645f2f8764c5d4c83e6411/umpost/conversion_driver_esm1p5.py#L138-L146
> 
> or other errors that `process` might raise too?

Yep, that's exactly it, ensuring the error handling block is exercised & works as expected. Having tests in place helps with changing code if the design doesn't suit requirements.

Do you want to try writing another `convert_fields_file_list()` test & mock to raise exceptions from `process()` and see how the driver script error handling works?

blimlim:
Great thank you! 

I've put together a couple of tests for the `convert_fields_file_list` exception handling, and modified the original function slightly to make the testing easier, though assuming these tests will change when we swap over to um2nc standalone, let me know if you have any suggestions!

https://github.com/ACCESS-NRI/um2nc-standalone/pull/33/commits/2a33186acaf294481e594df32614ba20cf6cfe8d#diff-275adce7802070a75673ca6a80a5a7bcf54f796f3a8874c6b60e321a1a05c476R101-R124



> * Fixing a setup bug

> * In ` test_convert_fields_file_list_empty()`, can you spot a small bug?

Hmm... I've stared at it way too long but haven't uncovered it yet. I'm wondering if it's an issue related to the context management, which I don't have a good understanding of its nuts and bolts yet. Any pointers would be much appreciated!

> Are you running the tests locally or on `gadi`?

Yeah been running the test on gadi, using the version of pytest that gets loaded in with the `um2netcdf4` module.

test/test_conversion_driver_esm1p5.py
```diff
@@ -77,20 +77,18 @@ def test_find_matching_fields_files():
 
 def test_convert_fields_file_list_single():
     with mock.patch("um2netcdf4.process") as mock_process:
-        mock_process.return_value = None
```
truth-quark: Setting the return value needs to be left in to ensure `None` is returned, otherwise the mock module returns another mock object, "representing" the function call:
```Python
>>> m1 = mock.Mock()
>>> res = m1.process()  # call a mock without setting anything on it
>>> res
<Mock name='mock.process()' id='126979898960448'>
>>>
>>> m1.process2.return_value = 5
>>> res2 = m1.process2()
>>> res2  # should return a value instead of a mock
5
```

test/test_conversion_driver_esm1p5.py
```diff
@@ -77,20 +77,18 @@ def test_find_matching_fields_files():
...
-        esm1p5_convert.convert_fields_file_list(
-            ["fake_file_1", "fake_file_2", "fake_file_3"], "fake_nc_write_dir"
-        )
+        esm1p5_convert.convert_fields_file_list(test_file_list, "fake_nc_write_dir")
 
-        assert mock_process.call_count == 3
+        assert mock_process.call_count == len(test_file_list)
 
 
 def test_convert_fields_file_list_empty():
```
truth-quark, line 94: This one also misses the "just in case" mock return value.

blimlim:
Ah thanks for clearing that up, that makes a lot of sense. I've added those `None` return values in. 

After copying 
```python
with mock.patch("um2netcdf4.process") as mock_process:
        mock_process.return_value = None 
```
several times, I was wondering whether it's better to group the different `test_convert_fields_file_list...` tests together?
I tried the following:
```python
class TestConvertFieldsFileList:
    def setup_method(self):
        self.patcher_process = mock.patch("um2netcdf4.process")
        self.mock_process = self.patcher_process.start()
        self.mock_process.return_value = None

    def teardown_method(self):
        self.patcher_process.stop()

    def test_convert_fields_file_list_single(self):
        esm1p5_convert.convert_fields_file_list(["fake_file"], "fake_nc_write_dir")
        self.mock_process.assert_called_once()

    def test_convert_fields_file_list_several(self):
        test_file_list = ["fake_file_1", "fake_file_2", "fake_file_3"]

        esm1p5_convert.convert_fields_file_list(test_file_list, "fake_nc_write_dir")

        assert self.mock_process.call_count == len(test_file_list)

...
```
but was wondering whether this is preferable, or if it's better in this case to leave all the tests separate?

truth-quark:
> Ah thanks for clearing that up, that makes a lot of sense. I've added those `None` return values in.
> 
> After copying
> 
> ```python
> with mock.patch("um2netcdf4.process") as mock_process:
>         mock_process.return_value = None 
> ```
> 
> several times, I was wondering whether it's better to group the different `test_convert_fields_file_list...` tests together?
> |cut some text|
> but was wondering whether this is preferable, or if it's better in this case to leave all the tests separate?

Sweet, you noticed the repetition.

The setup & teardown in a test class approach is somewhat older & adds some syntactic overhead. As the project uses `PyTest`, repetition should be solvable with [standalone test fixtures](https://docs.pytest.org/en/latest/how-to/fixtures.html).

Do you want to try:
* Seeing if base example fixture docs make sense
* Creating fixtures for `process()` for returning `None` & raising exception(s)
* Refactoring to remove manual mock creation

blimlim:
Sounds good! I've had a go at setting up the mocks using pytest fixtures – let me know if the set up looks sensible.

Initially I had a lot of repetitive fixtures for the different exception tests:
``` Python
@pytest.fixture
def allowed_error_message():
    message = esm1p5_convert.ALLOWED_UM2NC_EXCEPTION_MESSAGES["TIMESERIES_ERROR"]
    return message

@pytest.fixture
def mock_process_allowed_error(mock_process, allowed_error_message):
    mock_process.side_effect = Exception(allowed_error_message)
    yield
    mock_process.side_effect = None 


@pytest.fixture
def generic_error_message():
    message = "Test error"
    return message

@pytest.fixture
def mock_process_generic_error(mock_process, generic_error_message):
    # Set the patch of um2netcdf4.process to raise a non-excepted error
    mock_process.side_effect = Exception(generic_error_message)
    yield
    mock_process.side_effect = None
```

and so attempted to condense them into one fixture following [this example](https://docs.pytest.org/en/latest/how-to/fixtures.html#factories-as-fixtures).

``` python
@pytest.fixture
def mock_process_with_exception(mock_process):
    # Add a generic exception with chosen message to mock_process.
    # Yield function so that tests of different exception messages
    # can make use of the same fixture.
    def _mock_process_with_exception(error_message):
        mock_process.side_effect = Exception(error_message)

    yield _mock_process_with_exception

    mock_process.side_effect = None


...

def test_convert_fields_file_list_generic_error(mock_process_with_exception):
    generic_error_message = "Test error"
    mock_process_with_exception(generic_error_message)
    with pytest.raises(Exception) as exc_info:
        esm1p5_convert.convert_fields_file_list(["fake_file"], "fake_nc_write_dir")

    assert str(exc_info.value) == generic_error_message
```

I'm not sure whether this would be the best way of achieving that though and so any suggestions would be great!

I've also tried condensing the three `test_convert_fields_file_list...` tests into a single one after seeing the parameterization feature.

umpost/conversion_driver_esm1p5.py
```diff
@@ -0,0 +1,201 @@
...
+
+        fields_file_path = (
+            Path(fields_file_path)
+            if isinstance(fields_file_path, str)
+            else fields_file_path
+        )
+
+        nc_write_path = set_nc_write_path(fields_file_path, nc_write_dir)
+
+        print("Converting file " + fields_file_path.name)
```
truth-quark, line 143: This `print` is another bit of "buried" I/O. Is this carried over from earlier versions of the run script?

A possible option is modifying this func to return a pair of sequences, one for successful conversions & ones with errors. The higher level calling function can report success/failure to the user. This means tinkering with the test mocks though...

umpost/conversion_driver_esm1p5.py
```diff
@@ -0,0 +1,201 @@
...
+    run_id = xhist_nml["nlchisto"]["run_id"]
+    fields_file_name_pattern = set_esm1p5_fields_file_pattern(run_id)
+
+    atm_dir_contents = current_atm_output_dir.glob("*")
+
+    fields_file_path_list = find_matching_fields_files(
+        atm_dir_contents, fields_file_name_pattern
+    )
+
+    # TODO: What do we want to happen if no fields files are found?
```
truth-quark, line 187: Should there always be fields files to convert, or are there cases where zero fields files are valid?

truth-quark (commented):
I've left a few comments. We can decide what to include in this PR & anything to defer & I can update the review.

umpost/conversion_driver_esm1p5.py
```diff
@@ -0,0 +1,201 @@
...
+}
+
+
+def set_esm1p5_fields_file_pattern(run_id):
+    """
+    Generate regex pattern for finding current experiment's UM outputs.
+
+    Parameters
+    ----------
+    run_id : 5 character run ID for the current UM simulation.
```
truth-quark, line 51: Is it necessary to ensure the `run_id` is a 5 char string in this func?

umpost/conversion_driver_esm1p5.py
```diff
@@ -0,0 +1,201 @@
...
+    ----------
+    dir_contents : list of file paths, typically contents of a single directory.
+    fields_file_name_pattern : Regex pattern for matching fields file names.
+
+    Returns
+    -------
+    fields_file_paths : subset of dir_contents with names matching fields_file_name_pattern.
+    """
+
+    fields_file_paths = []
```
truth-quark, line 109: This block could be pythonised into a pair of list comprehensions:

```Python
...
filepaths = [Path(filepath) for filepath in dir_contents]
fields_file_paths = [p for path in filepaths if re.match(...)]
...
```
It's more of a style thing which avoids "manual" list ops.

umpost/conversion_driver_esm1p5.py
```diff
@@ -0,0 +1,201 @@
...
+    current_atm_output_dir = current_output_dir / "atmosphere"
+
+    if not (current_atm_output_dir.exists()):
+        raise FileNotFoundError(
+            errno.ENOENT, os.strerror(errno.ENOENT), current_atm_output_dir
+        )
+
+    # Create a directory for writing NetCDF files
+    current_run_nc_dir = current_atm_output_dir / "NetCDF"
+    current_run_nc_dir.mkdir(exist_ok=True)
```
truth-quark, line 174: Another stylistic idea from one of the clean code books. The var name `current_run_nc_dir` could be renamed to describe its intent as the nc output dir, then the code gains a self documenting quality & less comments are required.

umpost/conversion_driver_esm1p5.py
```diff
@@ -0,0 +1,201 @@
...
+
+        except Exception as exc:
+            # Not ideal here - um2netcdf4 raises generic exception when missing coordinates
+            if exc.args[0] in ALLOWED_UM2NC_EXCEPTION_MESSAGES.values():
+                warnings.warn("Unable to convert file: " + fields_file_path.name)
+            else:
+                raise
+
+
+def convert_esm1p5_output_dir(current_output_dir):
```
truth-quark, line 156: I've got linter warning about variable name shadowing with `current_output_dir`.

umpost/conversion_driver_esm1p5.py
```diff
@@ -0,0 +1,201 @@
...
+    for filepath in dir_contents:
+        filepath = Path(filepath)
+
+        if re.match(fields_file_name_pattern, filepath.name):
+            fields_file_paths.append(filepath)
+
+    return fields_file_paths
+
+
+def convert_fields_file_list(fields_file_path_list, nc_write_dir):
```
truth-quark, line 119: A minor style naming thing: `fields_file_path_list` could be shortened to `fields_file_paths` to imply a sequence var. My linter also detected there's a name mismatch between the arg name and the docstring name.

umpost/conversion_driver_esm1p5.py
```diff
@@ -0,0 +1,201 @@
...
+
+    # For ESM1.5 simulations, files start with run_id + 'a' (atmosphere) +
+    # '.' (absolute standard time convention) + 'p' (pp file).
+    # See get_name.F90 in the UM7.3 source code for details.
+    fields_file_name_pattern = rf"^{run_id}a.p[a-z0-9]+$"
+
+    return fields_file_name_pattern
+
+
+def set_nc_write_path(fields_file_path, nc_write_dir):
```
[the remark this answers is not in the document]
blimlim: Good idea, I've renamed it

umpost/conversion_driver_esm1p5.py
blimlim: I think there should generally be fields files for conversion, except in the very unlikely case that someone has switched off all atmospheric outputs. I've added in the following warning if no matching fields files are found:

https://github.com/ACCESS-NRI/um2nc-standalone/pull/33/files#diff-1c8c701fc2ecbb543f5a4261dfacede102fa9f692517fd0c9ea1e6baaf678218R205-R210

umpost/conversion_driver_esm1p5.py
blimlim: Good point. `run_id` produced by the version of the UM used by ESM1.5 should be a 5 character string. I've added in a value error in case something goes wrong here.

umpost/conversion_driver_esm1p5.py
blimlim: Good idea, added this in

umpost/conversion_driver_esm1p5.py
blimlim: I've renamed it to `nc_write_dir` to try and make this clearer

umpost/conversion_driver_esm1p5.py
blimlim: Just renamed it to `esm1p5_output_dir` and left the global one as is.

umpost/conversion_driver_esm1p5.py
blimlim: Good idea, just fixed this up

test/test_conversion_driver_esm1p5.py
```diff
@@ -0,0 +1,139 @@
...
+def mock_process_with_exception(mock_process):
+    # Add a generic exception with chosen message to mock_process.
+    # Yield function so that tests of different exception messages
+    # can make use of the same fixture.
+    def _mock_process_with_exception(error_message):
+        mock_process.side_effect = Exception(error_message)
+
+    yield _mock_process_with_exception
+
+    mock_process.side_effect = None
```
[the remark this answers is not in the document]
blimlim: Just removed this one

umpost/conversion_driver_esm1p5.py
blimlim: I've added a new issue for this in #47

umpost/conversion_driver_esm1p5.py
```diff
@@ -0,0 +1,224 @@
...
+
+# TODO: um2nc standalone will raise more specific exceptions.
+# See https://github.com/ACCESS-NRI/um2nc-standalone/issues/18
+# Improve exception handling here once those changes have been made.
+ALLOWED_UM2NC_EXCEPTION_MESSAGES = {
+    "TIMESERIES_ERROR": "Variable can not be processed",
+}
+
+
+def set_esm1p5_fields_file_pattern(run_id):
```
truth-quark, line 45: Line 45 also has the `set` instead of `get` prefix.

blimlim: Ah good point, just changed this now

truth-quark (approved):
Nice, this one is good to go!

blimlim:
Awesome! Thanks for your help getting it to here :) 

I'll get as second review to get a fresh pair of eyes on it, then will merge if all is ok!

jo-basevi (approved):
Everything looks really neat! Only had a small question re `Path`

umpost/conversion_driver_esm1p5.py
```diff
@@ -0,0 +1,224 @@
...
+    Returns
+    -------
+    None
+    """
+
+    esm1p5_output_dir = (
+        Path(esm1p5_output_dir)
+        if isinstance(esm1p5_output_dir, str)
+        else esm1p5_output_dir
+    )
```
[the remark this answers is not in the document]
blimlim: Good idea, I've added this in and it helps simplify the code. I think we want it to crash if any of the filepaths end up as `None`, and so changing it to what you've suggested helps ensure that happens.

PR #36, opened 2024-06-26
Refactor/extract heaviside functionality & add unit tests

### LINKED ISSUE #34: Heaviside refactor/extraction/testing
This issue is for extracting/testing `heaviside` functions from `process()` (part of https://github.com/ACCESS-NRI/um2nc-standalone/issues/14).

Tasks:

* [X] Move data extraction to function
* [X] Add minimum unit tests
* [X] Extract stash code conversion to int
* [X] Refactor logic into small boolean funcs
* [X] Bonus: better cube lookup from `item_codes`

truth-quark:
Fixes #34.

This change extracts heaviside pressure masking functionality to separate functions. I had to refactor & shrink the old logic to facilitate unit testing (otherwise hacky mocking code was required). I've also added a `cube_lookup` to begin simplifying cube ops in `process()` and making the workflow more pythonic.

Known gaps :
* Magic numbers
* Some vars/funcs need good names (some naming changes are "open" should someone think something)
* `process()` still lacks a test

For the review, any opinions on the restructure, gaps & logic correctness comments are welcome.

### DIFF
--- a/umpost/um2netcdf.py
+++ b/umpost/um2netcdf.py
@@ -307,10 +307,10 @@ def process(infile, outfile, args):
 
     cubes = iris.load(infile)
     cubes.sort(key=lambda cs: cs.attributes['STASH'])
-    cube_index = {to_item_code(c.attributes['STASH']): c for c in cubes}
+    cube_lookup = {to_item_code(c.attributes['STASH']): c for c in cubes}
 
     (need_heaviside_uv, heaviside_uv,
-     need_heaviside_t, heaviside_t) = check_pressure_level_masking(cube_index)
+     need_heaviside_t, heaviside_t) = check_pressure_level_masking(cube_lookup)
 
     do_mask = not args.nomask  # make warning logic more readable
 
@@ -507,13 +507,13 @@ def to_item_code(stash_code):
     return 1000 * stash_code.section + stash_code.item
 
 
-def check_pressure_level_masking(cube_index):
+def check_pressure_level_masking(cube_lookup):
     """
     Examines cubes for heaviside uv/t pressure level masking components.
 
     Parameters
     ----------
-    cube_index : dict style mapping of integer item codes to corresponding iris Cube objects.
+    cube_lookup : dict style mapping of integer item codes to corresponding iris Cube objects.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


jo-basevi (approved):
The changes look good to me

umpost/um2netcdf.py
```diff
@@ -307,36 +307,16 @@ def process(infile, outfile, args):
 
     cubes = iris.load(infile)
     cubes.sort(key=lambda cs: cs.attributes['STASH'])
+    cube_lookup = {to_item_code(c.attributes['STASH']): c for c in cubes}
```
blimlim, line 310: One other way we could do this is to to add item codes directly to the cubes as  attributes, e.g:
```python
for c in cubes:
   c.attributes['item_code'] = to_item_code(c.attributes['STASH'])
```
which can then be accessed later on by `check_pressure_level_masking`

blimlim (approved):
The changes look good! Just noted down the extra option of writing the item codes directly to the cubes in case it simplifies things.

umpost/um2netcdf.py
blimlim: That looks good to me!

truth-quark: The initial dict version is a touch simpler as the key/item code & cube are nicely paired. Moving item code into the `Stash` keeps things together, but it's a tad odd putting attributes into a sub data structure, requiring a custom `DummyCube` for testing.

I'll examine the rest of the script today to inform if one way is neater than another...

truth-quark: > One other way we could do this is to to add item codes directly to the cubes as attributes, e.g:
> ```python
> for c in cubes:
>    c.attributes['item_code'] = to_item_code(c.attributes['STASH'])
> ```
> which can then be accessed later on by `check_pressure_level_masking`

This changes cube _global_ attributes, which are written to _global_ attributes in NetCDFs. This adds a field for the test case, which "breaks" binary file equivalence between the repo's first version & [this version](https://github.com/ACCESS-NRI/um2nc-standalone/commit/055568f9abfc1ef12458d8edd46a079e7121174f).

l'll attempt something else, with a backup being to revert to the `{item_code: cube}` dict style.

blimlim: Ah I hadn't thought about that. So any changes to the attributes end up being written to the NetCDFs? I see why that might not be the ideal place for temporary processing information now.

truth-quark: Yep, all attrs end up in the NC file. As a test, I added another random str key & ran against the test file:
```
 // global attributes:
    :Conventions = "CF-1.6" ;
    :item_code = 5216ll ;
    :item_code_FAKE = "HERE IS FAKE" ;
    :source = "Data from Met Office Unified Model" ;
    :um_version = "7.3" ;
```
The junk attr made it in...

truth-quark: @blimlim how do you find the readability of [adding item_code directly to the object](https://github.com/ACCESS-NRI/um2nc-standalone/commit/9a076ece8ac252ae060dfe139072662c9e678b6f)?

It might look better in a side by side diff view...

blimlim: I think that looks ok, and trust your judgment on whether this or the `cube_lookup` is a better here

truth-quark: With `item_code` out an attr `dict`, the `DummyCube` for testing got simpler too.

Given the code will evolve over time, changes may be needed as more of the utility is picked apart. It's easy enough to change though.

PR #39, opened 2024-07-11
Fix heaviside test regression

truth-quark:
Relates to #34.

The original solution added custom records to `cube` _global attributes_. These were written to `NetCDF` _global attributes_, "breaking" the compatibility of testing outputs. This solution introduces a slight hack, dynamically adding `item_code` as an attribute of each `cube` instead of in the `attributes` dict.

I've manually confirmed running `um2nc` no longer includes additional global vars.

### DIFF
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -1,4 +1,5 @@
 import unittest.mock as mock
+from dataclasses import dataclass
 
 import umpost.um2netcdf as um2nc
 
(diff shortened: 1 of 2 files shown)


umpost/um2netcdf.py
```diff
@@ -345,9 +342,9 @@ def process(infile, outfile, args):
             stashcode = c.attributes['STASH']
             itemcode = to_item_code(stashcode)
```
blimlim, line 343: Would it be worth swapping this out for the new `item_code` attribute here, or is it something for further down the line?

blimlim (approved):
Changes look good to me! Just noted down a couple of small questions, though they aren't important to the code's functionality.

test/test_um2netcdf.py
```diff
@@ -100,20 +159,16 @@ def test_check_pressure_level_masking_need_heaviside_uv():
...
     assert need_heaviside_uv
     assert heaviside_uv is None
 
 
-def test_check_pressure_level_masking_need_heaviside_t():
-    ta_plev_cube = DummyCube({um2nc.ITEM_CODE: 30294})
-    heaviside_t_cube = DummyCube({um2nc.ITEM_CODE: 30304})
-
+def test_check_pressure_level_masking_need_heaviside_t(ta_plev_cube):
+    heaviside_t_cube = DummyCube(30304)
```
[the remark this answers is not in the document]
truth-quark: I'll leave `heaviside_t_cube` as is for now & convert to a fixture when it's needed for `apply_mask` tests.

umpost/um2netcdf.py
truth-quark: Yep, I've fixed it.

Issue #27, opened 2024-06-21
Future neatening/readability


truth-quark:
Future tasks:
* [ ] Rename `nonsnakecase` variables to `now_snake_case`
* [ ] Convert old style `%` formatting to f-strings
* [ ] Fix/reformat long strings
* [ ] Refactor repeated inline strings to "constants"
* [ ] Refactor magic numbers to named "constants"

These can likely be addressed during the unit test retrof/refactoring step, on a function by function basis.

PR #41, opened 2024-07-12
Refactor cube filtering

### LINKED ISSUE #38: Refactoring/test retrofit: cube filtering
Tasks:
* [x] Move cube filtering to function
* [x] Add unit tests
* [x] Fix cmd line interface with mutually exclusive block for include/exclude args

truth-quark:
Closes #38.

This smallish PR extracts cube include/exclude functionality from `process()` to a generator to make the workflow more modular. Tests are included to verify the include/exclude logic. As I understand, `include` and `exclude` are mutually exclusive, with users selecting (a subset of cubes to include, discarding all others) or (a subset of cubes to exclude, keeping all others).

Known gaps :
* `process()` still lacks a test
* This doesn't shrink `process()` much, but it's a small step towards the end product


test/test_um2netcdf.py
```diff
@@ -186,3 +186,34 @@ def test_check_pressure_level_masking_missing_heaviside_t(ta_plev_cube):
...
     assert heaviside_t is None
+
+
+# cube filtering tests
+# use wrap results in tuples to capture generator output in sequence
+
+def test_cube_filtering_mutually_exclusive():
+    include = [30201]
+    exclude = [30293]
+    cubes = (ua_plev_cube, heaviside_uv_cube)
```
blimlim, line 197: Just checking - does this test need to request the  `ua_plev_cube` and `heaviside_uv_cube` fixtures to work properly?

blimlim: Or actually does it not matter too much here as the `filtered_cubes` function raises the value error before doing anything with the cubes.

blimlim (approved):
Change look good to me! Just had a question about one of the tests where I got a bit confused about how it was working.

test/test_um2netcdf.py
truth-quark: I've fixed it for completeness.

umpost/um2netcdf.py
```diff
@@ -603,22 +594,58 @@ def check_pressure_warnings(need_heaviside_uv, heaviside_uv, need_heaviside_t, h
...
+    Parameters
+    ----------
+    cubes : Sequence of Iris Cube objects
+    include: Sequence of item_code (int) to include (discarding all others)
+    exclude: Sequence of item_code (int) to exclude (keeping all other cubes)
+    """
+    if include and exclude:
+        msg = "Include and exclude lists are mutually exclusive"
+        raise ValueError(msg)
+
```
utkarshgupta95, line 613: I am not sure abt this condition. Are you checking if the include and exclude list should not have anything in common? In that case convert list to set and compare with bitwise AND

umpost/um2netcdf.py
```diff
@@ -603,22 +594,58 @@ def check_pressure_warnings(need_heaviside_uv, heaviside_uv, need_heaviside_t, h
...
-                        nargs='+', help='List of stash codes to exclude')
+                        action='count', default=0,
+                        help='verbose output (-vv for extra verbose)')
+
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--include', dest='include_list', type=int,
+                       nargs='+', help='List of stash codes to include')
+    group.add_argument('--exclude', dest='exclude_list', type=int,
+                       nargs='+', help='List of stash codes to exclude')
+
```
utkarshgupta95, line 648: Given the growing size of the file and the numerous arguments, it might be beneficial to move the CLI arguments to a separate file in the future. This approach aligns with the separation of concerns principle, making the file easier to maintain. Additionally, this would be particularly useful if the library is also invoked by other Python programs, not just from the CLI.

test/test_um2netcdf.py
```diff
@@ -186,3 +186,34 @@ def test_check_pressure_level_masking_missing_heaviside_t(ta_plev_cube):
...
+# use wrap results in tuples to capture generator output in sequence
+
+def test_cube_filtering_mutually_exclusive(ua_plev_cube, heaviside_uv_cube):
+    include = [30201]
+    exclude = [30293]
+    cubes = (ua_plev_cube, heaviside_uv_cube)
+
+    with pytest.raises(ValueError):
+        tuple(um2nc.filtered_cubes(cubes, include, exclude))
+
```
utkarshgupta95, line 201: Do you wanna add assert here?

test/test_um2netcdf.py
utkarshgupta95, line 1: Consider adding a test for mutually inexclusive include and exclude lists for better test coverage

umpost/um2netcdf.py
truth-quark: This is from the original utility. Users can select variables/cubes by name. The original CLI specifies that a user can supply a list of var names to include XOR a list of names to exclude. As far as I can tell, the original code tried to avoid items in both lists.

I'm going to leave the interface as is as the main goal of step 1 is adding unit tests/restructuring.

umpost/um2netcdf.py
truth-quark: I've added that idea in #27.

test/test_um2netcdf.py
truth-quark: The assert is hidden in `pytest.raises()` :-)

test/test_um2netcdf.py
truth-quark: I'll skip this as the lists are intended to be exclusive (use one XOR the other).

utkarshgupta95 (approved):

PR #43, opened 2024-07-12
Extract history function

### LINKED ISSUE #42: Extract history to function
Tasks:
* [x] Extract file history attribute to function
* ~[ ] Move set history to end of `process()` to group `sman` I/O operations (for future fixes)~
* ~[ ] Move set CF conventions to end of `process()`~

Unit tests can likely be skipped as the code is very small.

truth-quark:
Closes #42.

This PR extracts & updates the file history attribute. String formatting has been updated with modern Python, and the hard-coded filename has been removed.

Known problems:
* The new version number needs to be determined (the method should ideally be decided in this PR).

### DIFF
--- a/umpost/um2netcdf.py
+++ b/umpost/um2netcdf.py
@@ -10,9 +10,11 @@ Note that um2netcdf depends on the following data access libraries:
 * Iris https://github.com/SciTools/iris
 """
 
+import os
 import sys
 import argparse
 import datetime
+import warnings
 
 from umpost import stashvar_cmip6 as stashvar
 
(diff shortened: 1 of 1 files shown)


umpost/um2netcdf.py
```diff
@@ -603,6 +604,16 @@ def check_pressure_warnings(need_heaviside_uv, heaviside_uv, need_heaviside_t, h
               "These fields will be skipped")
 
 
+def add_global_history(infile, iris_out):
+    version = -1  # FIXME: determine version
```
blimlim, line 608: Once the initial version number is decided, would it be moved to a constant to make it easier to remember to change when there are new updates?

blimlim (approved):
Changes look good to me!

umpost/um2netcdf.py
truth-quark: I suspect we'll use a versioning utility to auto-handle this (I'll add an issue now).

utkarshgupta95 (approved):
LGTM

PR #48, opened 2024-07-17
40/Refactor masking logic / reuse heaviside funcs

### LINKED ISSUE #40: Additional use of `require_heaviside_uv` and `require_heaviside_t` functions
I think this carries over from the original version of the code. In the `process` function, there are a couple of conditionals determining whether to apply the masks to a cube:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/055568f9abfc1ef12458d8edd46a079e7121174f/umpost/um2netcdf.py#L419-L420

https://github.com/ACCESS-NRI/um2nc-standalone/blob/055568f9abfc1ef12458d8edd46a079e7121174f/umpost/um2netcdf.py#L427

It looks like the second parts of the conditions are identical to the ones in the `require_heaviside_uv` and `require_heaviside_t` functions:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/055568f9abfc1ef12458d8edd46a079e7121174f/umpost/um2netcdf.py#L561-L563

https://github.com/ACCESS-NRI/um2nc-standalone/blob/055568f9abfc1ef12458d8edd46a079e7121174f/umpost/um2netcdf.py#L571-L573

Should we swap the conditions in `process` to use these functions?

truth-quark:
Closes #40.

This PR replaces the old stash code and item number checks with the pre-existing helper functions. I've also cleaned reduced some duplication of boolean checks. The unit tests pass & manual file generation matches the NC output from the original script from the 1st commit in this repo.

This PR is mainly to have others verify if the logic is correct :-)

### DIFF
--- a/umpost/um2netcdf.py
+++ b/umpost/um2netcdf.py
@@ -404,20 +404,19 @@ def process(infile, outfile, args):
 
             fix_level_coord(c, z_rho, z_theta)
 
-            if not args.nomask and stashcode.section == 30 and \
-                    (201 <= stashcode.item <= 288 or 302 <= stashcode.item <= 303):
+            if do_mask:
                 # Pressure level data should be masked
-                if heaviside_uv:
-                    apply_mask(c, heaviside_uv, args.hcrit)  # noqa TODO: fix referencing warning
-                else:
-                    continue
+                if require_heaviside_uv(c.item_code):
+                    if heaviside_uv:
+                        apply_mask(c, heaviside_uv, args.hcrit)
+                    else:
+                        continue
 
-            if not args.nomask and stashcode.section == 30 and (293 <= stashcode.item <= 298):
-                # Pressure level data should be masked
-                if heaviside_t:
-                    apply_mask(c, heaviside_t, args.hcrit)  # noqa TODO: fix referencing warning
-                else:
-                    continue
+                if require_heaviside_t(c.item_code):
+                    if heaviside_t:
+                        apply_mask(c, heaviside_t, args.hcrit)
+                    else:
+                        continue
 
             if args.verbose:
                 print(c.name(), c.item_code)


blimlim (approved):
Changes look good to me!

truth-quark:
Related to this work is refactoring the masking logic across `process()`, such as this initial experiment to clean up cube filtering and the logic for selecting masking:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/40/refactor-masking-filtering/umpost/um2netcdf.py

I've commented on the logic part in source issue #40 :
https://github.com/ACCESS-NRI/um2nc-standalone/issues/40#issuecomment-2235163183

Thinking how to address this, there's two main options:
* Expand issue #40's tasks & include logic refactoring here
* Raise a new issue

@blimlim & @jo-basevi do you have any thoughts on these options?

blimlim:
I think a separate issue would work well, but don't have a very strong preference!

jo-basevi:
I think either works. The same issue is good as it has the discussion already there, but sounds like the implementation has a bit more a wider refactoring with filtering and logic for selecting masking so a separate issue also makes sense? 

I've tried to follow the logic in the issue and don't fully understand it yet, but generally it sounds a good plan to filter non-maskable cubes early on and having additional specific warnings.

truth-quark:
For a few thoughts:
* I'm aiming to keep PRs to 50 changed lines or less & focused on one "topic"
* The logic in `process()` should be tested, but the function is still unwieldy at ~90 lines of code
* Testing `process()` now probably requires fiddly mock setup, which incurs repeat changes every time code blocks are extracted...

I'll migrate the masking logic & discussion blurb to another issue. Fixing/testing the masking logic will be easier, e.g. once the cube renaming code blocks are extracted.

truth-quark:
> I'll migrate the masking logic & discussion blurb to another issue.

This is now in issue #50.

jo-basevi (approved):
Sorry for the late review, changes look good to me

Issue #14, opened 2024-06-19
Refactoring: split process() into subcomponents


truth-quark:
Background: `process()` is ~140 SLOC, covering multiple processing steps. These should be split into functions to facilitate unit testing & future expansion or workflow modification.

Tasks:

* [ ] Expand args (results in large method signature, then split into a workflow steps)
* [ ] Extract to individual functions
  * [ ] Mule fields files block —> func
  * [ ] Heaviside uv & t block
  * [ ] Cube filtering, can this be done at the start/ignore processing certain cubes?
  * [ ] cube renaming
  * [ ] Fix/modify units
  * [ ] xconv fix
  * [ ] iris naming fixes
  * [ ] heaviside_uv masking
  * [ ] heaviside_t masking
* [ ] Add unit tests for individual components
* [ ] Refactor I/O step
* [ ] Refactor cmd line args to function signature args
* [ ] Convert magic numbers to constants
* [ ] Replace `print()`s with `warnings.warn()`

truth-quark:
Adding link to related ESM 1.5 project issue:
https://github.com/ACCESS-NRI/access-esm1.5-configs/issues/3

truth-quark:
@blimlim & @jo-basevi  I've updated this high level issue to better explain the work behind `process()`.

Does it make sense & is anything missing?

PR #52, opened 2024-07-24
Extract cube var renaming code to functions

### LINKED ISSUE #51: Extract cube renaming code from `process()`
The `process()` function contains several code blocks for renaming cube vars. These are interspersed with code blocks to fix units & cell methods. There's a few refactoring & cleaning tasks:

* [x] Refactor this block to 1 or more functions:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/00f4ee1de7b0fe42bff29cbcf67494c78c8398f6/umpost/um2netcdf.py#L343-L369
* [x] Move these blocks into functions OR merge into other functions (e.g. `c.standard_name` changes, since the logic might cross over)
https://github.com/ACCESS-NRI/um2nc-standalone/blob/00f4ee1de7b0fe42bff29cbcf67494c78c8398f6/umpost/um2netcdf.py#L382-L393
* [x] ~~Determine how related `fix_level_coords()` is for renaming & gather related functions.~~ (see #53)
* [x]  Is there duplicate or naturally exclusive logic that could be cleaned across the renaming functionality?
* [x] Add docstrings

Problems:
* Unit testing some blocks isn't useful as fields are simply being renamed
* Can the renaming logic be tested?

truth-quark:
Closes #51. Also related to #14.

Several elements are included in this PR (which is larger than I'd prefer):
* Extracting cube renaming code to funcs
* Adding multiple test fixtures and unit tests
* Reordering some naming logic (this didn't break testing on my device)

Overall, this moves ~37 lines of code out from `process()`.

Gaps:
* `process()` still lacks a test, but is now only ~1 page of code
* Some mocks may not represent real object config, but I've tweaked variables to skip code segments & force specific code segments to execute
* A few code branches aren't covered by the tests, usually where there's a normal return from nothing happening

For the review, any opinions on the restructure, gaps & correctness checks are welcome!

### DIFF
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -230,15 +230,8 @@ def x_wind_cube():
     return fake_cube
 
 
-UMStash = namedtuple("UMStash",
-                     "long_name, name, units, standard_name, uniquename")
-
-
-@pytest.fixture
-def empty_um_var():
-    """Return an 'empty' um_var lookup where all names are empty strings."""
-    um_var = UMStash("", "", "", "", "")
-    return um_var
+# UMStash = namedtuple("UMStash",
+#                      "long_name, name, units, standard_name, uniquename")
 
 
 CellMethod = namedtuple("CellMethod", "method")
@@ -254,34 +247,30 @@ def min_cell_method():
     return CellMethod("minimum")
 
 
-def test_fix_var_name_simple(x_wind_cube, empty_um_var):
+def test_fix_var_name_simple(x_wind_cube):
     # NB: ignores cell methods functionality
     assert x_wind_cube.var_name == "var_name"  # dummy initial value
 
-    for um_var in (None, empty_um_var):
-        um2nc.fix_var_name(x_wind_cube, um_var, simple=True)
-        assert x_wind_cube.var_name == "fld_s00i002", f"Failed with um_var={um_var}"
+    for unique in (None, ""):
+        um2nc.fix_var_name(x_wind_cube, unique, simple=True)
+        assert x_wind_cube.var_name == "fld_s00i002", f"Failed with um_var={unique}"
 
 
-def test_fix_var_name_cell_methods_adds_max(x_wind_cube,
-                                                    empty_um_var,
-                                                    max_cell_method):
+def test_fix_var_name_cell_methods_adds_max(x_wind_cube, max_cell_method):
     # ensure maximum cell methods add suffix to cube name
     x_wind_cube.cell_methods = [max_cell_method]
 
-    for um_var in (None, empty_um_var):
-        um2nc.fix_var_name(x_wind_cube, um_var, simple=True)
+    for unique in (None, ""):
+        um2nc.fix_var_name(x_wind_cube, unique, simple=True)
         assert x_wind_cube.var_name == "fld_s00i002_max"
 
 
-def test_fix_var_name_cell_methods_adds_min(x_wind_cube,
-                                                    empty_um_var,
-                                                    min_cell_method):
+def test_fix_var_name_cell_methods_adds_min(x_wind_cube, min_cell_method):
     # ensure maximum cell methods add suffix to cube name
     x_wind_cube.cell_methods = [min_cell_method]
 
-    for um_var in (None, empty_um_var):
-        um2nc.fix_var_name(x_wind_cube, um_var, simple=True)
+    for unique in (None, ""):
+        um2nc.fix_var_name(x_wind_cube, unique, simple=True)
         assert x_wind_cube.var_name == "fld_s00i002_min"
 
 
@@ -289,43 +278,44 @@ def test_fix_var_name_unique(x_wind_cube):
     # ensure um unique name given to cubes in non-simple mode
     # NB: ignores cell methods functionality
     unique = "unique_string_name"
-    um_unique = UMStash("", "", "", "", unique)
-    um2nc.fix_var_name(x_wind_cube, um_unique, simple=False)
+    um2nc.fix_var_name(x_wind_cube, unique, simple=False)
     assert x_wind_cube.var_name == unique
 
 
-def test_fix_standard_name_update_x_wind(x_wind_cube, empty_um_var):
+def test_fix_standard_name_update_x_wind(x_wind_cube):
     # test cube wind renaming block only
-    # use empty um_var_empty_std to skip renaming logic
-    um2nc.fix_standard_name(x_wind_cube, empty_um_var, verbose=False)
+    # use empty std name to bypass renaming logic
+    um2nc.fix_standard_name(x_wind_cube, "", verbose=False)
     assert x_wind_cube.standard_name == "eastward_wind"
 
 
-def test_fix_standard_name_update_wind(empty_um_var):
+def test_fix_standard_name_update_wind():
     # test cube wind renaming block only
-    # use empty um_var_empty_std to skip renaming logic
+    # use empty std name to bypass renaming logic
     m_cube = PartialCube("var_name", {'STASH': DummyStash(0, 3)}, "y_wind")
     m_cube.cell_methods = []
 
-    um2nc.fix_standard_name(m_cube, empty_um_var, verbose=False)
+    um2nc.fix_standard_name(m_cube, "", verbose=False)
     assert m_cube.standard_name == "northward_wind"
 
 
 def test_fix_standard_name_with_mismatch(x_wind_cube):
     # ensure mismatching standard names between cube & um uses the um std name
-    um_var = UMStash("", "", "", "fake", "")
-    um2nc.fix_standard_name(x_wind_cube, um_var, verbose=False)
-    assert x_wind_cube.standard_name == um_var.standard_name
+    standard_name = "fake"
+    assert x_wind_cube.standard_name != standard_name
+    um2nc.fix_standard_name(x_wind_cube, standard_name, verbose=False)
+    assert x_wind_cube.standard_name == standard_name
 
 
 def test_fix_standard_name_with_mismatch_warn(x_wind_cube):
     # as per standard name mismatch, ensuring a warning is raised
-    um_var = UMStash("", "", "", "fake", "")
+    standard_name = "fake"
+    assert x_wind_cube.standard_name != standard_name
 
     with pytest.warns():
-        um2nc.fix_standard_name(x_wind_cube, um_var, verbose=True)
+        um2nc.fix_standard_name(x_wind_cube, standard_name, verbose=True)
 
-    assert x_wind_cube.standard_name == um_var.standard_name
+    assert x_wind_cube.standard_name == standard_name
 
 
 def test_fix_standard_name_add_missing_name_from_um(x_wind_cube):
@@ -333,9 +323,7 @@ def test_fix_standard_name_add_missing_name_from_um(x_wind_cube):
     for std_name in ("", None):
         x_wind_cube.standard_name = std_name
         expected = "standard-name-slot"
-        um_var = UMStash("", "", "", expected, "")
-        assert um_var.standard_name == expected
-        um2nc.fix_standard_name(x_wind_cube, um_var, verbose=False)
+        um2nc.fix_standard_name(x_wind_cube, expected, verbose=False)
         assert x_wind_cube.standard_name == expected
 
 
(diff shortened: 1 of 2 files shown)


umpost/um2netcdf.py
```diff
@@ -635,6 +601,55 @@ def add_global_history(infile, iris_out):
...
+
+    # Could there be cases with both max and min?
+    if c.var_name:
+        if any([m.method == 'maximum' for m in c.cell_methods]):
+            c.var_name += "_max"
+        if any([m.method == 'minimum' for m in c.cell_methods]):
+            c.var_name += "_min"
+
+
+def rename_cube_names(c, umvar, verbose: bool):
```
blimlim, line 621: I'm not sure which would be cleaner here, but what would your opinion be on splitting this into two functions – one which handles the `standard_name`, and one which handles the `long_name`? E.g so we'd end up with a total of three renaming functions: `rename_cube_var_name`, `rename_cube_standard_name`, and `rename_cube_long_name` each handling a seperate part of the cube. It would require changing the tests though.

test/test_um2netcdf.py
```diff
@@ -217,3 +220,121 @@ def test_cube_filtering_no_include_exclude(ua_plev_cube, heaviside_uv_cube):
...
+
+@pytest.fixture
+def min_cell_method():
+    return CellMethod("minimum")
+
+
+def test_rename_cube_var_name_simple(x_wind_cube, um_var_empty_std):
+    # NB: ignores cell methods functionality
+    assert x_wind_cube.var_name == "var_name"  # dummy initial value
+    um2nc.rename_cube_var_name(x_wind_cube, None, simple=True)
```
blimlim, line 260: Should this be using the `um_var_empty_std` fixture here in place of `None`?

test/test_um2netcdf.py
```diff
@@ -217,3 +220,121 @@ def test_cube_filtering_no_include_exclude(ua_plev_cube, heaviside_uv_cube):
...
+    assert x_wind_cube.var_name == "fld_s00i002"
+
+
+def test_rename_cube_var_rename_with_cell_methods_max(x_wind_cube,
+                                                      um_var_empty_std,
+                                                      max_cell_method):
+    x_wind_cube.cell_methods = [max_cell_method]
+
+    um2nc.rename_cube_var_name(x_wind_cube, None, simple=True)
```
blimlim, line 269: Same question here about the `um_var_empty_std` fixture

test/test_um2netcdf.py
```diff
@@ -217,3 +220,121 @@ def test_cube_filtering_no_include_exclude(ua_plev_cube, heaviside_uv_cube):
...
+    assert x_wind_cube.var_name == "fld_s00i002_max"
+
+
+def test_rename_cube_var_rename_with_cell_methods_min(x_wind_cube,
+                                                      um_var_empty_std,
+                                                      min_cell_method):
+    x_wind_cube.cell_methods = [min_cell_method]
+
+    um2nc.rename_cube_var_name(x_wind_cube, None, simple=True)
```
blimlim, line 278: Same question about the `um_var_empty_std` fixture

test/test_um2netcdf.py
```diff
@@ -217,3 +220,121 @@ def test_cube_filtering_no_include_exclude(ua_plev_cube, heaviside_uv_cube):
...
+def test_standard_name_umvar_name_mismatch_warn(x_wind_cube):
+    um_var = UMStash("", "", "", "fake", "")
+
+    with pytest.warns():
+        um2nc.rename_cube_names(x_wind_cube, um_var, verbose=True)
+
+    assert x_wind_cube.standard_name == um_var.standard_name
+
+
+def test_rename_cubes_long_name(x_wind_cube):
```
blimlim, line 318: Is this test missing the `assert` right at the end?

blimlim (changes requested):
Changes look good, I agree that pulling this code out of `process` helps clean it up quite a bit! I just noticed a couple of small missing things in the tests, and had a question about the structure of the new functions, though that one I think is just a matter of preference.

From what I can tell, the tests pretty extensively cover the different cases. One case that I think might be missing is for the `rename_cube_var_name` where `simple = True` and `um_var` contains a uniquename, just to check that here the `simple` setting wins out.

umpost/um2netcdf.py
truth-quark: Splitting these is a good idea, I didn't notice there's 2 separate blocks after reordering the naming logic.

truth-quark: This is fixed in https://github.com/ACCESS-NRI/um2nc-standalone/pull/52/commits/763dedd03268b25e47475def88a8d08a2a0c12f2.

test/test_um2netcdf.py
truth-quark: I'll run both `None` and the empty um var through the test to cover both bases.

truth-quark: Fixed in https://github.com/ACCESS-NRI/um2nc-standalone/pull/52/commits/4fabc4bb0f20cee4ee343eaa0932745030e58e79.

test/test_um2netcdf.py
truth-quark: Fixed in https://github.com/ACCESS-NRI/um2nc-standalone/pull/52/commits/4fabc4bb0f20cee4ee343eaa0932745030e58e79.

test/test_um2netcdf.py
truth-quark: Also fixed in https://github.com/ACCESS-NRI/um2nc-standalone/pull/52/commits/4fabc4bb0f20cee4ee343eaa0932745030e58e79.

test/test_um2netcdf.py
truth-quark: Yes... whoops!
It's fixed here https://github.com/ACCESS-NRI/um2nc-standalone/pull/52/commits/f9e1a3766c81b630e574583d009db9830f11719c.

blimlim (approved):
Changes look good to me!

truth-quark:
Many small changes have gone in, so this should be nicer for @jo-basevi to review....

blimlim (approved):

test/test_um2netcdf.py
```diff
@@ -217,3 +220,139 @@ def test_cube_filtering_no_include_exclude(ua_plev_cube, heaviside_uv_cube):
...
+
+
+def test_cube_um_standard_name_mismatch(x_wind_cube):
+    # ensure mismatching standard names between cube & um uses the um std name
+    um_var = UMStash("", "", "", "fake", "")
+    um2nc.rename_cube_standard_names(x_wind_cube, um_var, verbose=False)
+    assert x_wind_cube.standard_name == um_var.standard_name
+
+
+def test_test_cube_um_standard_name_mismatch_warn(x_wind_cube):
```
jo-basevi, line 321: Was `test_test_` intentional? 
```suggestion
def test_cube_um_standard_name_mismatch_warn(x_wind_cube):
```

umpost/um2netcdf.py
```diff
@@ -635,6 +602,81 @@ def add_global_history(infile, iris_out):
...
+    Parameters
+    ----------
+    cube : iris cube to modify (changes the name in place)
+    um_var : the UM Stash code structure
+    """
+    # Temporary work around for xconv
+    if cube.long_name and len(cube.long_name) > XCONV_LONG_NAME_LIMIT:
+        cube.long_name = cube.long_name[:XCONV_LONG_NAME_LIMIT]
+    elif um_var.long_name:
+        cube.long_name = um_var.long_name
```
jo-basevi, line 677: In the original code the condition was `if not cube.long_name and  um_var.long_name`, just  checking it was intentional to overwrite the `cube.long_name` if `cube.long_name` exists and is smaller than or equal the long name limit?

I may have missed it, but was this line covered in the tests?

umpost/um2netcdf.py
```diff
@@ -635,6 +602,81 @@ def add_global_history(infile, iris_out):
...
+                # TODO: remove verbose arg & always warn? Control warning visibility at cmd line?
+                if verbose:
+                    msg = (f"Standard name mismatch section={stash_code.section}"
+                           f" item={stash_code.item} standard_name={cube.standard_name}"
+                           f" UM var name={um_var.standard_name}")
+                    warnings.warn(msg)
+
+                cube.standard_name = um_var.standard_name
+    elif um_var.standard_name:
+        # If there's no standard_name or long_name from iris, use one from STASH
```
jo-basevi, line 660:
```suggestion
        # If there's no standard_name from iris, use one from STASH
```
Maybe add a similar comment to `rename_cube_long_names()`?

jo-basevi (commented):
The refactor and tests looks really neat!  Just some fairly minor questions

test/test_um2netcdf.py
```diff
@@ -217,3 +220,139 @@ def test_cube_filtering_no_include_exclude(ua_plev_cube, heaviside_uv_cube):
...
+def test_rename_cube_var_name_simple(x_wind_cube, empty_um_var):
+    # NB: ignores cell methods functionality
+    assert x_wind_cube.var_name == "var_name"  # dummy initial value
+
+    for um_var in (None, empty_um_var):
+        um2nc.rename_cube_var_name(x_wind_cube, um_var, simple=True)
+        assert x_wind_cube.var_name == "fld_s00i002", f"Failed with um_var={um_var}"
+
+
+def test_rename_cube_var_rename_cell_methods_adds_max(x_wind_cube,
```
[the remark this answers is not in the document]
truth-quark: Whoops, several test names were wrong due to earlier function splitting work. These are fixed in https://github.com/ACCESS-NRI/um2nc-standalone/pull/52/commits/9a560fe5f1e949f0489b1fb3440ebef5d2e0eac9.

test/test_um2netcdf.py
truth-quark: Ahh no, that's me being unco! It's fixed here: https://github.com/ACCESS-NRI/um2nc-standalone/pull/52/commits/9a560fe5f1e949f0489b1fb3440ebef5d2e0eac9.

umpost/um2netcdf.py
truth-quark: Oops, that logic was a bug from attempting to shrink things too much :-) I've reverted it.

Just in case, I've added some tests to check the do nothing conditional in https://github.com/ACCESS-NRI/um2nc-standalone/pull/52/commits/068c15aa54811d266b4b6d1d2bfe4f355b25188b.

umpost/um2netcdf.py
truth-quark: That's updated in https://github.com/ACCESS-NRI/um2nc-standalone/pull/52/commits/5be1420434ce79bf6923f0f9a177f645a4e59de4.

truth-quark:
@jo-basevi I requested a quick re-review as I made half a dozen commits of small changes...

truth-quark:
@blimlim & @jo-basevi I had some further thoughts on function naming consistency. The older code uses a `fix_ ... ` prefix, but I've inadvertently used a different scheme. Do you see any issue with changing the new code like: `rename_cube_long_name()` --> `fix_long_name()` ?

jo-basevi (approved):
LGTM

jo-basevi:
> Do you see any issue with changing the new code like: rename_cube_long_name() --> fix_long_name() ?

I can't see any issues, I think either function name will work

blimlim:
I agree, I think either name should be ok!

truth-quark:
> > Do you see any issue with changing the new code like: rename_cube_long_name() --> fix_long_name() ?
> 
> I can't see any issues, I think either function name will work

OK. I've done another refactor on another branch to:
* explore out the "fix" rename
* replace the `um_var` args with a plain data (this removes several `obj.attribute` lookups in the `fix` funcs)

This branch adds a couple of commits on top of `51/extract-renaming`:
https://github.com/ACCESS-NRI/um2nc-standalone/compare/51/extract-renaming-to-fix

Unit tests got simpler as this diff shows:
https://github.com/ACCESS-NRI/um2nc-standalone/commit/4aaa6077611ae05a4933773813f115d5f1b67dbe

@jo-basevi or @blimlim   would one of you mind a quick final look over the 26/7/2024 commits to check the new changes?

blimlim:
Just had a read through, I think those changes look good. A couple questions about the tests popped up while I was having a read through, which I've added as comments.

truth-quark:
> Just had a read through, I think those changes look good. A couple questions about the tests popped up while I was having a read through, which I've added as comments.

For reference, comments & replies are here https://github.com/ACCESS-NRI/um2nc-standalone/commit/4aaa6077611ae05a4933773813f115d5f1b67dbe

truth-quark:
> From what I can tell, the tests pretty extensively cover the different cases. One case that I think might be missing is for the `rename_cube_var_name` where `simple = True` and `um_var` contains a uniquename, just to check that here the `simple` setting wins out.

I've added an non-empty value to test this case in https://github.com/ACCESS-NRI/um2nc-standalone/pull/52/commits/8ab63954c105ff93b7664203ff33a513517d8ee2.

PR #57, opened 2024-07-26
47/driver script status reporting

### LINKED ISSUE #47: Driver script – improving success/failure reporting
The conversion driver script for ESM1.5 currently reports on the files it tries to convert and whether there were any permitted exceptions from within the `convert_fields_file_list` function:

https://github.com/ACCESS-NRI/um2nc-standalone/blob/1c15a92767ef112e449058d61aa3a3c3edce2f13/umpost/conversion_driver_esm1p5.py#L149-L160

This "buried I/O" could be pulled out of this function and dealt with by the higher level calling function. @truth-quark had the idea of modifying `convert_fields_file_list` to return a sequence specifying which conversions were successful, and which encountered allowed errors, so that the higher level `convert_esm1p5_output_dir` function could report them to the user.

blimlim:
This pull request closes #47.

It pulls the success/failure reporting out from the `convert_fields_file_list` function into a new `report_results` function, with the goal of making this bit of I/O more easily testable. Any suggestions about the structure and the tests would be very welcome!


umpost/conversion_driver_esm1p5.py
```diff
@@ -127,28 +128,58 @@ def convert_fields_file_list(fields_file_paths, nc_write_dir):
...
-                              fields_file_path.name)
+                failed.append((fields_file_path, exc))
             else:
+                # raise any unexpected errors
                 raise
 
+    return succeeded, failed
+
+
+def report_results(succeeded, failed):
```
truth-quark, line 159: An option here is to _format_ a report in this function, returning a `str`.

Then, `convert_esm1p5_output_dir()` can return `succeeded/failed`, which can be fed into the report formatter (in the main block). Then, `print` I/O for reporting is right up in the top level main block.

This should make the testing a bit simpler, as the reporting output can be tested without needing to mock `print`.

test/test_conversion_driver_esm1p5.py
```diff
@@ -142,3 +152,32 @@ def test_convert_esm1p5_output_dir_error():
...
             "/test_convert_esm1p5_output_dir_error/fake/path/"
         )
+
+def test_report_results_success():
+    succeeded_files = ["fake_file_1", "fake_file_2", "fake_file_3"]
+    # succeeded argument to report_results typically a list of Path objects
+    succeeded = [Path(p) for p in succeeded_files] 
+    failed = []
+
+    with mock.patch("builtins.print", return_value = None) as mock_print:
```
truth-quark, line 162: I've added a comment further down, for modifying the code to avoid `print` mocking.

test/test_conversion_driver_esm1p5.py
```diff
@@ -142,3 +152,32 @@ def test_convert_esm1p5_output_dir_error():
...
+
+def test_report_results_failed():
+    succeeded = []
+    failed = [
+        (Path("fake_file_1"), Exception("Error 1")),
+        (Path("fake_file_2"), Exception("Error 2")),
+        (Path("fake_file_3"), Exception("Error 3"))
+    ]
+
+    with mock.patch("warnings.warn", return_value = None) as mock_warning:
```
truth-quark, line 175: Does the `with pytest.warns(): ...` context manager work in place of the mock here?

truth-quark (changes requested):
This is a good first draft & I've found a few small things:
* I've made a few code structure comments
* Minor styling: there's a couple of PEP warnings with 1 or 3 empty lines instead of 2

There's a bit of subtle fun happening with the tests running against `um2netcdf4` on `gadi`, but not against the local `umpost/um2netcdf`. If I run all the tests, this breask by default unless I modify imports & mock "paths". One option (probably for another issue) is to allow local testing against `umpost/um2netcdf` by adding an environment variable that the tests use change the imports and mock paths. This can be removed when moving completely to the standalone version.

umpost/conversion_driver_esm1p5.py
```diff
@@ -127,28 +128,58 @@ def convert_fields_file_list(fields_file_paths, nc_write_dir):
...
-        fields_file_path = Path(fields_file_path)
+    for fields_file_path in fields_file_paths:
 
         nc_write_path = get_nc_write_path(fields_file_path, nc_write_dir)
 
-        print("Converting file " + fields_file_path.name)
-
         try:
             um2netcdf4.process(fields_file_path, nc_write_path, ARG_VALS)
+            succeeded.append(fields_file_path)
```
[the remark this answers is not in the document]
blimlim: Good question. We'll want to eventually delete the input files by default, and so perhaps its better to report the output files instead.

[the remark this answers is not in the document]
blimlim: I've updated this to use the NetCDF output paths when reporting successful conversions

One thing to note with this change, I had previously been testing that the reported successes matched the list of input paths, which no longer works when reporting the output paths. Do you think it's sufficient to just test the length of the reported success list, as below?

```python
def test_convert_fields_file_list_success(mock_process, input_list):
    input_list_paths = [Path(p) for p in input_list]
    succeeded, _ = esm1p5_convert.convert_fields_file_list(
        input_list_paths, "fake_nc_write_dir")

    assert mock_process.call_count == len(input_list)

    # TODO: Are we ok to not check the following?
    # assert succeeded == input_list_paths
```

umpost/conversion_driver_esm1p5.py
```diff
@@ -127,28 +128,58 @@ def convert_fields_file_list(fields_file_paths, nc_write_dir):
...
+    Returns
+    -------
+    None
+    """
+    for fields_file_path in succeeded:
+        print(f"Successfully converted {fields_file_path}")
+
+    for fields_file_path, exception in failed:
+        formatted_traceback = "".join(
+            traceback.format_exception(exception)
```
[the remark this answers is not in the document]
blimlim: Good idea, I've added this in

umpost/conversion_driver_esm1p5.py
blimlim: I've had a go at this, and split it into separate `format_successes` and `format_failures` to try and make the testing easier. Let me know if these changes look ok!

test/test_conversion_driver_esm1p5.py
blimlim: I'd used the patching here in order to count the number of times it was called, which possibly was unnecessary. I think we can skip this now though that the printing/warning has been moved up into the main block.

blimlim:
Thanks for the review and suggestions!

I've had a go at putting those changes in, let me know whether you have any other ideas. 

I wasn't too sure on how to best test the failure report formatting when the new `quiet` flag is `False`. I tried to set up a chain of exceptions to test it with:
```python
@pytest.fixture 
def raise_two_exceptions():
    def _raise_two_exceptions(exception_1, exception_2):
        try:
            raise exception_1 
        except Exception:
            raise exception_2 
    return _raise_two_exceptions


def test_format_failures_not_quiet(raise_two_exceptions):
    # Test that a multiple exceptions are reported when present in
    # stack trace and quiet is false.

    exception_1 = ValueError("Error 1")
    exception_2 = TypeError("Error_2")

    try:
        raise_two_exceptions(exception_1, exception_2)
    except Exception as exc:
        exc_with_traceback = exc 


    failed_file = Path("fake_file")
    fake_error = exc_with_traceback

    failed = [(failed_file, fake_error)]

    formatted_failure_report = esm1p5_convert.format_failures(
        failed,
        False
    )[0]

    assert type(exception_1).__name__ in formatted_failure_report
    assert type(exception_2).__name__ in formatted_failure_report

    assert exception_1.args[0] in formatted_failure_report
    assert exception_2.args[0] in formatted_failure_report
```

It would be great to know whether there might be a better way of testing it though!

blimlim:
I've updated the success/failure reporting with the suggested changes, including:
- Modifying the reporting function:
    - Splitting it into separate `format_successes` and `format_failures` functions to improve readability and testability.
    - Modifying these functions to return their formatted reports as strings, allowing I/O to be moved up to the top level and making testing easier.
    - Adding in a `-q, --quiet` command line flag to control the amount of detail in the error reporting.
 - Simplifying the test setup for the failure reporting by using `raise X from Y` to create chained exceptions.

Would be great to hear if any other suggestions come up from these changes!

truth-quark:
Let's start with a temp fix so tests "out of the box" irrespective of running `pytest` on `gadi` or locally. I've whipped up a quick guard block to change imports here: https://github.com/ACCESS-NRI/um2nc-standalone/commit/f4bf12904f5ea47450d5f137ceca9cf50e7d9a1e

Does it work in your `gadi` test environment?

blimlim:
Yeah that works for me! I've just merged it in

umpost/conversion_driver_esm1p5.py
```diff
@@ -127,28 +128,88 @@ def convert_fields_file_list(fields_file_paths, nc_write_dir):
...
+            report = report_base + "Final reported error: \n" + repr(exception)
+        else:
+            formatted_traceback = "".join(
+                traceback.format_exception(exception)
+            )
+            report = report_base + "Stack trace: \n" + formatted_traceback
+
+        failure_reports.append(report)
+
+    return failure_reports
```
truth-quark, line 220: I had similar thoughts about converting this to a generator & other mods:
1. The `if quiet` check is called on every loop iteration. This check can be "promoted" to the first/one off bool check, with the branches calling the loops.
2. It's less of a concern in this project, but concatenating strings is inefficient. One concise fix is to tweak the report format slightly: `yield report_base`, then `yield f"Final reported error ... {something}\n"`. This avoids the `str + str` & the caller simply prints some lines. This should to work the the tracebacks too.
3. Converting to a generator reduces memory use as report funcs emit the report gradually instead of creating it all at once.

umpost/conversion_driver_esm1p5.py
```diff
@@ -188,25 +252,47 @@ def convert_esm1p5_output_dir(esm1p5_output_dir):
...
+    if len(atm_dir_fields_files) == 0:
         warnings.warn(
             f"No files matching pattern '{fields_file_name_pattern}' "
             f"found in {current_atm_output_dir.resolve()}. No files will be "
             "converted to NetCDF."
         )
 
-   
+        # TODO: Check a better way of doing this
+        return  [], [] # Don't try to run the conversion
```
truth-quark, line 272: This is fine as the reporting funcs receive empty sequences & do nothing with the correct type.

umpost/conversion_driver_esm1p5.py
```diff
@@ -188,25 +252,47 @@ def convert_esm1p5_output_dir(esm1p5_output_dir):
...
+
+    return succeeded, failed
 
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument(
         "current_output_dir", help="ESM1.5 output directory to be converted", type=str
     )
+    parser.add_argument("--quiet", "-q", action="store-true",
```
truth-quark, line 287: This needs to be `store_true`. I'm not sure what happens if the action string doesn't match the values in the docs though...

umpost/conversion_driver_esm1p5.py
```diff
@@ -188,25 +252,47 @@ def convert_esm1p5_output_dir(esm1p5_output_dir):
...
 
     current_output_dir = args.current_output_dir
-    convert_esm1p5_output_dir(current_output_dir)
+
+    succeeded, failed = convert_esm1p5_output_dir(current_output_dir)
+
+    # Report results to user
+    success_reports = format_successes(succeeded)
+    failure_reports = format_failures(failed, args.quiet)
+    for success_message in success_reports:
```
truth-quark, line 303: It's a minor stylistic comment, but this could be `for success_message in format_successes(succeeded):` if using the generator approach.

test/test_conversion_driver_esm1p5.py
```diff
@@ -106,33 +114,48 @@ def _mock_process_with_exception(error_message):
...
+        [fake_file_path], "fake_nc_write_dir")
+
+    assert failed[0][0] == fake_file_path
+
+    # TODO: Testing the exception part of the reported failures will be easier
+    # once um2nc specific exceptions are added.
 
 
-def test_convert_fields_file_list_generic_error(mock_process_with_exception):
+def test_convert_fields_file_list_fail_generic(mock_process_with_exception):
```
truth-quark, line 153: Is this the test for when the conversion fails in a more critical fashion?

test/test_conversion_driver_esm1p5.py
```diff
@@ -142,3 +165,71 @@ def test_convert_esm1p5_output_dir_error():
...
+    # an iterable as input?
+
+    success_reports = esm1p5_convert.format_successes(succeeded_pairs)
+
+    # Check that the successful outputs make it into the report
+    for i, successful_output in enumerate(succeeded_outputs):
+        assert str(successful_output) in success_reports[i]
+
+
+def test_format_failures_quiet():
```
truth-quark, line 193: Minor style thing: the func could be called `test_format_failures_quiet_mode()` for a touch more clarity.

test/test_conversion_driver_esm1p5.py
```diff
@@ -142,3 +165,71 @@ def test_convert_esm1p5_output_dir_error():
...
+        failed,
+        True
+    )
+    for i, (file, exception) in enumerate(failed):
+        assert str(file) in formatted_failure_reports[i]
+        assert repr(exception) in formatted_failure_reports[i]
+
+
+
+def test_format_failures_not_quiet():
```
truth-quark, line 210: Minor style thing: how does `test_format_failures_standard_mode()` sound?

truth-quark (changes requested):
This is a nice template for the modularising driver scripts/reporting functionality within the org.

I've listed a few changes - we can cover these next week if needed?

test/test_conversion_driver_esm1p5.py
blimlim: Yeah! The main idea here is to have a list of exceptions which we want the driver to skip past when they occur without affecting the conversion of the rest of the files. E.g. ESM1.5 currently produces some timeseries output which will always cause the missing lon/lat coordinate error in `um2netcdf`, and we want the driver to just continue on past these files. For other exceptions though, we want the whole thing to crash which I'm testing here.

blimlim: Would it be worth renaming to something like `test_convert_fields_file_list_fail_critical` (and I guess adding a comment explaining the test...)

test/test_conversion_driver_esm1p5.py
blimlim: Yeah that sounds good! Just added in

umpost/conversion_driver_esm1p5.py
blimlim: Good idea. I've moved the `if quiet` check outside of the loop. 

For the string concatenation, is it ok to push those changes down the track? I think one challenge with the failure reporting, is that the final failure report is given to the user as a warning:
```python
    for failure_message in format_failures(failed, args.quiet):
        warnings.warn(failure_message)
```
and so it might be difficult to avoid combining the different parts of the reports together into a single string.

umpost/conversion_driver_esm1p5.py
```diff
@@ -127,28 +128,88 @@ def convert_fields_file_list(fields_file_paths, nc_write_dir):
...
+    success_reports: list of formatted reports of successful conversions.
+    """
+    success_reports = []
+    
+    for input_path, output_path in succeeded:
+        success_reports.append(
+            f"Successfully converted {output_path}"
+        )
+
+    return success_reports
```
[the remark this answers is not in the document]
truth-quark: Nice, the success function has shrunk down nicely as a generator. The failure one looks fine & simpler too.

umpost/conversion_driver_esm1p5.py
truth-quark: In https://github.com/ACCESS-NRI/um2nc-standalone/commit/440cf29f2de1324f93a9d4ecd24b304, it looks like you're yielding 1 report per failure, which should fit into a warning.

If the failure report was generating individual lines, something like `print(failure_msg, file=sys.stderr)` should work. This touches `sys.stderr`, which isn't quite as obvious as using the `warnings` module.

umpost/conversion_driver_esm1p5.py
truth-quark: Also, if `None, None` was returned, the workflow has to handle both `None` and sequence inputs. This way is simpler :-)

test/test_conversion_driver_esm1p5.py
truth-quark: Why does it seem easier to think of var/func names when reviewing someone else's work? 🤔

umpost/conversion_driver_esm1p5.py
```diff
@@ -188,25 +263,44 @@ def convert_esm1p5_output_dir(esm1p5_output_dir):
...
+                            "exceptions raised during conversion when flag is included."
+                            "Otherwise report full stack trace."
+                        )
+                        )
     args = parser.parse_args()
 
     current_output_dir = args.current_output_dir
-    convert_esm1p5_output_dir(current_output_dir)
+
+    succeeded, failed = convert_esm1p5_output_dir(current_output_dir)
```
truth-quark, line 299: There's some variable name shadowing with `succeeded` and `failed`.  Using `successes` and `failures` could work as var names, as they're names vs the style of prepending an underscore to "de-shadow" a var name.

truth-quark (approved):
This one is good to go with 2 simple fixes:
* Minor formatting to remove PyCharm linter errors. I added these in another branch https://github.com/ACCESS-NRI/um2nc-standalone/commit/8f7b6a65484b2172122d9e4a16f0a1f65d4a0a9a
* The minor name shadowing issue

blimlim:
> * Minor formatting to remove PyCharm linter errors. I added these in another branch [8f7b6a6](https://github.com/ACCESS-NRI/um2nc-standalone/commit/8f7b6a65484b2172122d9e4a16f0a1f65d4a0a9a)

Ah thanks for cleaning these up. I need to find a linter for vscode which highlights these types of things...

umpost/conversion_driver_esm1p5.py
blimlim: Good idea! Just fixed this up

umpost/conversion_driver_esm1p5.py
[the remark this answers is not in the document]
blimlim: As discussed further on, `convert_fields_file_list` now returns tuples of `(input_path, output_path)` for successful conversions.

blimlim:
Just pinging for a rereview of the final small style fixes as the branch protection requires one!

truth-quark (approved):
I added a commit to fix a missing newline @ the end.

PyCharm's linter is happy with the formatting/non-shadowing, so this change is good to merge!

CodeGat (approved):

Issue #26, opened 2024-06-21
Automatic code formatting


truth-quark:
Task: configure automatic code formatting with `black`, `ruff` or suitable tool.

Also document instructions for users to run formatters via pre-commit hooks.

atteggiani:
I vote for `Ruff` for both [linting](https://docs.astral.sh/ruff/linter/) and [formatting](https://docs.astral.sh/ruff/formatter/).

[pre-commit](https://pre-commit.com/) hooks and CI could be set up following replace_landsurface's [pre-commit-config](https://github.com/ACCESS-NRI/replace_landsurface/blob/a7bbe25b976c1372761670a7730355972d1d7a57/.pre-commit-config.yaml) and [CI](https://github.com/ACCESS-NRI/replace_landsurface/blob/96d67b3310bf54122bff14790d5fbb717d0575ef/.github/workflows/ci.yml).

Issue #11, opened 2024-06-18
Which versions of Python to support?


truth-quark:
Which versions of Python should we support?

From the June 2024 version list here:
https://devguide.python.org/versions/

Python 3.8 is close to EOL & 3.9 is more halfway through its lifecycle. 3.12 is the current active version.

Is focusing on 3.10 to 3.12 sufficient for `esm1.5` postprocessing?

truth-quark:
From the 6/8/2024 ESM1.5 standup, the decision is to support Python v3.10 and above.

truth-quark:
Tasks:
* [ ] Add a warning if the python env version <= 3.9

PR #58, opened 2024-07-31
Refactor process() testing & cleaning cube masking/filtering logic

### LINKED ISSUE #50: Clean & optimise masking logic/workflow
Some `heaviside` masking logic in `process()` can be simplified & optimised (this is a subset of #14).

One `um2nc-standalone` feature is using `heaviside_(uv|t)` cubes to mask data in some other cubes. Sometimes, both the `heaviside_(uv|t)` cubes are missing, preventing the masking operation. If masking cannot take place, it is understood that the _unmasked_ cubes are _excluded_ from the NC output.

As of mid-July 2024, masking logic is spread across the `process()` func & is somewhat tricky. This issue outlines some problems.

First, the decision to perform masking is based on flags determined here:

https://github.com/ACCESS-NRI/um2nc-standalone/blob/ab7b131333300854f0c37a3320026b0b53c6d15a/umpost/um2netcdf.py#L315-L323

The `do_mask` flag doesn't include chained logic to check if masking can actually occur:

https://github.com/ACCESS-NRI/um2nc-standalone/blob/ab7b131333300854f0c37a3320026b0b53c6d15a/umpost/um2netcdf.py#L318

Further down the `process()` workflow, the `do_mask` check _below_ assesses if a cube is required or excluded. This is _near the end_ of the workflow loop, **after** data processing steps. Thus, cubes may be modified then discarded (which should be avoided):

https://github.com/ACCESS-NRI/um2nc-standalone/blob/ab7b131333300854f0c37a3320026b0b53c6d15a/umpost/um2netcdf.py#L407-L419

**Solutions:**

It's simpler & more efficient to completely exclude _non-maskable_ cubes from processing, ideally early in the workflow. An initial suggestion is:

* Use [heaviside flags](https://github.com/ACCESS-NRI/um2nc-standalone/blob/40/refactor-masking/umpost/um2netcdf.py#L315-L316) to determine if masking is possible.
* Expand `do_mask` definition, checking if the user wants masking AND `heaviside uv|t` cubes are present. This should resolve to `False` if **both** `heaviside uv|t` cubes are missing.
* Is this possible: simplify `check_pressure_level_masking()` to **only** return `heaviside uv|t` if needed AND the _mask layer is present_? This would combine the `needs_heaviside_uv|t` flag into heaviside existence tests `if heaviside_t: ...`
* Fix the logic for generating the variable skipping warnings
* If masking isn't possible, filter the cubes to remove any requiring masking. This subsequently excludes some cubes from ever reaching this loop: https://github.com/ACCESS-NRI/um2nc-standalone/blob/ab7b131333300854f0c37a3320026b0b53c6d15a/umpost/um2netcdf.py#L339
* Refine the logic & simplify control flow. The `continue` statements _below_ relate to a loop higher in the code. These `continue` statements quietly skip cube writing, so control flow & behaviour is less clear: https://github.com/ACCESS-NRI/um2nc-standalone/blob/ab7b131333300854f0c37a3320026b0b53c6d15a/umpost/um2netcdf.py#L407-L419

An initial experiment cleaning up cube filtering and masking logic is here:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/40/refactor-masking-filtering/umpost/um2netcdf.py

This needs more work to compress some bool checks & warnings. Consider filtering in the existing logic checking function?

truth-quark:
Resolves #50 (also related to #14).

This is a fairly complicated PR around testing & refactoring the central `process()` workflow function in `um2netcdf/um2nc`.

This work is intended to provide initial testing & cleanup of `process()`, providing a framework to ensure the workflow is consistent over future refactoring steps.

**For background:**
* `um2nc` processes UM files, modifying variables & saving to NetCDF files
* The core `um2nc` driver is the `process()` workflow function (it's also the main API entry point)
* Commit 0 in this repo had no unit tests
* `process()` was initially 140 lines of processing logic & too difficult to unit test
* Recent work extracted blocks of functionality (e.g. variable renaming) to separate functions. This simplified & compressed `process()` to focus on the workflow logic
* As `process()` shrank, I identified tricky logic & inefficient data handling (see #50 notes)
* The requirement to retrofit unit testing naturally encourages refactoring to simplify workflow logic

**Known Problems:**
* Two `process()`  tests require _substantial setup_ & multiple ugly ~~nested~~ `mock.patch()` calls. This is partly due to complexity in `process()`, touching separate I/O libraries.
* Both `mule` and `iris` fields file objects are complex
  - Multiple attrs are created _dynamically_ when opened
  - Mocking these is tricky with the `spec=` option, since `spec` only detects init time attributes
* Some mocks may not represent real object config, however I've tweaked some code to skip some side issues
* `process()` doesn't return success or failure data for the cubes

**Future steps/direction/thoughts:**
* For upcoming steps, `process()` is likely to gain cube modification functionality as it is extracted from `cubewrite()`
* `process()` tests should be flexible to expand to cover the additional mod steps
* `process()` needs reordering to divide into 3 broad steps: input I/O, cube modification, output I/O
  - Idea: sub-steps could be refactored to independent sub-workflow funcs, which suggests `process()` tests could be split into smaller chunks. This ought to reduce setup dependencies, reducing test setup burden (e.g. less `mock.patch()` nesting)

For the review, any opinions on the restructure, gaps & correctness checks are welcome! It would be good to reduce the complex setup.

### DIFF
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -175,45 +175,47 @@ def test_process_all_cubes_filtered(air_temp_cube, mule_vars, std_args,
 def test_process_masking(air_temp_cube, precipitation_flux_cube,
                          heaviside_uv_cube, heaviside_t_cube,
                          mule_vars, std_args, fake_in_path, fake_out_path):
-    """Run process() with masking cubes."""
-    with mock.patch("mule.load_umfile"):  # ignore m_load_umfile as process_mule_vars is mocked
-        with mock.patch("umpost.um2netcdf.process_mule_vars") as m_mule_vars:
-            m_mule_vars.return_value = mule_vars
+    """Run process() with pressure level masking cubes present."""
+    with (
+        mock.patch("mule.load_umfile"),  # ignore m_load_umfile as process_mule_vars is mocked
+        mock.patch("umpost.um2netcdf.process_mule_vars") as m_mule_vars,
+        mock.patch("iris.load") as m_iris_load,
+        mock.patch("iris.fileformats.netcdf.Saver") as m_saver,  # prevent I/O
 
-            with mock.patch("iris.load") as m_iris_load:
-                # add cube requiring heaviside_t masking to enable both uv & t code branches
-                geo_potential_cube = DummyCube(30297, "geopotential_height")
+        # TODO: fix lat/lon & levels requires c.coord attributes
+        #       use fixtures to add attrs & remove the patches?
+        mock.patch("umpost.um2netcdf.fix_latlon_coord") as m_coord,
+        mock.patch("umpost.um2netcdf.fix_level_coord") as m_level,
+        mock.patch("umpost.um2netcdf.apply_mask") as m_apply_mask,
+        mock.patch("umpost.um2netcdf.cubewrite") as m_cubewrite,
+    ):
+        m_mule_vars.return_value = mule_vars
 
-                cubes = [air_temp_cube, precipitation_flux_cube, geo_potential_cube,
-                         heaviside_uv_cube, heaviside_t_cube]
+        # add cube requiring heaviside_t masking to enable both uv & t code branches
+        geo_potential_cube = DummyCube(30297, "geopotential_height")
 
-                for c in cubes:
-                    attrs = {um2nc.STASH: DummyStash(*split_item_code(c.item_code))}
-                    c.attributes = attrs
-                    c.cell_methods = []
+        cubes = [air_temp_cube, precipitation_flux_cube, geo_potential_cube,
+                 heaviside_uv_cube, heaviside_t_cube]
 
-                m_iris_load.return_value = cubes
+        for c in cubes:
+            attrs = {um2nc.STASH: DummyStash(*split_item_code(c.item_code))}
+            c.attributes = attrs
+            c.cell_methods = []
 
-                with mock.patch("iris.fileformats.netcdf.Saver") as m_saver:  # prevent I/O
-                    m_sman = mock.Mock()
-                    m_saver().__enter__.return_value = m_sman
+        m_iris_load.return_value = cubes
+        m_sman = mock.Mock()
+        m_saver().__enter__.return_value = m_sman
 
-                    # TODO: fix lat/lon & levels requires c.coord attributes
-                    #       use fixtures to add attrs & remove the patches?
-                    with mock.patch("umpost.um2netcdf.fix_latlon_coord") as m_coord:
-                        with mock.patch("umpost.um2netcdf.fix_level_coord") as m_level:
-                            with mock.patch("umpost.um2netcdf.apply_mask") as m_apply_mask:
-                                with mock.patch("umpost.um2netcdf.cubewrite") as m_cubewrite:
-                                    um2nc.process(fake_in_path, fake_out_path, std_args)
+        um2nc.process(fake_in_path, fake_out_path, std_args)
 
-                                    assert m_sman.update_global_attributes.called
-                                    assert m_sman.update_global_attributes.call_count == 2
-                                    assert m_saver.write.called is False  # write I/O prevented
-                                    assert m_coord.called
-                                    assert m_level.called
-                                    assert m_apply_mask.called
-                                    assert m_cubewrite.called  # real cubewrite() should be prevented
-                                    assert m_cubewrite.call_count == len(cubes)
+        assert m_sman.update_global_attributes.called
+        assert m_sman.update_global_attributes.call_count == 2
+        assert m_saver.write.called is False  # write I/O prevented
+        assert m_coord.called
+        assert m_level.called
+        assert m_apply_mask.called
+        assert m_cubewrite.called  # real cubewrite() should be prevented
+        assert m_cubewrite.call_count == len(cubes)
 
 
 def split_item_code(item_code: int):
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -105,46 +105,48 @@ def test_process_without_masking(air_temp_cube, precipitation_flux_cube, mule_va
     #        use these tests to gradually refactor process()
     # TODO: move towards a design where input & output I/O is extracted from process()
     #       process()'s core should operate with *data only* args
-    with mock.patch("mule.load_umfile"):  # ignore m_load_umfile as process_mule_vars is mocked
-        with mock.patch("umpost.um2netcdf.process_mule_vars") as m_mule_vars:
-            m_mule_vars.return_value = mule_vars
+    with (
+        mock.patch("mule.load_umfile"),  # ignore m_load_umfile as process_mule_vars is mocked
+        mock.patch("umpost.um2netcdf.process_mule_vars") as m_mule_vars,
+        mock.patch("iris.load") as m_iris_load,
+        mock.patch("iris.fileformats.netcdf.Saver") as m_saver,  # prevent I/O
 
-            with mock.patch("iris.load") as m_iris_load:
-                cubes = [air_temp_cube, precipitation_flux_cube]
+        # TODO: fix lat/lon & levels requires c.coord attribute
+        #       use fixtures to add attrs & remove the patches?
+        mock.patch("umpost.um2netcdf.fix_latlon_coord") as m_coord,
+        mock.patch("umpost.um2netcdf.fix_level_coord") as m_level,
+        mock.patch("umpost.um2netcdf.apply_mask") as m_apply_mask,
+        mock.patch("umpost.um2netcdf.cubewrite") as m_cubewrite,
+    ):
+        m_mule_vars.return_value = mule_vars
+        cubes = [air_temp_cube, precipitation_flux_cube]
 
-                for c in cubes:
-                    c.attributes = {um2nc.STASH: DummyStash(*split_item_code(c.item_code))}
-                    c.cell_methods = []
+        for c in cubes:
+            c.attributes = {um2nc.STASH: DummyStash(*split_item_code(c.item_code))}
+            c.cell_methods = []
 
-                    # TODO: replace with DummyCubeWithCoords ()?
-                    #       or add coord fixtures?
-                    c.coord["latitude"] = 0.0  # FIXME
-                    c.coord["longitude"] = 0.0  # FIXME
+            # TODO: replace with DummyCubeWithCoords ()?
+            #       or add coord fixtures?
+            c.coord["latitude"] = 0.0  # FIXME
+            c.coord["longitude"] = 0.0  # FIXME
 
-                m_iris_load.return_value = cubes
+        m_iris_load.return_value = cubes
 
-                with mock.patch("iris.fileformats.netcdf.Saver") as m_saver:  # prevent I/O
-                    # mock `sman` var to prevent I/O
-                    m_sman = mock.Mock()
-                    m_saver().__enter__.return_value = m_sman
+        # mock `sman` var to prevent I/O
+        m_sman = mock.Mock()
+        m_saver().__enter__.return_value = m_sman
+        std_args.verbose = True  # test some warning branches
 
-                    # TODO: fix lat/lon & levels requires c.coord attribute
-                    #       use fixtures to add attrs & remove the patches?
-                    with mock.patch("umpost.um2netcdf.fix_latlon_coord") as m_coord:
-                        with mock.patch("umpost.um2netcdf.fix_level_coord") as m_level:
-                            with mock.patch("umpost.um2netcdf.apply_mask") as m_apply_mask:
-                                with mock.patch("umpost.um2netcdf.cubewrite") as m_cubewrite:
-                                    std_args.verbose = True  # test some warning branches
-                                    um2nc.process(fake_in_path, fake_out_path, std_args)
+        um2nc.process(fake_in_path, fake_out_path, std_args)
 
-                                    assert m_sman.update_global_attributes.called
-                                    assert m_saver.write.called is False  # write I/O prevented
-                                    assert m_coord.called
-                                    assert m_level.called
-                                    assert m_apply_mask.called is False
-                                    assert m_cubewrite.called  # real cubewrite() prevented
-                                    assert m_cubewrite.call_count == 1
-                                    assert m_cubewrite.call_args_list[0].args[0] == precipitation_flux_cube
+        assert m_sman.update_global_attributes.called
+        assert m_saver.write.called is False  # write I/O prevented
+        assert m_coord.called
+        assert m_level.called
+        assert m_apply_mask.called is False
+        assert m_cubewrite.called  # real cubewrite() prevented
+        assert m_cubewrite.call_count == 1
+        assert m_cubewrite.call_args_list[0].args[0] == precipitation_flux_cube
 
 
 def test_process_all_cubes_filtered(air_temp_cube, mule_vars, std_args,
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -158,7 +158,7 @@ def test_process_no_heaviside_drop_cubes(air_temp_cube, precipitation_flux_cube,
         cubes = [air_temp_cube, precipitation_flux_cube, geo_potential_cube]
 
         m_iris_load.return_value = cubes
-        m_saver().__enter__.return_value = mock.Mock("mock_sman")
+        m_saver().__enter__ = mock.Mock(name="mock_sman")
         std_args.verbose = True  # test some warning branches
 
         # air temp & geo potential should be dropped in process()
@@ -183,7 +183,7 @@ def test_process_all_cubes_filtered(air_temp_cube, geo_potential_cube,
     ):
         m_mule_vars.return_value = mule_vars
         m_iris_load.return_value = [air_temp_cube, geo_potential_cube]
-        m_saver().__enter__.return_value = mock.Mock("mock_sman")
+        m_saver().__enter__ = mock.Mock(name="mock_sman")
 
         # all cubes should be dropped
         assert um2nc.process(fake_in_path, fake_out_path, std_args) == []
@@ -220,7 +220,7 @@ def test_process_mask_with_heaviside(air_temp_cube, precipitation_flux_cube,
             c.cell_methods = []
 
         m_iris_load.return_value = cubes
-        m_saver().__enter__.return_value = mock.Mock("mock_sman")
+        m_saver().__enter__ = mock.Mock(name="mock_sman")
 
         # all cubes should be processed & not dropped
         processed = um2nc.process(fake_in_path, fake_out_path, std_args)
@@ -248,7 +248,7 @@ def test_process_no_masking_keep_all_cubes(air_temp_cube, precipitation_flux_cub
         cubes = [air_temp_cube, precipitation_flux_cube, geo_potential_cube]
 
         m_iris_load.return_value = cubes
-        m_saver().__enter__.return_value = mock.Mock("mock_sman")
+        m_saver().__enter__ = mock.Mock(name="mock_sman")
         std_args.nomask = True
 
         # all cubes should be kept with masking off
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -68,6 +68,7 @@ def set_default_attrs(cube, item_code: int, var_name: str):
                           "coord": {"latitude": 0.0,  # TODO: real val = ?
                                     "longitude": 0.0},  # TODO: real val
                           "cell_methods": [],
+                          "data": None,
                           })
 
     section, item = split_item_code(item_code)
@@ -105,7 +106,7 @@ def geo_potential_cube():
 def std_args():
     # TODO: make args namedtuple?
     args = mock.Mock()
-    args.nomask = False
+    args.nomask = False  # perform masking if possible
     args.nohist = False
     args.nckind = 3
     args.include_list = None
@@ -130,48 +131,37 @@ def fake_out_path():
 def test_process_no_heaviside_drop_cubes(air_temp_cube, precipitation_flux_cube,
                                          mule_vars, std_args,
                                          fake_in_path, fake_out_path):
-    """Attempts end-to-end test of process(), ignoring cubes requiring masking."""
+    """Attempt end-to-end process() test, dropping cubes requiring masking."""
 
-    # FIXME: this convoluted setup is a big code stench
+    # FIXME: this convoluted setup is a code smell
     #        use these tests to gradually refactor process()
     # TODO: move towards a design where input & output I/O is extracted from process()
-    #       process()'s core should operate with *data only* args
+    #       process() should eventually operate on *data only* args
     with (
         mock.patch("mule.load_umfile"),  # ignore m_load_umfile as process_mule_vars is mocked
         mock.patch("umpost.um2netcdf.process_mule_vars") as m_mule_vars,
         mock.patch("iris.load") as m_iris_load,
         mock.patch("iris.fileformats.netcdf.Saver") as m_saver,  # prevent I/O
-
-        # TODO: fix lat/lon & levels requires c.coord attribute
-        #       use fixtures to add attrs & remove the patches?
-        mock.patch("umpost.um2netcdf.fix_latlon_coord") as m_coord,
-        mock.patch("umpost.um2netcdf.fix_level_coord") as m_level,
-        mock.patch("umpost.um2netcdf.apply_mask") as m_apply_mask,
-        mock.patch("umpost.um2netcdf.cubewrite") as m_cubewrite,
+        mock.patch("umpost.um2netcdf.fix_latlon_coord"),
+        mock.patch("umpost.um2netcdf.fix_level_coord"),
+        mock.patch("umpost.um2netcdf.apply_mask"),
+        mock.patch("umpost.um2netcdf.cubewrite"),
     ):
         m_mule_vars.return_value = mule_vars
         cubes = [air_temp_cube, precipitation_flux_cube]
 
         m_iris_load.return_value = cubes
-
-        # mock `sman` var to prevent I/O
-        m_sman = mock.Mock()
+        m_sman = mock.Mock()  # mock `sman` var to prevent I/O
         m_saver().__enter__.return_value = m_sman
         std_args.verbose = True  # test some warning branches
 
-        # air temp should be excluded in this setup
-        # TODO: how to test precip layer is the only modified cube?
-        um2nc.process(fake_in_path, fake_out_path, std_args)
+        # air temp should be dropped in process()
+        processed = um2nc.process(fake_in_path, fake_out_path, std_args)
+        assert len(processed) == 1
+        cube = processed[0]
 
-        # TODO: focus asserts on the cube
-        assert m_sman.update_global_attributes.called
-        assert m_sman.write.called is False  # write I/O prevented
-        assert m_coord.called
-        assert m_level.called
-        assert m_apply_mask.called is False
-        assert m_cubewrite.called  # real cubewrite() prevented
-        assert m_cubewrite.call_count == 1
-        assert m_cubewrite.call_args_list[0].args[0] == precipitation_flux_cube
+        assert cube is precipitation_flux_cube
+        assert cube.data is None  # masking wasn't applied
 
 
 def test_process_all_cubes_filtered(air_temp_cube, mule_vars, std_args,
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -5,58 +5,104 @@ from collections import namedtuple
 import umpost.um2netcdf as um2nc
 
 import pytest
+import numpy as np
+
 import mule
 import mule.ff
 
 
-def test_process(ua_plev_cube, heaviside_uv_cube, ta_plev_cube):
-    """TODO"""
+@pytest.fixture
+def z_sea_rho_data():
+    # data ripped from aiihca.paa1jan.subset: ff.level_dependent_constants.zsea_at_rho
+    # TODO: dtype is object, should it be float?
+    data = np.array([9.9982061118072, 49.998881525751194, 130.00023235363918,
+                     249.99833311211358, 410.00103476788956, 610.000486354252,
+                     850.0006133545584, 1130.0014157688088, 1449.9989681136456,
+                     1810.0011213557837, 2210.0000245285087, 2649.9996031151773,
+                     3129.9998571157903, 3650.000786530347, 4209.99846587549,
+                     4810.000746117935, 5449.999776290966, 6129.999481877941,
+                     6849.999862878861, 7610.000919293723, 8409.998725639172,
+                     9250.001132881924, 10130.00029005526, 11050.000122642545,
+                     12010.000630643768, 13010.001814058938, 14050.40014670717,
+                     15137.719781928794, 16284.973697054254, 17506.96881530842,
+                     18820.820244130424, 20246.59897992768, 21808.13663216417,
+                     23542.18357603375, 25520.960854349545, 27901.358260464756,
+                     31063.888598164976, 36081.76331548462, -1073741824.0], dtype=object)
+    return data
 
-    # FIXME: this convoluted setup is a huge code stench
+
+@pytest.fixture
+def z_sea_theta_data():
+    # data ripped from aiihca.paa1jan.subset: ff.level_dependent_constants.zsea_at_theta
+    # TODO: dtype is object, should it be float?
+    data = np.array([0.0, 20.000337706971997, 80.00135082788799, 179.9991138793904,
+                     320.00147782819437, 500.00059170758476, 720.0003810009191,
+                     980.0008457081975, 1279.9980603460624, 1619.9998758812287,
+                     1999.9984413469813, 2420.001607710036, 2880.0015240036764,
+                     3379.9981902279037, 3919.9994573494323, 4500.001399884905,
+                     5120.000092350965, 5779.999460230968, 6479.999503524915,
+                     7220.000222232806, 8000.001616354641, 8819.999760407061,
+                     9679.998579873429, 10579.998074753737, 11519.998245047991,
+                     12499.999090756188, 13520.000611878331, 14580.799681536007,
+                     15694.639882321579, 16875.311437270288, 18138.62619334655,
+                     19503.01036943094, 20990.18759042441, 22626.081748420565,
+                     24458.285403646936, 26583.640230535515, 29219.080215877355,
+                     32908.69305496925, 39254.833576], dtype=object)
+    return data
+
+
+@pytest.fixture
+def mule_vars(z_sea_rho_data, z_sea_theta_data):
+    """Simulate mule variables from aiihca.paa1jan.subset data."""
+    d_lat = 1.25  # spacing manually copied from aiihca.paa1jan.subset file
+    d_lon = 1.875
+    return um2nc.MuleVars(um2nc.GRID_NEW_DYNAMICS, d_lat, d_lon, z_sea_rho_data, z_sea_theta_data)
+
+
+def test_process(ua_plev_cube, heaviside_uv_cube, ta_plev_cube, mule_vars):
+    """Attempts end to end test of process()."""
+
+    # FIXME: this convoluted setup is a big code stench
     #        use this to gradually refactor process()
-    with (mock.patch("mule.load_umfile") as m_load_umfile):
+    #        add naming vars to the dummy cubes
+    with mock.patch("mule.load_umfile") as m_load_umfile:
         m_fields_file = mock.MagicMock(spec=mule.ff.FieldsFile)
         m_load_umfile.return_value = m_fields_file
 
-        with mock.patch("umpost.um2netcdf.get_grid_type") as m_grid_type:
-            m_grid_type.return_value = um2nc.GRID_NEW_DYNAMICS
+        with mock.patch("umpost.um2netcdf.process_mule_vars") as m_mule_vars:
+            m_mule_vars.return_value = mule_vars
 
-            with mock.patch("umpost.um2netcdf.get_grid_spacing") as m_spacing:
-                m_spacing.return_value = None, None  # TODO: use reasonable values
-                with mock.patch("umpost.um2netcdf.get_z_sea_constants") as m_sea:
-                    m_sea.return_value = None, None  # TODO: use reasonable values
+            with mock.patch("iris.load") as m_iris_load:
+                with mock.patch("iris.fileformats.netcdf.Saver") as m_saver:  # prevent I/O
+                    with mock.patch("umpost.um2netcdf.fix_latlon_coord") as m_coord:
+                        with mock.patch("umpost.um2netcdf.fix_level_coord") as m_level:
+                            with mock.patch("umpost.um2netcdf.apply_mask") as m_apply_mask:
+                                with mock.patch("umpost.um2netcdf.cubewrite") as m_cubewrite:
+                                    cubes = [ua_plev_cube, heaviside_uv_cube, ta_plev_cube]
 
-                    with mock.patch("iris.load") as m_iris_load:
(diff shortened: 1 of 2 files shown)


test/test_um2netcdf.py
```diff
@@ -5,7 +5,209 @@
...
+                m_iris_load.return_value = cubes
+
+                with mock.patch("iris.fileformats.netcdf.Saver") as m_saver:  # prevent I/O
+                    # mock `sman` var to prevent I/O
+                    m_sman = mock.Mock()
+                    m_saver().__enter__.return_value = m_sman
+
+                    # TODO: fix lat/lon & levels requires c.coord attribute
+                    #       use fixtures to add attrs & remove the patches?
+                    with mock.patch("umpost.um2netcdf.fix_latlon_coord") as m_coord:
```
truth-quark, line 119: Here's an example of nested `mock.patch()` fun. These are _pseudo practical_ mocks to ignore code which requires a `c.coord` data attribute. Spencer is working on a test data structure with `coords`, which could potentially fit here & reduce the number of patches.

test/test_um2netcdf.py
```diff
@@ -5,7 +5,209 @@
...
+
+                with mock.patch("iris.fileformats.netcdf.Saver") as m_saver:  # prevent I/O
+                    m_sman = mock.Mock()
+                    m_saver().__enter__.return_value = m_sman
+
+                    infile = "/tmp/fake_input_fields_file"
+                    outfile = "/tmp/fake_input_fields_file.nc"
+                    um2nc.process(infile, outfile, std_args)
+
+                    assert m_sman.update_global_attributes.called is False
```
truth-quark, line 158: This logic is assumed as a proxy for `process()` exiting early if all the cubes are filtered out.

umpost/um2netcdf.py
```diff
@@ -352,34 +345,61 @@ def process(infile, outfile, args):
...
 
             if args.verbose:
                 print(c.name(), c.item_code)
 
+            # TODO: split cubewrite ops into funcs & bring those steps into process() workflow
+            #       or a sub process workflow function (like process_mule_vars())
             cubewrite(c, sman, args.compression, args.use64bit, args.verbose)
 
 
+MuleVars = collections.namedtuple("MuleVars", "grid_type, d_lat, d_lon, z_rho, z_theta")
```
truth-quark, line 372: For info: `mule` functionality had to be extracted to a sub-workflow simplify test setup (otherwise 4 _more_ mock patches were required). What are folks' thoughts on extracting blocks like this to sub-workflows?

CodeGat:
Added branch protections to this repo so it'll restrict merging without a review.

umpost/um2netcdf.py
```diff
@@ -530,24 +550,33 @@ def is_heaviside_t(item_code):
     return item_code == 30304
 
 
-def check_pressure_warnings(need_heaviside_uv, heaviside_uv, need_heaviside_t, heaviside_t):
+def no_masking_cubes(cubes, heaviside_uv, heaviside_t, verbose: bool):
```
truth-quark, line 553: Note to self: this is missing standalone unit tests.

test/test_um2netcdf.py
```diff
@@ -5,7 +5,209 @@
...
+                    # mock `sman` var to prevent I/O
+                    m_sman = mock.Mock()
+                    m_saver().__enter__.return_value = m_sman
+
+                    # TODO: fix lat/lon & levels requires c.coord attribute
+                    #       use fixtures to add attrs & remove the patches?
+                    with mock.patch("umpost.um2netcdf.fix_latlon_coord") as m_coord:
+                        with mock.patch("umpost.um2netcdf.fix_level_coord") as m_level:
+                            with mock.patch("umpost.um2netcdf.apply_mask") as m_apply_mask:
+                                with mock.patch("umpost.um2netcdf.cubewrite") as m_cubewrite:
```
CodeGat, line 122: Could we remove some of the nesting by having a multiple context manager? We're using `python >=3.10`, right? Could do something like this:
```py
with (
    mock.patch("umpost.um2netcdf.fix_latlon_coord") as m_coord,
    mock.patch("umpost.um2netcdf.fix_level_coord") as m_level,
    mock.patch("umpost.um2netcdf.apply_mask") as m_apply_mask,
    mock.patch("umpost.um2netcdf.cubewrite") as m_cubewrite,
):
    # usages of those contexts
```

test/test_um2netcdf.py
```diff
@@ -5,7 +5,209 @@
...
+                                    assert m_cubewrite.call_args_list[0].args[0] == precipitation_flux_cube
+
+
+def test_process_all_cubes_filtered(air_temp_cube, mule_vars, std_args):
+    """Ensure process() exists early if all cubes are removed in filtering."""
+    with mock.patch("mule.load_umfile"):  # ignore m_load_umfile as process_mule_vars is mocked
+        with mock.patch("umpost.um2netcdf.process_mule_vars") as m_mule_vars:
+            m_mule_vars.return_value = mule_vars
+
+            with mock.patch("iris.load") as m_iris_load:
```
CodeGat, line 145: In fact can a lot of these contexts be used as a multiple context `with`? Keeps the nesting down...assuming the `m_mule_vars.return_value` doesn't need to be set early.

CodeGat (commented):
I don't think I'd be able to comment of the design as it's a bit above my level...but I do have some suggestions

test/test_um2netcdf.py
marc-white: I'm new to the ACCESS-NRI ecosystem, can Python >= 3.10 be guaranteed? I seem to recall seeing some Python 3.9 references in other projects/systems...

test/test_um2netcdf.py
```diff
@@ -5,7 +5,209 @@
...
+                    m_saver().__enter__.return_value = m_sman
+
+                    # TODO: fix lat/lon & levels requires c.coord attribute
+                    #       use fixtures to add attrs & remove the patches?
+                    with mock.patch("umpost.um2netcdf.fix_latlon_coord") as m_coord:
+                        with mock.patch("umpost.um2netcdf.fix_level_coord") as m_level:
+                            with mock.patch("umpost.um2netcdf.apply_mask") as m_apply_mask:
+                                with mock.patch("umpost.um2netcdf.cubewrite") as m_cubewrite:
+                                    infile = "/tmp/fake_input_fields_file"
+                                    outfile = "/tmp/fake_input_fields_file.nc"
```
marc-white, line 124: Can the tempfile module be used here instead of direct reference to /tmp?

test/test_um2netcdf.py
truth-quark: Issue #11 discusses our initial intent is to support Py 3.10+, although it's possible users require 3.8 or 3.9.
* Apparently the parenthesised context managers exist as a new feature in Py3.9: https://stackoverflow.com/questions/68924790/parenthesized-context-managers-work-in-python-3-9-but-not-3-8
* From 3.10+ we're safe with parenthesised context manager support:
https://docs.python.org/3.10/whatsnew/3.10.html#parenthesized-context-managers
* v3.9 is EOL in Oct 2025: https://devguide.python.org/versions/, so ~15 months off.

This leaves some options:
* Conservative deferral: upgrade to new ctx managers in October 2025
* Short term defer: assume 3.9 is safe with the new features, but defer upgrade to Oct 2024/3.8 EOL
* Fix now: assume baseline 3.10 adoption across the board

@aidanheerdegen would other stakeholders need to comment on their Py version requirements?

umpost/um2netcdf.py
```diff
@@ -352,34 +345,61 @@ def process(infile, outfile, args):
...
+    Returns
+    -------
+    A MuleVars data structure.
+    """
+    if isinstance(fields_file, mule.ancil.AncilFile):
+        raise NotImplementedError('Ancillary files are currently not supported')
+
+    if mule.__version__ == "2020.01.1":
+        msg = "mule 2020.01.1 doesn't handle pathlib Paths properly"
+        raise NotImplementedError(msg)  # fail fast
```
marc-white, line 394: You could move this to an __init__ at some level to make it fail even faster, unless you wanted other aspects of the module to be available with mule 2020.01.1

test/test_um2netcdf.py
aidanheerdegen: Should get a comment from @jo-basevi. I recall the `payu` conda environments are Py3.9, but AFAICT that is because it is the default, and should be simple to specify a higher minimum version

https://github.com/ACCESS-NRI/payu-condaenv/blob/main/env.yml

There is an open issue in dev-docs about supported python versions, and I think we should have an organisation policy on what that should be, and a process to keep it updated.

https://github.com/ACCESS-NRI/dev-docs/issues/15

aidanheerdegen: I should have said above, I think we should shoot for Py3.10 as a guaranteed minimum python version.

test/test_um2netcdf.py
```diff
@@ -5,7 +5,209 @@
...
+                section, item = split_item_code(air_temp_cube.item_code)
+                air_temp_cube.attributes = {um2nc.STASH: DummyStash(section, item)}
+                m_iris_load.return_value = [air_temp_cube]
+
+                with mock.patch("iris.fileformats.netcdf.Saver") as m_saver:  # prevent I/O
+                    m_sman = mock.Mock()
+                    m_saver().__enter__.return_value = m_sman
+
+                    infile = "/tmp/fake_input_fields_file"
+                    outfile = "/tmp/fake_input_fields_file.nc"
```
marc-white, line 155: Switch to using tempfile

test/test_um2netcdf.py
```diff
@@ -5,7 +5,209 @@
...
+                    m_saver().__enter__.return_value = m_sman
+
+                    # TODO: fix lat/lon & levels requires c.coord attributes
+                    #       use fixtures to add attrs & remove the patches?
+                    with mock.patch("umpost.um2netcdf.fix_latlon_coord") as m_coord:
+                        with mock.patch("umpost.um2netcdf.fix_level_coord") as m_level:
+                            with mock.patch("umpost.um2netcdf.apply_mask") as m_apply_mask:
+                                with mock.patch("umpost.um2netcdf.cubewrite") as m_cubewrite:
+                                    infile = "/tmp/fake_input_fields_file"
+                                    outfile = "/tmp/fake_input_fields_file.nc"
```
marc-white, line 194: Switch to using tempfile

umpost/um2netcdf.py
```diff
@@ -40,6 +40,9 @@
 # TODO: what is this limit & does it still exist?
 XCONV_LONG_NAME_LIMIT = 110
 
+NC_FORMATS = {1: 'NETCDF3_CLASSIC', 2: 'NETCDF3_64BIT',
+              3: 'NETCDF4', 4: 'NETCDF4_CLASSIC'}
```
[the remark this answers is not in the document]
CodeGat: Further to your comment @marc-white, should you run `black` over this new code @truth-quark?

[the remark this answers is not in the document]
marc-white: @CodeGat as an aside, would be pleased to be pointed towards any standard ACCESS-NRI has in this regard (i.e., what should be run over code before merge)

test/test_um2netcdf.py
```diff
@@ -5,7 +5,209 @@
...
+                     18820.820244130424, 20246.59897992768, 21808.13663216417,
+                     23542.18357603375, 25520.960854349545, 27901.358260464756,
+                     31063.888598164976, 36081.76331548462, -1073741824.0], dtype=object)
+    return data
+
+
+@pytest.fixture
+def z_sea_theta_data():
+    # data ripped from aiihca.paa1jan.subset: ff.level_dependent_constants.zsea_at_theta
+    # TODO: dtype is object, should it be float?
```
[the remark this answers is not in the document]
truth-quark: For reference, I manually copied that array text from a REPL, manually opening the `aiihca.paa1jan.subset` file & looking at the var. `dtype=obj ` came from that, which seems weird. I'll examine `mule`'s source for an answer...

test/test_um2netcdf.py
```diff
@@ -5,7 +5,209 @@
...
+def mule_vars(z_sea_rho_data, z_sea_theta_data):
+    """Simulate mule variables from aiihca.paa1jan.subset data."""
+    d_lat = 1.25  # spacing manually copied from aiihca.paa1jan.subset file
+    d_lon = 1.875
+    return um2nc.MuleVars(um2nc.GRID_NEW_DYNAMICS, d_lat, d_lon, z_sea_rho_data, z_sea_theta_data)
+
+
+@pytest.fixture
+def air_temp_cube():
+    # copied from aiihca.paa1jan.subset file
```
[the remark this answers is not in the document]
truth-quark: Some sample outputs are in this dir: `gadi:/g/data/tm70/esm15_testing/sample_output/`.

The output subset file is here (intended as a smaller test file for quicker testing:
`gadi:/g/data/tm70/esm15_testing/sample_output/aiihca.paa1jan.subset`

As the project grows, I suspect a small data repo will be required to maintain file sets for testing.

test/test_um2netcdf.py
truth-quark: It can & `pytest`'s `tmp_path` fixture has similar functionality. The downside is that both touch the filesystem, which I'm attempting to avoid with the unit tests. I'm working to a 3 level testing strategy:
1. `Unit:` external dependencies are mocked, tests are fast quick feedback during dev
2. `Integration:` command line run(s) against a few data files (slower)
3. `Full:` process a collection of multiple real data files

I could replace the `/tmp/fake...` paths with nonsense paths that don't exist or will fail due to permissions should the mocks fail for some reason.

umpost/um2netcdf.py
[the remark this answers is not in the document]
truth-quark: > Super nit-picky, but makes it more legible, easier to add to/comment out later, etc.

Indeed, it's fixed in https://github.com/ACCESS-NRI/um2nc-standalone/pull/58/commits/d99d82e13afa5551158ad87cae90d975793d2f30.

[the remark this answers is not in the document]
truth-quark: > Further to your comment @marc-white, should you run `black` over this new code @truth-quark?

Probably, although I'll defer that in this PR. Running `black` on the test module created more of a _hideous freak show_ & reduced readability. I'm happy to temporarily have long lines while versioning etc is agreed upon, then revisit formatting. The current work is testing/fixing the code structure, with formatting & neatening a future concern. Does anyone have any comments on this approach?

test/test_um2netcdf.py
marc-white: Ah, right. So if I'm reading this correctly, the `mock.patch` calls above have disabled all of the I/O functionality, so you're just sending in fake filenames so there's something for `process()` to absorb.

I like the idea of replacing them with complete garbage though; as you mention, it provides a level of cover against mock failure (in particular, you won't try to play with a system's `/tmp` directory should the mocks fail out).

test/test_um2netcdf.py
marc-white: See above comment on L123-124.

test/test_um2netcdf.py
truth-quark: > Should get a comment from @jo-basevi. I recall the `payu` conda environments are Py3.9, but AFAICT that is because it is the default, and should be simple to specify a higher minimum version

I realised the parenthesised context manager block is limited to `test_um2netcdf.py`, **not** the workflow script. In which case, does `payu` ever run the unit tests for tools it calls in the workflow? (if not (& `gadi` envs are not an issue), we can probably treat the tests as 3.10+ & bask in modern Python joy...)

jo-basevi: I don't think payu runs unit tests for tools it calls in the workflow, so I think it's safe to use `3.10`+ for tests. Maybe newer versions of payu environments should move to using python 3.10

test/test_um2netcdf.py
truth-quark: > Ah, right. So if I'm reading this correctly, the `mock.patch` calls above have disabled all of the I/O functionality

Yep, code like the `mock.patch("mule.load_umfile")` call dynamically replaces the real `mule.load_umfile()` with a generic mock object that mostly just records calls & attr lookups made to it. The I/O func calls are redirected to the mock & the `process()` code continues obliviously :-)

It's not ideal, but the project imposes a few constraints where mocking helps.

> so you're just sending in fake filenames so there's something for `process()` to absorb. (Update: I missed the mock to `Saver` further up, c.f. L150-155).

Yep, the mock just "absorbs" the call, records the details & returns control.

A future step is refactoring the input I/O calls to a small func & moving `process()`'s core to func that **only** takes data args (thus avoids filesystem I/O). Once the core takes data only args, some mocking can be discarded as the tests should just provide raw data to process. Getting there requires a few evolutionary refactoring steps.

> I like the idea of replacing them with complete garbage though; as you mention, it provides a level of cover against mock failure (in particular, you won't try to play with a system's `/tmp` directory should the mocks fail out).

This is fixed in https://github.com/ACCESS-NRI/um2nc-standalone/pull/58/commits/f2c628c610629a2f459ea11ea1a64f8ef4fa67bb.

umpost/um2netcdf.py
marc-white: I really like it. It's cleaner, it removes non-controversial code out of the main workflow so it's easier to read, and typically makes debugging easier when something explodes (i.e., you get a granulated function/function output to look at, rather than a huge monolothic one).

umpost/um2netcdf.py
```diff
@@ -161,6 +168,7 @@ def fix_level_coord(cube, z_rho, z_theta):
                 c_sigma.var_name = 'sigma_theta'
 
 
+# TODO: split cube ops into functions, this will likely increase process() workflow steps
```
marc-white, line 171: What's your current idea for this? I'd recommend either:

- Keep `cubewrite` as a wrapper function, and break it up internally; or,
- Split it into 'data modification' and 'actual disk write' parts, and break those up internally as required.

I'd advise against showing too much of the 'sausage making' inside `process()` itself, if it can be avoided.

umpost/um2netcdf.py
marc-white: (Also, should have added I much prefer having a single object floating around, instead of a collection of assorted variable names to keep track of.)

umpost/um2netcdf.py
truth-quark: The outline for retro-fixing/testing `cubewrite()` is in line with `process()` rework:
* Extract its half dozen modification code blocks to separate functions & test
* Replace the code blocks with calls to the new functions
* Iteratively reduce `cubewrite()` to a unit testable size & add tests
* Refactor: cube modification calls into `process()`
  - Aim: refactor `cubewrite()` to a function that only takes cubes & writes to NetCDF
  - Gather cube modification steps into the `process()` workflow (improve proximity of modification funcs)
* `process()` is likely to become somewhat oversized
  - Refactor to extract sensible sub-workflows where needed (e.g. cube renaming, time repairs)
  - Gather these under "middle manager" sub-workflow functions & call these from `process()`

Ideally, `process()` will lay out the high level workflow. Mid layer functions will perform larger subtasks & the lowest level is task modification functions. The code currently doesn't have an obvious hierarchy, so it's less readable.

I'm assuming future ESM versions will be released, possibly requiring new or modified workflows. With a modular set of composition friendly components, new/adapted workflows can be built alongside existing ones.

umpost/um2netcdf.py
truth-quark: I added this error message to convert a comment to code, although I assumed `mule 2020` would be outdated by now. This is probably wrong if `mule 2020` versions are still mentioned in the `conda coecms` channel.

@MartinDix did you add the code comment about the `mule 2020` version not handling pathlibs properly?

test/test_um2netcdf.py
```diff
@@ -5,7 +5,217 @@
...
+                    #       use fixtures to add attrs & remove the patches?
+                    with mock.patch("umpost.um2netcdf.fix_latlon_coord") as m_coord:
+                        with mock.patch("umpost.um2netcdf.fix_level_coord") as m_level:
+                            with mock.patch("umpost.um2netcdf.apply_mask") as m_apply_mask:
+                                with mock.patch("umpost.um2netcdf.cubewrite") as m_cubewrite:
+                                    um2nc.process(fake_in_path, fake_out_path, std_args)
+
+                                    assert m_sman.update_global_attributes.called
+                                    assert m_sman.update_global_attributes.call_count == 2
+                                    assert m_saver.write.called is False  # write I/O prevented
```
blimlim, line 208: I'm not too clear on how mocking context managers works, but was wandering if the `.write` method be being called from the `m_saver` mock or the `m_sman` mock?

test/test_um2netcdf.py
```diff
@@ -5,7 +5,217 @@
...
+                                    assert m_coord.called
+                                    assert m_level.called
+                                    assert m_apply_mask.called
```
blimlim, line 211: Tiny idea: Wondering whether it's worth checking that `m_apply_mask` is twice as expected.

blimlim (changes requested):
Thanks for putting this together, I agree that it's not easy to put tests together for `process` and so it's nice to see them being added in!

I've added a few comments which are mostly small readability/clarity suggestions, though am not too fussed about other opinions on these.

The main thing that I'd like to get a second look at is the masking/filtering logic. From what I can understand, the results could be different with the refactored logic than the original logic in certain cases. I might not be correct here, but I think it would be worth to double check.

test/test_um2netcdf.py
truth-quark: As of the 6/8/2024 ESM1.5 standup, we'll support Python v3.10+ (see #11).

I'll do a few things:
* Reformat the unit tests
* Add a runtime warning if the minimum version isn't met (task in #11 )

The tests should also get cleaner & require less mocking as the modification functions are fixed up & can be called from `process()` by the tests.

truth-quark: I've fixed the multiple nesting issue in that func with  https://github.com/ACCESS-NRI/um2nc-standalone/pull/58/commits/ec45532cc02ae1694e98804994a03a8d76e770a9.

test/test_um2netcdf.py
truth-quark: Yep, that's fixed with https://github.com/ACCESS-NRI/um2nc-standalone/pull/58/commits/ec45532cc02ae1694e98804994a03a8d76e770a9, with other commits fixing the other test nesting.

This is a good step forward as future mock removal is a matter of removing a line of code. **Without** this, removing mocks requires adjusting indentation for the remaining test function body...

umpost/um2netcdf.py
```diff
@@ -300,17 +308,8 @@ def apply_mask(c, heaviside, hcrit):
...
     ff = mule.load_umfile(str(infile))
-
-    if isinstance(ff, mule.ancil.AncilFile):
-        raise NotImplementedError('Ancillary files are currently not supported')
-
-    # TODO: eventually move these calls closer to their usage
-    grid_type = get_grid_type(ff)
-    dlat, dlon = get_grid_spacing(ff)
-    z_rho, z_theta = get_z_sea_constants(ff)
+    mv = process_mule_vars(ff)
```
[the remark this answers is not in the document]
truth-quark: Yes, this should be renamed as it's misleading. It's not processing mule vars, it's accessing them.

It could be called `extract_model_levels()`, although is that still too non-descriptive?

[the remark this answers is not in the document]
blimlim: I'd been thinking mostly about the variable name `mv` rather than the function, though we could rename the function too. I think the function is extracting the model level information with the `z_rho` and `z_theta` outputs, but also information about the horizontal grid with the `grid_type`, `d_lat` and `d_lon` outputs. Would something like `extract_grid_info()` or `extract_grid_metadata()` work here?

test/test_um2netcdf.py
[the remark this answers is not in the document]
CodeGat: Fair enough!

umpost/um2netcdf.py
```diff
@@ -319,19 +318,17 @@ def process(infile, outfile, args):
...
-        check_pressure_warnings(need_heaviside_uv, heaviside_uv,
-                                need_heaviside_t, heaviside_t)
+    if do_mask is False:
+        cubes = list(no_masking_cubes(cubes, heaviside_uv, heaviside_t, args.verbose))
 
-    # TODO: can NC type be a single arg?
-    #       defer to new process() API
-    nc_formats = {1: 'NETCDF3_CLASSIC', 2: 'NETCDF3_64BIT',
-                  3: 'NETCDF4', 4: 'NETCDF4_CLASSIC'}
+        if not cubes:
```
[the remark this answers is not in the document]
truth-quark: Yep, that's part of the plan for fixing `process()` (perhaps in another PR to constrain this one). Here's an example of the desired logic (the tests still pass):
https://github.com/ACCESS-NRI/um2nc-standalone/blob/89fdc1512d662f3f8a57982befdefaa94c55df60/umpost/um2netcdf.py#L323-L358

umpost/um2netcdf.py
```diff
@@ -319,19 +318,17 @@ def process(infile, outfile, args):
...
 
-    do_mask = not args.nomask  # make warning logic more readable
+    do_mask = not args.nomask and ((need_heaviside_uv and heaviside_uv is not None) or
+                                   need_heaviside_t and heaviside_t is not None)
 
-    if do_mask:
-        # TODO: rename func to better name
-        check_pressure_warnings(need_heaviside_uv, heaviside_uv,
-                                need_heaviside_t, heaviside_t)
+    if do_mask is False:
```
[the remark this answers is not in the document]
truth-quark: I'm pretty sure there is a bug in the code, with the logic needing something like:
```python
do_mask = not args.nomask. # make easier to read
cannot_mask = ... # frig with logic

if do_mask:
    if cannot_mask:
        # user wants masking but is impossible - dump non-maskable cubes
        cubes = list(no_masking_cubes(cubes, heaviside_uv, heaviside_t, args.verbose))
else:
    # user wants to ignore masking, do not filter
    pass
```
I've got a rough demo test in to cover the switching off masking case. A fix needs the following:
* Test(s) for masking is off case
* Simplify the boolean logic & naming
* Likely other tweaks

umpost/um2netcdf.py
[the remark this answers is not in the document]
truth-quark: This is somewhat fiddly as the func does 2 things: extracting z sea data & grid metadata. I'm tempted to leave it as is, given splitting the func means changing data structures & another mock (or swapping to opening/reading test data files across several tests).

How's this hybrid naming idea: `model_levels = extract_grid_model_levels()`?
The generic `MuleVars` requires a rename too. Perhaps `ModelLevels = namedtuple(...)`?

test/test_um2netcdf.py
```diff
@@ -5,7 +5,217 @@
...
+    return "/tmp-does-not-exist/fake_input_fields_file"
+
+
+@pytest.fixture
+def fake_out_path():
+    # use junk paths to protect against accidentally touching filesystems
+    return "/tmp-does-not-exist/fake_input_fields_file.nc"
+
+
+def test_process_without_masking(air_temp_cube, precipitation_flux_cube, mule_vars,
```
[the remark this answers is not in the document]
truth-quark: Yep, the naming needs to clarify the 3 main states:
* Do masking & masking is possible
* Masking desired but not possible with missing `heaviside`, drop layers
* No masking

To suggest some names `test_process_mask_with_heaviside`, `test_process_masking_disabled` & `test_process_masking_no_heaviside_drop_cubes`. The last name needs some work...

test/test_um2netcdf.py
truth-quark: Ooops, that's a bug as the `write()` assert should be called on the mock sman. This is an example of mock fragility & testing implementation details. An improvement is `process()` returning all modified cubes for testing. This may not be a bad idea while the implementation & testing is in flux.

In the meantime, an intermediate fix with the correct mocks is in:  https://github.com/ACCESS-NRI/um2nc-standalone/pull/58/commits/386ff1da9f0a6b012f63ef34f7ceaf437084c8c0

blimlim: The intermediate fix looks good!

test/test_um2netcdf.py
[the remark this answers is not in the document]
blimlim: Those names sound good to me!

umpost/um2netcdf.py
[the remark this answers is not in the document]
blimlim: Ah I see, happy to leave as is for the moment in that case! I think a downside of something like `extract_grid_model_levels()` is that (to me) I interpret it as extracting the "grid's model levels" rather than "grid and model levels", both of which make sense but I think only the latter would be accurate here.

umpost/um2netcdf.py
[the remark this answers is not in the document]
blimlim: I think something like that should work!

test/test_um2netcdf.py
truth-quark: Oops, testing cubes can be handled by checking the data vars after the `process()` call, thus skipping the need to return anything.

umpost/um2netcdf.py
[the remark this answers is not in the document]
truth-quark: I've had a shot at cleaning `process()` masking logic & reworking the tests to focus on asserting against results (ignoring some implementation details). The tests are improving.

The work is in an evaluation branch: https://github.com/ACCESS-NRI/um2nc-standalone/tree/50/refactor-process-masking-fixmocks, so it's probably easiest to compare that with the https://github.com/ACCESS-NRI/um2nc-standalone/tree/50/refactor-process-masking.

I think the logic is fixed. The only point cubes are dropped is if masking is on AND the corresponding `heaviside_ut/t` is missing. `no_masking_cubes()` removes `uv & t` dependent layers.

@blimlim how readable do you find the logic checks now?

test/test_um2netcdf.py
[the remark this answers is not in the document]
truth-quark: These are changed in the masking logic fix evaluation branch, which I can merge if the suggested changes look good.

test/test_um2netcdf.py
```diff
@@ -5,7 +5,217 @@
...
+                                 std_args, fake_in_path, fake_out_path):
+    """Attempts end-to-end test of process(), ignoring cubes requiring masking."""
```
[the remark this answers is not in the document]
truth-quark: I've added some comments like this in the eval branch work...

test/test_um2netcdf.py
truth-quark: These mock asserts are likely to get chopped from the other eval branch work...

test/test_um2netcdf.py
truth-quark: @blimlim's work on #63 adds a `DummyCoordinate` data class to simulate `coord` attrs. When #63 is merged, I'll refactor the `test_process_...()` test setup to remove the `coord` mock & ensure the prod code is called.

Removing the `fix_level_coord()` & `cubewrite()` mocks can be handled in separate PRs (working on a model of gradually improving tests as `um2nc` is improved).

test/test_um2netcdf.py
truth-quark: Asserting an `apply_mask()` mock call has been removed...

test/test_um2netcdf.py
truth-quark: I'll resolve this as the code is deleted.

test/test_um2netcdf.py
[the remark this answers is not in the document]
truth-quark: Comments added for the test now known as `test_process_no_heaviside_drop_cubes()`.

test/test_um2netcdf.py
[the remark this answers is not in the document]
truth-quark: The evaluation work is merged, @blimlim do the new names make more sense now?

umpost/um2netcdf.py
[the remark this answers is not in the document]
truth-quark: The cleaner logic checks are now in & ready for re-review...

umpost/um2netcdf.py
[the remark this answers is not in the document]
truth-quark: I ended up fixing this for cleaner logic here: https://github.com/ACCESS-NRI/um2nc-standalone/blob/10f33e96a63c7999e90f0d60dc9611bfa3fd2fed/umpost/um2netcdf.py#L318-L331

umpost/um2netcdf.py
```diff
@@ -352,34 +349,61 @@ def process(infile, outfile, args):
...
+    Returns
+    -------
+    A MuleVars data structure.
+    """
+    if isinstance(fields_file, mule.ancil.AncilFile):
+        raise NotImplementedError('Ancillary files are currently not supported')
+
+    if mule.__version__ == "2020.01.1":
+        msg = "mule 2020.01.1 doesn't handle pathlib Paths properly"
+        raise NotImplementedError(msg)  # fail fast
```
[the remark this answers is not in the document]
truth-quark: Let's move this to a new, small issue as the problem ties in with environment setup issues.

test/test_um2netcdf.py
truth-quark: This is fixed with `process()` now returning a list of processed cubes (or empty list if everything is filtered out).

umpost/um2netcdf.py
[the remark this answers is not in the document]
truth-quark: I'll reference this discussion in #27, where it can be addressed as a minor issue.

umpost/um2netcdf.py
truth-quark: Tests not required as `non_masking_cubes()` is executed via `process()` tests.

blimlim (changes requested):
Changes are looking good to me! The masking logic looks like its doing what it should be, and I think is also easier to read.

From what I can understand, the tests look good too.

I just had a couple of small suggested changes to do with function names and docstrings which I think can now be updated to reflect changes in what the functions are doing.

test/test_um2netcdf.py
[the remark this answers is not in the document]
blimlim: New comments look good!

test/test_um2netcdf.py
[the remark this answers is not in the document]
blimlim: The new names look good to me!

umpost/um2netcdf.py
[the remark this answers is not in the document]
blimlim: New logic is looking good I think!

umpost/um2netcdf.py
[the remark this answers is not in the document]
blimlim: This looks good!

blimlim:
All tests passed on Gadi!

test/test_um2netcdf.py
```diff
@@ -5,7 +5,266 @@
...
+        m_saver().__enter__ = mock.Mock(name="mock_sman")
+        std_args.verbose = True  # test some warning branches
+
+        # air temp & geo potential should be dropped in process()
+        processed = um2nc.process(fake_in_path, fake_out_path, std_args)
+        assert len(processed) == 1
+        cube = processed[0]
+
+        assert cube is precipitation_flux_cube
+        assert cube.data is None  # masking wasn't applied
```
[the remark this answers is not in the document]
truth-quark: This is a somewhat contrived assertion for ensuring masking code wasn't called. I've added a comment to clarify this.

umpost/um2netcdf.py
```diff
@@ -462,52 +490,39 @@ def to_item_code(stash_code):
...
-            need_heaviside_t [bool], heaviside_t [iris cube or None])
-
+    TODO docs
     """
-    # Check whether there are any pressure level fields that should be masked.
-    # Can use temperature to mask instantaneous fields, so really should check
-    # whether these are time means
-    need_heaviside_uv = need_heaviside_t = False
+    # Check whether there are any pressure level fields that should be masked. Can use temperature
+    # to mask instantaneous fields, so really should check whether these are time means
```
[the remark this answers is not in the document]
truth-quark: Ooops, I only half fixed the function earlier. I've updated this to `get_heaviside_cubes()` & fixed the docs. This is a classic example of docstrings going out of date when only the code is refactored.

Fixed in https://github.com/ACCESS-NRI/um2nc-standalone/pull/58/commits/700c58d22fb3cb400e75d8eafd8509f48ce48754.

umpost/um2netcdf.py
```diff
@@ -352,33 +351,62 @@ def process(infile, outfile, args):
...
 
+    return cubes
+
+
+MuleVars = collections.namedtuple("MuleVars", "grid_type, d_lat, d_lon, z_rho, z_theta")
+
+
+def process_mule_vars(fields_file: mule.ff.FieldsFile):
+    """
+    Extract model levels with mule.
```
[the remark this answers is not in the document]
truth-quark: Yep! I've updated it in https://github.com/ACCESS-NRI/um2nc-standalone/pull/58/commits/5c6a28afd927ae3ac69f92c1247b4a9b2515426a.

test/test_um2netcdf.py
```diff
@@ -5,7 +5,266 @@
...
+
+
+def test_process_all_cubes_filtered(air_temp_cube, geo_potential_cube,
+                                    mule_vars, std_args,
+                                    fake_in_path, fake_out_path):
+    """Ensure process() exists early if all cubes are removed in filtering."""
```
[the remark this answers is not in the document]
truth-quark: Well spotted 1 char typo! It's fixed in https://github.com/ACCESS-NRI/um2nc-standalone/pull/58/commits/5c6a28afd927ae3ac69f92c1247b4a9b2515426a.

umpost/um2netcdf.py
```diff
@@ -530,24 +545,32 @@ def is_heaviside_t(item_code):
     return item_code == 30304
 
 
-def check_pressure_warnings(need_heaviside_uv, heaviside_uv, need_heaviside_t, heaviside_t):
+def non_masking_cubes(cubes, heaviside_uv, heaviside_t, verbose: bool):
     """
-    Prints warnings if either of heaviside uv/t are required and not present.
+    Yields cubes that do not require pressure level masking.
```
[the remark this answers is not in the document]
truth-quark: Good idea. I've updated the docstring in https://github.com/ACCESS-NRI/um2nc-standalone/pull/58/commits/bd78fd47b20b8c9a57b7d1a93c5f2522bbaede59.

umpost/um2netcdf.py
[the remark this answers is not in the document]
truth-quark: Let's defer this for work on #26.

truth-quark:
@blimlim I've made the changes, which definitely make the docs better.

Do you want to go over the fixes quickly & resolve any the sub-discussions you're happy with?

test/test_um2netcdf.py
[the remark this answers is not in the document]
blimlim: Ah awesome, the comment makes sense!

umpost/um2netcdf.py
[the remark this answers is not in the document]
blimlim: Updated docs look good!

umpost/um2netcdf.py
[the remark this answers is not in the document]
blimlim: I think that look good!

blimlim (approved):
Nice!

CodeGat (approved):
Had a re-review of the changes and it looks good. Note: I'm not a python expert so I might have missed some stuff.

truth-quark:
> Had a re-review of the changes and it looks good. Note: I'm not a python expert so I might have missed some stuff.

As it's an evolving project, omissions can be fixed :-). We're in a better position now with testing covering key code blocks.

test/test_um2netcdf.py
truth-quark: This is covered in #69.

umpost/um2netcdf.py
truth-quark: This is now for #70.

truth-quark:
The _freak show_ is merged, with 100+ comments...

PR #63, opened 2024-08-02
55/refactor fix_latlon_coord

### LINKED ISSUE #55: Refactor `fix_latlon_coord()` function
The `fix_latlon_coord()` function can be cleaned in line with the rest of the `un2nc-standalone` refactor.

Tasks:
* [ ] Move func lower/into call order in the module
* [x] Consider splitting to 2 funcs for `fix_latlon_coord_bounds` & `fix_latlon_coord_var_name`
* [x] Add unittests
* [x] Add docstring
* [x] Add constants for repeated strings
* [x] Bounds arrays should be constants
* [x] `_add_coord_bounds()` should drop leading `_`, it's a nested function, inaccessible as an "implementation detail"
* [x] Clean logic to reduce bounds checks
* [x] Fix up exception handling
  * [x] Move `try/except` from `process()` into `fix_latlon_coord()` (simplifies `process()` for readability)
  * [x] Replace `Exception` class with something more specific
  * [x] Update run script to handle alternate exception class?
* [x] Do the cube coords for lat/lon need to be looked up several times???
* [x] Convert if check calcs to plain functions (for readability & testing)
  * [x] Replace grid strings with existing constants
  * [x] Use descriptive var for values like `lon.points[0]`

blimlim:
This pull request closes #55. Apologies that there's a bit in here... It refactors the `fix_latlon_coord` function, which was used to add bounds, convert types, and add variable names for each cube's horizontal coordinates.

The main structural changes are to split the function into several more modular ones in order to help with readability and unit testing. Other changes to try and improve readability and testability include:
- Adding docstring
- Adding constants for repeated strings and preset bounds arrays
- Modifying exception handling so that it occurs within the `fix_latlon_coord` function, and raises a more specific exception.
- Update run script to handle alternate exception class?
- Removes redundant cube coordinate lookups
- Convert if check calcs to plain functions
- Use descriptive names for variables used in the checks

To help with the unit tests, the pull request adds a `DummyCoordinate` class to mimic iris cube coordinate objects with imitations of the `name` and `has_bounds` methods needed to test the new functions. It also adds an extension to the `DummyCube` class, `DummyCubeWithCoords` which can be used to add `DummyCoordinate` to a `DummyCube`.

Any suggestions about the structure, logic, tests, and anything else would be great!


test/test_um2netcdf.py
```diff
@@ -1,6 +1,8 @@
 import unittest.mock as mock
 from dataclasses import dataclass
 from collections import namedtuple
+import numpy as np
+from iris.exceptions import CoordinateNotFoundError
 
 import umpost.um2netcdf as um2nc
```
marc-white, line 7: Out of interest, how am I meant to get this working for testing? I can't `pip install -e` the umpost directory in its current state, so I'm getting test collection errors.

truth-quark: Whoops, due to prior history, this project has started oddly. Spencer is using a `gadi`'s environment & I'm reusing a local virtualenv from another project. There's been no setup docs until last week, when I grabbed docs from a related project. Do you feel like yet another PR review for docs? https://github.com/ACCESS-NRI/um2nc-standalone/pull/62
(these docs are missing `pip -e`)

marc-white: I do like me some good docs...

umpost/um2netcdf.py
```diff
@@ -92,46 +102,194 @@ def convert_proleptic(time):
...
+
+    Parameters
+    ----------
+    latitude_points: (array) 1D array of latitude grid points.
+    grid_type: (string) model horizontal grid type.
+    dlat: (float) meridional spacing between latitude grid points.
+    """
+    
+    min_latitude = latitude_points[0]
+    min_latitude_v_ND = -90.+0.5*dlat
```
truth-quark, line 195: Small styling thing: PyCharm's linter flags `ND` as non-lowercase. `min_lat_v_nd_grid` is an option, but somewhat  verbose. Perhaps `min_lat_nd_grid`?

umpost/um2netcdf.py
```diff
@@ -92,46 +102,194 @@ def convert_proleptic(time):
...
+    return (
+        (min_latitude == -90 and grid_type == GRID_END_GAME) or
```
truth-quark, line 197: `-90` could probably be converted to a constant to reduce magic numbers in the func.

umpost/um2netcdf.py
```diff
@@ -92,46 +102,194 @@ def convert_proleptic(time):
...
+
+def is_lat_river_grid(latitude_points):
+    """
+    Check whether latitude points are on the river routing grid.
+
+    Parameters
+    ----------
+    latitude_points: (array) 1D array of latitude grid points.
+    """
+    return len(latitude_points) == 180
```
truth-quark, line 168: `180` should probably be a constant to avoid magic numbers.

umpost/um2netcdf.py
```diff
@@ -92,46 +102,194 @@ def convert_proleptic(time):
...
+def is_lon_river_grid(longitude_points):
+    """
+    Check whether longitude points are on the river routing grid.
+
+    Parameters
+    ----------
+    longitude_points: (array) 1D array of longitude grid points.
+    """
+
+    return len(longitude_points) == 360
```
truth-quark, line 180: `360` should also be a constant to avoid magic numbers.

umpost/um2netcdf.py
```diff
@@ -92,46 +102,194 @@ def convert_proleptic(time):
...
+    ----------
+    longitude: (array) 1D array of longitude grid points.
+    grid_type: (string) model horizontal grid type.
+    dlon: (float) zonal spacing between longitude grid points. 
+    """
+    min_longitude = longitude_points[0]
+    min_longitude_u_ND = 0.5*dlon
+
+    return (
+        (min_longitude == 0 and grid_type == GRID_END_GAME) or
```
truth-quark, line 219: This block also has the `ND` name style & the odd magic number.

umpost/um2netcdf.py
```diff
@@ -92,46 +102,194 @@ def convert_proleptic(time):
...
+    )
+
+
+def is_lon_u_grid(longitude_points, grid_type, dlon):
+    """
+    Check whether longitude points are on the lon_u grid.
+
+    Parameters
+    ----------
+    longitude: (array) 1D array of longitude grid points.
```
truth-quark, line 211: PyCharm flagged the params docs needs `longitude` as `longitude_points`.

umpost/um2netcdf.py
```diff
@@ -92,46 +102,194 @@ def convert_proleptic(time):
...
+    if cube_coordinate.has_bounds(): 
+        # Don't change bounds if already exist
+        return
+    
+    if len(cube_coordinate.points) > 1:
+        cube_coordinate.guess_bounds()
     else:
-        lon.var_name = 'lon'
+        # For length 1, assume it's global. guess_bounds doesn't work in this case
+        cube_coordinate.bounds = GLOBAL_COORD_BOUNDS[coordinate_name]
```
truth-quark, line 251: The logic/flow can be somewhat simplified by inverting the `has_bounds()` check:
```Python
if not cube_coordinate.has_bounds(): 
    if len(cube_coordinate.points) > 1:
        cube_coordinate.guess_bounds()
    else:
        # For length 1, assume it's global. guess_bounds doesn't work in this case
        cube_coordinate.bounds = GLOBAL_COORD_BOUNDS[coordinate_name]
```
I prefer this new code to the original as you've improved overall readability.

umpost/um2netcdf.py
```diff
@@ -92,46 +102,194 @@ def convert_proleptic(time):
...
+
```
truth-quark, line 252: Minor formatting: there's an extra blank line here.

umpost/um2netcdf.py
```diff
@@ -92,46 +102,194 @@ def convert_proleptic(time):
...
+    Parameters
+    ----------
+    cube: Iris cube object (modified in place).
+    grid_type: (string) model horizontal grid type.
+    dlat: (float) meridional spacing between latitude grid points.
+    NB - Only applies to variables on the main horizontal grids, 
+    and not the river grid.
+    dlon: (float) zonal spacing between longitude grid points. 
+    NB - Only applies to variables on the main horizontal grids, 
+    and not the river grid.
```
truth-quark, line 270: I'd move the NB comments into the docstring body, to keep the params section clearer.

umpost/um2netcdf.py
```diff
@@ -92,46 +102,194 @@ def convert_proleptic(time):
...
+        latitude_coordinate.points = latitude_coordinate.points.astype(np.float64)
+        longitude_coordinate.points = longitude_coordinate.points.astype(np.float64)
+        
+        add_latlon_coord_bounds(latitude_coordinate)
+        add_latlon_coord_bounds(longitude_coordinate)
+        
+        fix_lat_coord_name(latitude_coordinate, grid_type, dlat)
+        fix_lon_coord_name(longitude_coordinate, grid_type, dlon)
+        
+    except iris.exceptions.CoordinateNotFoundError:
```
truth-quark, line 287: If the `cube.coord(name)` lookups are the only calls where `CoordinateNotFoundError` can occur, the `except` block can be moved to cover those 2 lines.

umpost/um2netcdf.py
```diff
@@ -92,46 +102,194 @@ def convert_proleptic(time):
...
+        warnings.warn(
+            "Missing latitude or longitude coordinate for variable (possible timeseries?): \n"
+            f"{cube}\n"
+        )
+        raise
```
truth-quark, line 292: There's a mixed error handling here with a warning & an exception. The original code shields the coord error behind an `Exception`, which is too generic.

Is suspected time series detection still considered a complete failure as `um2nc` doesn't handle time series?

To improve the error, how about:
```Python
except iris.exceptions.CoordinateNotFoundError as err:
    msg = "Missing latitude or longitude coordinate for variable (possible timeseries?): \n{cube}\n"
    raise PostProcessingError(msg) from err
```

truth-quark (changes requested):
Wow, there is much more testing that I initially thought!

For some initial thoughts: fixing `um2nc` is tricky on account of uncertain requirements (gathered from the code), retro fitting tests & that the processing code contains substantial nested boolean logic. The latter causes a combinational testing blowout (more than the the fix naming funcs).

In both the naming & lat/long fixes, our testing approach is blowing out in an attempt to cover most/all code paths. Part of the problem in being this thorough, is writing tests against some internal implementation details. Some mocks `test_fix_lat_coord_name()` & the process/masking fixes are a symptom of coupling to internal details.

Additionally, I assume writing the lat/long tests was tricky, which implies future maintenance/code changes could be harder (e..g changing processing code potentially breaks multiple tests). This could happen if cube modification requirements change.

Thus, I've been rethinking our testing strategy, helped with these TDD videos (~1 hour each):

Ian Cooper: “TDD, Where did it all go wrong?”
https://www.youtube.com/watch?v=EZ05e7EMOLM

Ian Cooper: “TDD Revisited” (more instructional)
https://www.youtube.com/watch?v=IN9lftH0cJc

One key message is focusing testing to external interfaces, not implementation details. This suggests focusing tests around `fix_latlon_coords()` & skipping direct testing of sub functions. Calling the sub-functions can be treated as an implementation detail, with the tests asserting correct end results. As `um2nc` cube mod functions often don't return anything, we'll have to make assertions on modified cube data, skipping assertions on mocks. Some mocking will be needed as an intermediate stop gap while the codebase is redesigned.

As a next step, I think we should try an experiment to see how this works, trying these steps:
* Temporarily comment out the grid type & bounds tests as an implementation detail
  - e.g. `test_is_lat/lon_river_grid`, `test_add_latlon_coord_bounds_has_bounds()`
* Test a common use case with a `test_fix_latlon_coords_type_change()` type function
  - Remove `mock` calls, so `fix_latlon_coords()` executes code for grids, bounds & coord naming fixes etc
  - Are more compound data fixtures needed to make a valid input?
* Try assertions on the modded cube (check bounds, data types etc)
* Run coverage tests to explore how this higher level testing covers the code

If the above is straightforward:
* configure a cube with different data to execute other branches
* run tests & re-analyse the coverage

Otherwise, if the testing is fiddly/tricky with setup or otherwise:

* adjust experiment, test against the mid level funcs `add_latlon_coord_bounds()` & `fix_..._coord_name()`
* explore the coverage...

Analysing higher level testing & coverage is probably a good pair dev exercise. It's also likely our code repair efforts will require starting with coupled tests, then iteratively refactoring the processing code & tests. Thus while things are harder now, it's temporary while iterating towards better design.

Hopefully this makes sense!

umpost/um2netcdf.py
blimlim: I think `min_lat_v_nd_grid` is a good idea – I've just updated it to this. For each of the grid types (`ND` or `EG`), there are two different latitude grids (the regular latitude points and the latitude_v points), and so keeping the `v` in the variable name is important for distinguishing the two.

`dlat` is the cell size (in the latitude direction), and so `0.5*dlat` is meant to end up as half of the cell size here.

umpost/um2netcdf.py
blimlim: Good idea. I've had a go at this but unsure of the best approach.

In the condition `min_latitude == -90`, the `-90` refers to the minimum latitude in the `end game` grid type's `latitude_v` grid, and so it would make sense to define `min_lat_v_eg_grid = -90`. 

For me, I think it reads clearest if we define this at the same time we define the minimum `lat_v` for the `new dynamics` grid `min_lat_v_nd_grid`. This has to be set within the function as it depends on the `dlat` argument, and so I've tried:

```python
    min_latitude = latitude_points[0]
    min_lat_v_nd_grid = -90.+0.5*dlat
    min_lat_v_eg_grid = -90
    return (
        (min_latitude == min_lat_v_eg_grid and grid_type == GRID_END_GAME) or
```
rather than setting it as a global constant. Let me know if this looks ok!

umpost/um2netcdf.py
blimlim: Good idea, I've just added the constant `NUM_LAT_RIVER_GRID`.

umpost/um2netcdf.py
blimlim: I've just added in a constant `NUM_LON_RIVER_GRID`.

umpost/um2netcdf.py
blimlim: I've renamed this to `min_lon_u_nd_grid` to match the format of the previous example, and added in a variable `min_lon_u_eg_grid` for the magic number.

umpost/um2netcdf.py
blimlim: Good point! Fixed up in latest commit.

umpost/um2netcdf.py
blimlim: Great suggestion. I think it helps a lot with readability – I've added this in in the latest commit.

umpost/um2netcdf.py
blimlim: Good catch – fixed up now.

umpost/um2netcdf.py
blimlim: Good idea, I've added this in

umpost/um2netcdf.py
blimlim: Good question. We expect the time series error to come up during automated post-processing of ESM1.5 simulations, as one of the model's output files contains time series. Whichever error is raised here, is the one that the conversion driver should allow to pass in:

https://github.com/ACCESS-NRI/um2nc-standalone/blob/1c86df606ca01eec677781dd2f99c10b26b07268/umpost/conversion_driver_esm1p5.py#L144-L150

I'm wondering whether it would make sense to add a new type of `PostProcessingError` here, e.g. `TimeSeriesError`, and then this would let us simplify the error handling in the driver script? Let me know what you think.

umpost/um2netcdf.py
truth-quark: This naming works equally as well as module level constants, given the vars still get descriptive names.

I also noticed the `return (...)` block is trickier to parse, with an if block being simpler:
```Python
...
min_lat_v_eg_grid = -90

if grid_type == GRID_END_GAME:
    return min_latitude == min_lat_v_eg_grid
elif grid_type == GRID_NEW_DYNAMICS:
    return np.allclose(min_lat_v_nd_grid, min_latitude)
```

umpost/um2netcdf.py
truth-quark: Small thing, should this be `NUM_LAT_RIVER_GRID_POINTS`, as in the number of points on the lat river grid?

umpost/um2netcdf.py
```diff
@@ -92,46 +105,194 @@ def convert_proleptic(time):
...
+    """
+    Add a 'var_name' attribute to a latitude coordinate object
+    based on the grid it lies on.
+
+    Parameters
+    ----------
+    lat_coordinate: coordinate object from iris cube (edits in place).
+    grid_type: (string) model horizontal grid type.
+    dlat: (float) meridional spacing between latitude grid points.
+    NB - Only applies to variables on the main horizontal grids,
```
truth-quark, line 118: Small thing, I noticed the "NB" in these param docs.

umpost/um2netcdf.py
```diff
@@ -92,46 +105,194 @@ def convert_proleptic(time):
...
+    """
+    Add a 'var_name' attribute to a longitude coordinate object
+    based on the grid it lies on.
+
+    Parameters
+    ----------
+    lon_coordinate: coordinate object from iris cube (edits in place).
+    grid_type: (string) model horizontal grid type.
+    dlon: (float) zonal spacing between longitude grid points. 
+    NB - Only applies to variables on the main horizontal grids,
```
truth-quark, line 145: There's another NB here too.

umpost/um2netcdf.py
truth-quark: This looks good as the try/except is really small & easy to follow.

Also, the `TODO` looks like it's left over from an earlier discussion.

umpost/um2netcdf.py
truth-quark: If it's simpler to handle, adding a new error `UnsupportedTimeSeriesError" (?) is fine. Error handling needs rework module wide, so even if this is a temp solution, it can be refined later.

umpost/um2netcdf.py
truth-quark: Does this need a `_POINTS` suffix as per the latitude constant suggestion?

umpost/um2netcdf.py
```diff
@@ -92,46 +105,194 @@ def convert_proleptic(time):
...
+    min_lon_u_nd_grid = 0.5*dlon
+    min_lon_u_eg_grid = 0
+
+    return (
+        (min_longitude == min_lon_u_eg_grid and grid_type == GRID_END_GAME) or
+        (
+            np.allclose(min_lon_u_nd_grid, min_longitude) and
+            grid_type == GRID_NEW_DYNAMICS
+        )
+    )
```
truth-quark, line 229: Here's another block where using an if/else is clearer.

umpost/um2netcdf.py
blimlim: Good catch, should be fixed now!

umpost/um2netcdf.py
blimlim: Should be fixed now!

umpost/um2netcdf.py
blimlim: Ah yep, I've removed that leftover `TODO`

umpost/um2netcdf.py
blimlim: Yep, good idea

umpost/um2netcdf.py
blimlim: Yeah great idea, that makes it easier to read/understand. I've added this in.

umpost/um2netcdf.py
blimlim: Good idea, I've added in an `UndupportedTimeSeriesError`

umpost/um2netcdf.py
blimlim: This one should be fixed now too!

blimlim:
I've had a go at switching over to more higher level testing for the `fix_latlon_coords` function rather than testing all the implementation details, and have a prototype in this branch (https://github.com/ACCESS-NRI/um2nc-standalone/blob/55/refactor-fix_latlon_coord-higher-level-testing/test/test_um2netcdf.py). 

There are a couple of parts I'm not certain on the approach I took, so it would be great to run through it and see if it makes sense.

In an attempt to make it easier to run a single end to end test of `fix_latlon_coords`, I added a bunch of extra things onto the `DummyCoordinate` class which aren't part of the standard `iris` cube coordinates. Specifically the `DummyCoordinates` now also hold the `MuleVars` that are used by the coordinate fixing functions, and also can hold attribute values expected at the end of the test:

https://github.com/ACCESS-NRI/um2nc-standalone/blob/00f503c4386f254ebefc476fca46e26c11088326/test/test_um2netcdf.py#L791-L823

https://github.com/ACCESS-NRI/um2nc-standalone/blob/00f503c4386f254ebefc476fca46e26c11088326/test/test_um2netcdf.py#L868-L889

And then the expected and processed values can be compared in the test:

https://github.com/ACCESS-NRI/um2nc-standalone/blob/00f503c4386f254ebefc476fca46e26c11088326/test/test_um2netcdf.py#L927-L982


I ended up sticking the expected values into the `DummyCoordinates` as I couldn't work out an alternative way of pairing the `DummyCoordinates`, auxiliary testing data (e.g. the `MuleVars`), and the expected values using something like `pytest.mark.parametrize`, especially if the `DummyCoordinates` are constructed as fixtures.

I think some downsides though are that the expected values are part of the objects that go through the processing, and so if something went wrong, the expected values could be modified. Also with this setup, each `DummyCoordinate` is set up very specifically for a single test, e.g:

```
@pytest.fixture
def lat_v_nd_dummy():
    # Dummy coordinate imitating imitating the real
    # lat_v grid from ESM1.5 (which uses the New Dynamics grid).

    d_lat = 1.25
    d_lon = 1.875
    lat_v_points = np.arange(-90.+0.5*d_lat, 90, d_lat, dtype="float32")

    mv = um2nc.MuleVars(um2nc.GRID_NEW_DYNAMICS, d_lat, d_lon, [], [])

    lat_v_nd_dummy_coord = DummyCoordinate(
        coordname=um2nc.LAT_COORD_NAME,
        points=lat_v_points,
        mulevars=mv,
        # Expected values after fix_latlon_tests
        expected_var_name="lat_v",
        expected_points_type=np.dtype("float64"),
        expected_bounds=DUMMY_GUESSED_BOUNDS
    )

    return lat_v_nd_dummy_coord
```
is only really usable for testing `fix_latlon_coords`, and not really anything else that might need a `DummyCoordinate`.


---

One idea was to revert the `DummyCoordinate`  to the simpler state where it just mimics an `iris` coordinate, without all the other testing data, and then add a `TestingCoordinate` class which holds both a `DummyCoordinate`, the corresponding `MuleVars`, and the expected values for a specific test. I think this would help with the above issues, but might involve adding lot's of fixtures.

Wondering whether you have any thoughts on the approach so far and the above ideas!

truth-quark:
> I've had a go at switching over to more higher level testing for the `fix_latlon_coords` function rather than testing all the implementation details, and have a prototype in this branch (https://github.com/ACCESS-NRI/um2nc-standalone/blob/55/refactor-fix_latlon_coord-higher-level-testing/test/test_um2netcdf.py).

I'll comment here & reference the alt branch, then all the comments are in 1 spot.

> There are a couple of parts I'm not certain on the approach I took, so it would be great to run through it and see if it makes sense.
> 
> In an attempt to make it easier to run a single end to end test of `fix_latlon_coords`, I added a bunch of extra things onto the `DummyCoordinate` class which aren't part of the standard `iris` cube coordinates. Specifically the `DummyCoordinates` now also hold the `MuleVars` that are used by the coordinate fixing functions, and also can hold attribute values expected at the end of the test:

Approaching fiddly problems with experimental code is completely fine & recommended in TDD. It's part of the cyclic process of _getting something to work, then refining it_. A few tweaks are possible to working towards a solution.

For coverage, even this one larger test touches quite a bit of the `fix_lat/lon` code, indicating higher level testing helps avoid coupling to implementation details. 

I'd recommend starting with the `um2nc.fix_latlon_coords()` function & working "backwards" to guide the minimum set of required inputs for test setup. It's also possible to simplify the testing & fixtures, with some iterative steps:
* `d_lat` & `d_lon` are defined per fixture & can be factored out to module level vars (shrinking the tests)
* There's a dependency to the `MuleVars` data struct, which isn't explicitly required for `um2nc.fix_latlon_coords()`
  - This hints at `MuleVars` being a candidate for eventual removal
  - `mv` is used to contain multiple vars for a function call
* Adding expected values to the `DummyCoord` changes the API, ideally something to avoid
* `test_fix_latlon_coords()` tests both `river` & `u/v` grid types
  - this looks to make the assertions harder, as the expected values have to be passed into the test

Does splitting `test_fix_latlon_coords()` into separate tests for `river` & `u/v` allow you to modify the test assertions to directly check expected values? e.g. `expected_var_name="lon_u"` becomes something like `assert cube.var_name == "lon_u"`. If this more specific testing works, can you remove the expected values from `DummyCoordinate`s & still have the tests pass?

Hopefully that makes sense!

blimlim:
Sweet, I've had a go at these changes in [the experiment branch](https://github.com/ACCESS-NRI/um2nc-standalone/blob/55/refactor-fix_latlon_coord-higher-level-testing/test/test_um2netcdf.py).

There are quite a few cases to test because of the two different `ENDGAME` and `NEW_DYNAMICS` grid types, and the different behaviours around adding bounds. I've tried to bundle any tests together which have the same expected results, so we have:

- One test for variables on the `standard` longitude and latitude grids, which tests both `ENDGAME` and `NEW_DYNAMICS` grid types:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/317beaed64acea8ad346423cbe72680d90465e8e/test/test_um2netcdf.py#L953-L962
- One test for variables on the `lat_v` and `lon_u` grids, which tests both `ENDGAME` and `NEW_DYNAMICS` grid types:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/317beaed64acea8ad346423cbe72680d90465e8e/test/test_um2netcdf.py#L896-L901
- One test for coordinates on the river grid, where the grid type doesn't matter:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/317beaed64acea8ad346423cbe72680d90465e8e/test/test_um2netcdf.py#L844-L846

- And then individual tests for the more specific coordinate bounds scenarios:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/317beaed64acea8ad346423cbe72680d90465e8e/test/test_um2netcdf.py#L1054
https://github.com/ACCESS-NRI/um2nc-standalone/blob/317beaed64acea8ad346423cbe72680d90465e8e/test/test_um2netcdf.py#L1019

As you mentioned, moving the expected values into the individual tests helps simplify `DummyCoordinate`, which now acts more like an iris coordinate:

https://github.com/ACCESS-NRI/um2nc-standalone/blob/317beaed64acea8ad346423cbe72680d90465e8e/test/test_um2netcdf.py#L664-L684

---

It would be great to know how these changes are looking to you. It does seem like to me there is quite a bit of repetition between the tests, though perhaps some checks don't need to be repeated across them all. The only thing we expect to be different between the `test_fix_latlon_coords_uv`, `test_fix_latlon_coords_river`, and `test_fix_latlon_coords_standard` tests are the `var_names`. Everything to do with the bound and point datatypes should be the same for each, and so I'm wondering if we should split those off into a separate test, rather than repeating the same bounds and datatype checks in each of the three above tests?


---
One other quick question: In order to get the tests with the `DummyCubeWithCoordinate` s to work, I had to comment out this part of the `DummyCube` definition:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/317beaed64acea8ad346423cbe72680d90465e8e/test/test_um2netcdf.py#L405-L408

From what I can understand, the only place this would be accessed during the tests of `process` would be if `apply_mask` was called, which I think would result in a key error. I just wanted to check that it's ok with you to remove this line though!

truth-quark:
Ok, there's a few points of test duplication in the experimental branch, which results in longer tests while potentially obscuring some core test logic. A few things can be tweaked to help reduce code & iteratively improve the tests.

The pre-modification assertion block appears 3 times:
```Python
# Checks prior to modifications.
assert cube_lat_coord.points.dtype == np.dtype("float32")
assert cube_lon_coord.points.dtype == np.dtype("float32")

assert not cube_lat_coord.has_bounds()
assert not cube_lon_coord.has_bounds()

assert cube_lat_coord.var_name is None
assert cube_lon_coord.var_name is None
```
To simplify, I'd refactor to a helper function like `def assert_something(lat_coord, lon_coord): ...` & call this in the tests.

Another code change is around the named vars. There's 3 assertion blocks like:
```Python
expected_lon_name = "lon_river"
expected_lat_name = "lat_river"
expected_dtype = np.dtype("float64")

# Checks post modifications.
assert cube_lat_coord.var_name == expected_lat_name
assert cube_lon_coord.var_name == expected_lon_name

assert cube_lat_coord.points.dtype == expected_dtype
assert cube_lon_coord.points.dtype == expected_dtype
```

These can be simplified by directly using common named "constants" :

```Python
DTYPE_FLOAT64 = np.dtype("float64")

def test_fix_latlon_something():
    # ... earlier test code

    # Checks post modifications.
    assert cube_lat_coord.var_name == um2nc.LAT_RIVER  # add & use constants
    assert cube_lon_coord.var_name == um2nc.LON_RIVER
    
    assert cube_lat_coord.points.dtype == DTYPE_FLOAT64
    assert cube_lon_coord.points.dtype == DTYPE_FLOAT64
```

There's also the `dlat`/`dlon` vars:

```Python
# Supplementary grid information used by fix_latlon_coords.
# Grid type, d_lat, d_lon don't matter for river grid, but must be set.
d_lat = D_LAT_N96
d_lon = D_LON_N96

# ... more of the test body

um2nc.fix_latlon_coords(cube_with_river_coords, grid_type, d_lat, d_lon)
```

As the tests using the `aiihca.paa1jan.subset` data, I think the `dlat`/`dlon` values are always the same, so we can use the "constants" directly: 
```Python
# Supplementary grid information used by fix_latlon_coords.
# Grid type, d_lat, d_lon don't matter for river grid, but must be set.

# ... the test body

um2nc.fix_latlon_coords(cube_with_river_coords, grid_type, D_LAT_N96, D_LON_N96)
```

As a step towards simplicity & reducing duplication, do you want to try these initial smaller fixes first?

I'm anticipating that refining the code & removing duplication will shrink the tests. This should make it easier to review logic across the tests & consider the approach to more complex elements (like the mocks).

blimlim:
Good ideas! I've added in those changes fro the pre test assertions and using the constants in the experiment branch. It definitely helps clean up the tests. 

A quick question I had about the constants – I was wondering whether it's ok with the datatype assertions to do:
```
 assert cube_lat_coord.points.dtype == np.dtype.("float64")
```
directly, rather than defining an extra constant and having 
```
 assert cube_lat_coord.points.dtype == DTYPE_FLOAT64
```

truth-quark:
> A quick question I had about the constants – I was wondering whether it's ok with the datatype assertions to do:
> 
> ```
>  assert cube_lat_coord.points.dtype == np.dtype.("float64")
> ```
> 
> directly, rather than defining an extra constant

Both ways work out the same, so it's more a matter of preference.

Since the dtype check occurs 3 times, it's worth replacing with a helper func to simplify some tests, e.g.:

```Python
def assert_dytpe_float64(cube_lat, cube_lon):
    assert cube_lat.points.dtype == np.dtype("float64")
    assert cube_lon.points.dtype == np.dtype("float64")
```

truth-quark:
I've experimented with using `iris` coordinate objects in other tests, which could be a viable solution for tests such as these over mocking them. There's an example here:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/53/fix-level-coord/test/test_um2netcdf.py#L682-L691

A starting point would be replacing a `DummyCoordinate` with `iris.coord.DimCoord` in a fixture & running the tests. With luck, all the `DummyCoordinate`s can be substituted & the mocks removed.

truth-quark:
I've experimented with removing `test_process_...()` mocks, replacing with some `coordinate`/`coordinate_model` setup:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/53/fix-level-coord-eval/test/test_um2netcdf.py

This seems to work on the surface, although testing ought to be expanded to ensure the model level variants are set up correctly with cubes that have model levels.

blimlim:
I've swapped over from the `DummyCoords` to directly using `iris.coords.DimCoord` objects. It looks like it works out which is great! And it helps reduces the complexity of setting up the tests.

Using the `DimCoords` lets us remove the mocks of the `guess_bounds()` method, though one change in the tests from this is that I'm no longer testing whether this method has been called, and instead just checking that bounds have been added:
```python
   assert_unmodified_coordinates(lat_coordinate, lon_coordinate)

    um2nc.fix_latlon_coords(cube_with_uv_coords, grid_type,
                            D_LAT_N96, D_LON_N96)

    assert lat_coordinate.var_name == um2nc.VAR_NAME_LAT_STANDARD
    assert lon_coordinate.var_name == um2nc.VAR_NAME_LON_STANDARD

    assert_dtype_float64(lat_coordinate, lon_coordinate)
    assert_has_bounds(lat_coordinate, lon_coordinate)
```
Do you think that's sufficient for the bounds testing, or would it be better to make sure the `guess_bounds` method is actually called perhaps by providing hard-coded expected bounds?

I ended up bringing everything into this branch to try and make it easier to discuss.

test/test_um2netcdf.py
```diff
@@ -651,6 +662,379 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+    lon_river_coord = iris.coords.DimCoord(
+        points=lon_river_points,
+        standard_name=um2nc.LONGITUDE
+    )
+    return lon_river_coord
+
+
+@pytest.fixture
+def lat_v_nd():
+    # iris DimCoord imitating imitating the real
```
truth-quark, line 689: There's a duplicate "imitating" here (PyCharm again).

test/test_um2netcdf.py
```diff
@@ -395,7 +403,10 @@ def __init__(self, item_code, var_name=None, attributes=None, units=None):
         self.units = None or units
         self.standard_name = None
         self.long_name = None
-        self.coord = {}
+        # TODO: Can I remove this... It breaks DummyCubeWithCoords
+        # It could be required if apply_mask is called during process
+        # tests. Would cause a KeyError anyway?
+        # self.coord = {}
```
truth-quark, line 409: The `self.coord = {}` is likely my error as `coord` is a lookup function in `iris`.

There's a few options for handling this:
* delete the line entirely
* replace with a `def coords(): raise NotImplementedError("coords needs implementation...")` in the class

I'd personally add `coords()` with an error, so any future tests using the method will fail fast. As there's a few dummy cube type objects across PRs, it's likely these will need refactoring into a couple of reusable DummyCubes which work across most/all of the tests. Other tests can subclass DummyCube to specialise the fake API too.

test/test_um2netcdf.py
```diff
@@ -651,6 +662,379 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
         assert ua_plev_cube.units == orig  # nothing should happen as there's no cube.units
 
 
+@pytest.fixture
+def lat_river():
+    # iris DimCoord imitating UM V7.3s river grid. Must have length 180.
```
truth-quark, line 667: Should the comment be tweaked to indicate this is on a half cell offset grid?
(and on line 678)

test/test_um2netcdf.py
```diff
@@ -651,6 +662,379 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
         assert ua_plev_cube.units == orig  # nothing should happen as there's no cube.units
 
 
+@pytest.fixture
+def lat_river():
```
truth-quark, line 666: For clarity, should the fixture name be tweaked to `lat_river_coord(s?)`?

test/test_um2netcdf.py
```diff
@@ -651,6 +662,379 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+    lat_river_points = np.arange(-90., 90, dtype="float32") + 0.5
+    lat_river_coord = iris.coords.DimCoord(
+        points=lat_river_points,
+        standard_name=um2nc.LATITUDE
+    )
+    return lat_river_coord
+
+
+@pytest.fixture
+def lon_river():
```
truth-quark, line 677: Also here, is `lon_river_coord(s?)` better?

test/test_um2netcdf.py
```diff
@@ -651,6 +662,379 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+    lon_river_points = np.arange(0., 360., dtype="float32") + 0.5
+    lon_river_coord = iris.coords.DimCoord(
+        points=lon_river_points,
+        standard_name=um2nc.LONGITUDE
+    )
+    return lon_river_coord
+
+
+@pytest.fixture
+def lat_v_nd():
```
truth-quark, line 688: Flagging fixture naming idea here too...

test/test_um2netcdf.py
```diff
@@ -651,6 +662,379 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+
+    lat_v_nd_coord = iris.coords.DimCoord(
+        points=lat_v_points,
+        standard_name=um2nc.LATITUDE,
+    )
+    return lat_v_nd_coord
+
+
+@pytest.fixture
+def lon_u_nd():
```
truth-quark, line 702: Flagging another fixture naming instance here...

test/test_um2netcdf.py
```diff
@@ -651,6 +662,379 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+
+    lon_u_nd_coord = iris.coords.DimCoord(
+        points=lon_u_points,
+        standard_name=um2nc.LONGITUDE,
+    )
+    return lon_u_nd_coord
+
+
+@pytest.fixture
+def lat_v_eg():
```
truth-quark, line 715: Flagging another fixture naming instance here...

test/test_um2netcdf.py
```diff
@@ -651,6 +662,379 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+
+    lat_v_eg_coord = iris.coords.DimCoord(
+        points=lat_v_points,
+        standard_name=um2nc.LATITUDE
+    )
+    return lat_v_eg_coord
+
+
+@pytest.fixture
+def lon_u_eg():
```
truth-quark, line 728: Flagging another fixture naming instance here...

test/test_um2netcdf.py
```diff
@@ -651,6 +662,379 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+
+    lon_u_eg_coord = iris.coords.DimCoord(
+        points=lon_u_points,
+        standard_name=um2nc.LONGITUDE
+    )
+    return lon_u_eg_coord
+
+
+@pytest.fixture
+def lat_standard_nd():
```
truth-quark, line 741: Flagging another fixture naming instance here...

test/test_um2netcdf.py
```diff
@@ -651,6 +662,379 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+@pytest.fixture
+def lon_standard_eg():
+    # iris DimCoord imitating the standard longitude
+    # grid from CM2 (which uses the End Game grid).
+    lon_points = np.arange(0.5*D_LON_N96, 360, D_LON_N96, dtype="float32")
+    lon_standard_eg_coord = iris.coords.DimCoord(
+        points=lon_points,
+        standard_name=um2nc.LONGITUDE
+    )
+    return lon_standard_eg_coord
```
truth-quark, line 793: I noticed each fixture creates a `DimCoord`, so there's 10 reps of this call with different args. The fixture section is ~130 lines. For a thought experiment, repetition could be cut down although it'd potentially rely on some smoke & mirrors decorators. If each fixture returned the data & name args, a decorator could instantiate & return `DimCoord`s. The tradeoff is cutting ~25 lines of code for more complexity.

test/test_um2netcdf.py
```diff
@@ -651,6 +662,379 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+
+
+class DummyCubeWithCoords(DummyCube):
+    # DummyCube with coordinates, which can be filled with
+    # iris.DimCoord objects for testing.
+    def __init__(self, dummy_cube, coords=[]):
```
truth-quark, line 799: There's a small warning, `coords=[]` supplies a default mutable arg. The simplest fix is changing the the `[]` to `None` & checking for that case in the coord loop below.

test/test_um2netcdf.py
```diff
@@ -651,6 +662,379 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+                         dummy_cube.attributes,
+                         dummy_cube.units)
+
+        # Set up coordinate dictionary to hold each dummy coordinate, with keys
+        # given by the coordinate names. This ensures that the key used to
+        # access a coordinate via the coord method matches the
+        # coordinate's name.
+        self.coordinate_dict = {}
+        for coord in coords:
+            self.coordinate_dict[coord.name()] = coord
```
truth-quark, line 811: Is this simpler as a dict comprehension?

test/test_um2netcdf.py
```diff
@@ -651,6 +662,379 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+
+    def coord(self, coordinate_name):
+        return self.coordinate_dict[coordinate_name]
+
+
+def assert_unmodified_coordinates(lat_coord, lon_coord):
```
truth-quark, line 817: Does renaming to `assert_coordinates_are_unmodified()` read more clearly to you?

test/test_um2netcdf.py
```diff
@@ -651,6 +662,379 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+    """
+
+    cube_with_river_coords = DummyCubeWithCoords(
+        dummy_cube=ua_plev_cube,
+        coords=[lat_river, lon_river]
+    )
+
+    # Supplementary grid information used by fix_latlon_coords.
+    # Grid type doesn't matter for river grid, but must be set.
+    grid_type = um2nc.GRID_END_GAME
```
truth-quark, line 856: Minor style thing, `grid_type` can be replaced & the constant used directly (+ reduce the test slightly)

test/test_um2netcdf.py
```diff
@@ -651,6 +662,379 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+        assert_dtype_float64(lat_coordinate, lon_coordinate)
+        assert_has_bounds(lat_coordinate, lon_coordinate)
+
+
+def test_fix_latlon_coords_single_point(ua_plev_cube):
+    """
+    Test that single point longitude and latitude coordinates
+    are provided with global bounds.
+    """
+    grid_type = um2nc.GRID_NEW_DYNAMICS
```
truth-quark, line 957: Minor thing, line 976 can use the constant directly.

test/test_um2netcdf.py
```diff
@@ -651,6 +662,379 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+    assert np.array_equal(lon_coord_single.bounds, expected_lon_bounds)
+
+
+def test_fix_latlon_coords_has_bounds(ua_plev_cube):
+    """
+    Test that existing coordinate bounds are not modified by
+    fix_latlon_coords.
+    """
+    # Grid information required to run fix_latlon_coords6
+    grid_type = um2nc.GRID_NEW_DYNAMICS
```
truth-quark, line 990: Ditto for minor fix with the grid type const here.

test/test_um2netcdf.py
```diff
@@ -651,6 +662,379 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+    Test that fix_latlon_coords raises the right type of error when a cube
+    is missing coordinates.
+    """
+    def _raise_CoordinateNotFoundError(coord_name):
+        # Include an argument "coord_name" to mimic the signature of an
+        # Iris cube's ".coord" method
+        raise iris.exceptions.CoordinateNotFoundError(coord_name)
+
+    # Following values don't matter for test. Just needed as arguments
+    grid_type = um2nc.GRID_NEW_DYNAMICS
```
truth-quark, line 1027: There's another minor style/const here.

test/test_um2netcdf.py
```diff
@@ -651,6 +662,379 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+
+    # Replace coord method to raise UnsupportedTimeSeriesError
+    ua_plev_cube.coord = _raise_CoordinateNotFoundError
+
+    with (
+        pytest.raises(um2nc.UnsupportedTimeSeriesError)
+    ):
+        um2nc.fix_latlon_coords(ua_plev_cube, grid_type, D_LAT_N96, D_LON_N96)
```
truth-quark, line 1035: Minor styling, this can fit into 2 lines & be <= 80 char line length.

umpost/um2netcdf.py
```diff
@@ -99,46 +126,193 @@ def convert_proleptic(time):
...
+    min_latitude = latitude_points[0]
+    min_lat_v_nd_grid = -90.+0.5*dlat
+    min_lat_v_eg_grid = -90
+
+    if grid_type == GRID_END_GAME:
+        return min_latitude == min_lat_v_eg_grid
+    elif grid_type == GRID_NEW_DYNAMICS:
+        return np.allclose(min_lat_v_nd_grid, min_latitude)
+    else:
+        return False  # TODO: Is this situation ever valid?
```
truth-quark, line 229: It's possible code could reach this point if a data file is supplied with a different staggering, although `get_grid_type()` should raise an exception to prevent this.

There's no harm in leaving it in as `False` is probably a more useful return value than `None`.

truth-quark:
> Using the `DimCoords` lets us remove the mocks of the `guess_bounds()` method, though one change in the tests from this is that I'm no longer testing whether this method has been called, and instead just checking that bounds have been added:
> 
> ```python
> ... snipped the code ...
> ```
> 
> Do you think that's sufficient for the bounds testing, or would it be better to make sure the `guess_bounds` method is actually called perhaps by providing hard-coded expected bounds?

I think we can skip checking `guess_bounds()` if is called inside the fix function. Also, the results of `guess_bounds()` are calculated from external code, so we can generally skip testing the exact details. If the dependencies are open source, we can get an idea if they have tests (`iris` does). The main cases I'd consider testing results from external code is if there's suspected risks or the dependency is flaky. 

I'm gradually adding these sort of ideas to the testing strategy doc here: https://github.com/ACCESS-NRI/um2nc-standalone/issues/65

truth-quark:
@blimlim Since there's mostly naming/small fixes under individual comments, do you want to fix those & then I'll do the approving review?

blimlim:
Sounds like a good plan :)

test/test_um2netcdf.py
blimlim: Just realised 
```python
    lat_points_middle = np.arange(-88.75, 89., D_LAT_N96)
    lat_points = np.concatenate(([-90],
                                 lat_points_middle,
                                 [90.]
                                 ), dtype="float32")
```
is exactly the same as `lat_points_middle = np.arange(-90, 91., D_LAT_N96)`...

test/test_um2netcdf.py
```diff
@@ -651,6 +669,355 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+    return DimCoord_maker
+
+
+@pytest.fixture
+@to_iris_DimCoord
+def lat_river_coord():
+    # iris DimCoord imitating UM V7.3s 1x1 degree river grid.
+    # Must have length 180.
+    lat_river_points = np.arange(-90., 90, dtype="float32") + 0.5
+    return lat_river_points, um2nc.LATITUDE
```
truth-quark, line 690: Wow, the decorator approach is less explicit but really neatens the fixtures. Nice!

test/test_um2netcdf.py
blimlim: Good catch, fixed up in latest version

test/test_um2netcdf.py
blimlim: I've gone with the `coords()` method which raises a `NotImplementedError` when called for the base `DummyCube`

test/test_um2netcdf.py
blimlim: Ah that's the `_u` and `_v` grids that are offset by half a grid cell. This one is on the "river" grid which is higher resolution (i.e. 1x1 degree rather than 1.25x1.875). I've updated the comments on each of the fixtures to clarify these things.

test/test_um2netcdf.py
blimlim: Good idea, I've updated all the coordinate fixture names.

test/test_um2netcdf.py
blimlim: New decorator approach is in!

test/test_um2netcdf.py
blimlim: Yep! Added in in latest commits

test/test_um2netcdf.py
blimlim: Yeah I think that's clearer, I've gone with `assert_coordinates_are_unmodified()`.

test/test_um2netcdf.py
blimlim: Good idea, using the constants directly now

umpost/um2netcdf.py
```diff
@@ -99,46 +126,193 @@ def convert_proleptic(time):
...
+
+    try:
+        latitude_coordinate = cube.coord(LATITUDE)
+        longitude_coordinate = cube.coord(LONGITUDE)
+    except iris.exceptions.CoordinateNotFoundError:
+        msg = (
+            "Missing latitude or longitude coordinate for variable (possible timeseries?): \n"
+            f"{cube}\n"
+        )
+        raise UnsupportedTimeSeriesError(msg)
```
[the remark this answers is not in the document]
blimlim: Good idea, this is added in.

blimlim:
I've added in those changes, which are ready for re-review. While putting them together, I noticed something important I'd neglected. 

The updates here swapped the type of exception raised when the `longitude` or `latitude` coordinates are missing:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/4e0d637f73dc2d17c325fe0c20a761b840fd3d79/umpost/um2netcdf.py#L300-L305

Where we've replaced a generic `Exception` with the new `UnsupportedTimeSeriesError`. 

The ESM 1.5 conversion driver was meant to except this error, and move onto the next file, but I'd forgotten to update the conversion driver's exception handling to use the new `UnsupportedTimeSeriesError`, meaning that if the updated code were to be run with an actual simulation, the conversion would crash as the conversion driver wasn't excepting the right type of error.

The tests in place in the conversion driver wouldn't catch this, as they use mocking to prescribe which error is raised by `um2nc.process`, and so the conversion driver and the mocked error were simultaneously using the wrong error type, allowing the tests to pass.

I've updated the tests and the exception handling to use the `UnsupportedTimeSeriesError`, however the same problem could easily happen again here. I'm not sure what would be the best way to guard against this... I think actually running `process` in the test would require a whole lot of complicated mocking and setup, and so I'm wondering whether this is something we could test with future integration tests. Do you think working out how to better test this could be done as a separate issue?

test/test_um2netcdf.py
```diff
@@ -395,7 +403,12 @@ def __init__(self, item_code, var_name=None, attributes=None, units=None):
         self.units = None or units
         self.standard_name = None
         self.long_name = None
-        self.coord = {}
+
+    def coord(self):
```
truth-quark, line 407: `coord()` needs a placeholder arg for the coord lookup name. Since we're ignoring it here, the signature can be `def coord(self, _):`, then `DummyCube` & `DummyCubeWithCoords` have the same method.

I suspect `DummyCube` & `DummyCubeWithCoords` can be merged in in a future refactor too.

test/test_um2netcdf.py
```diff
@@ -395,7 +403,12 @@ def __init__(self, item_code, var_name=None, attributes=None, units=None):
...
+        raise NotImplementedError(
+            "coord() method not implemented on DummyCube"
+        )
+
```
truth-quark, line 411: The linter complained about the extra line break here.

I've added #97 as a reminder for sorting out code auto formatting.

test/test_um2netcdf.py
```diff
@@ -651,6 +664,355 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
         assert ua_plev_cube.units == orig  # nothing should happen as there's no cube.units
 
 
+def to_iris_DimCoord(points_and_name_func):
+
+    def DimCoord_maker():
```
truth-quark, line 669: Minor style thing: the linter is complaining about capitalised letters in the function names.

test/test_um2netcdf.py
truth-quark: I did a quick experiment to check merging `DummyCube` & `DummyCubeWithCoords` here:
https://github.com/ACCESS-NRI/um2nc-standalone/commit/6cf2e325787d270dccb15292729adfd3d5dd9d78

Merging the classes was straightforward & the tests still passed. I'd defer the real refactor until outstanding PRs are merged, then the change can be assessed against all the affected tests...

test/test_um2netcdf.py
blimlim: Ah yep, fixed this up

test/test_um2netcdf.py
blimlim: Added this in!

test/test_um2netcdf.py
blimlim: Added in!

umpost/um2netcdf.py
```diff
@@ -99,46 +126,193 @@ def convert_proleptic(time):
...
+
+    try:
+        latitude_coordinate = cube.coord(LATITUDE)
+        longitude_coordinate = cube.coord(LONGITUDE)
+    except iris.exceptions.CoordinateNotFoundError as CoordError:
+        msg = (
+            "Missing latitude or longitude coordinate for variable (possible timeseries?): \n"
+            f"{cube}\n"
+        )
+        raise UnsupportedTimeSeriesError(msg) from CoordError
```
truth-quark, line 305: There's another styling thing here with the capitalised var name thanks to my earlier comment. I parsed that as a type, not a variable. How is `coord_error` instead?

test/test_um2netcdf.py
```diff
@@ -148,7 +156,7 @@ def test_process_no_heaviside_drop_cubes(air_temp_cube, precipitation_flux_cube,
 
         # TODO: lat/long & level coord fixes require more internal data attrs
         #       skip temporarily to manage test complexity
-        mock.patch("umpost.um2netcdf.fix_latlon_coord"),
+        mock.patch("umpost.um2netcdf.fix_latlon_coords"),
```
truth-quark, line 159: There's an opportunity to remove these mocks when the new & improved `fix_latlon_coords()` is merged. This could require modifying the test data, but the real function is called.

_Democking_ this case is probably better in another smaller PR.

test/test_um2netcdf.py
```diff
@@ -651,6 +663,355 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+        # coordinate's name.
+        if coords is not None:
+            self.coordinate_dict = {
+                coord.name(): coord for coord in coords
+            }
+        else:
+            self.coordinate_dict = {}
+
+    def coord(self, coordinate_name):
+        return self.coordinate_dict[coordinate_name]
```
truth-quark, line 799: I noticed the `coord()` lookup doesn't match the `iris` interface that raises the `CoordinateNotFoundError`. There's an example implementation here:

https://github.com/ACCESS-NRI/um2nc-standalone/blob/d8da6b1fe9ff8080f9c2f451bc28232a4316770e/test/test_um2netcdf.py#L705-L710

Once more PRs are merged, it's probably worth refactoring the multiple test cubes into 1-2 classes for reuse.

test/test_um2netcdf.py
```diff
@@ -651,6 +663,355 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+def assert_coordinates_are_unmodified(lat_coord, lon_coord):
+    """
+    Helper function to check that a coordinate's attributes match
+    those expected for a coordinate that has not yet been modified
+    by fix_latlon_coords.
+    """
+    for coord in [lat_coord, lon_coord]:
+        assert coord.points.dtype == np.dtype("float32")
+        assert not coord.has_bounds()
+        assert coord.var_name is None
```
truth-quark, line 811: Could the var name ever be an empty string?

test/test_um2netcdf.py
```diff
@@ -651,6 +663,355 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+
+
+def assert_dtype_float64(lat_coord, lon_coord):
+    assert lat_coord.points.dtype == np.dtype("float64")
+    assert lon_coord.points.dtype == np.dtype("float64")
```
truth-quark, line 816: I thought of a stylistic simplification like this:
```suggestion
def is_float64(lat_coord, lon_coord):  # becomes an is_something() bool function
    return (lat_coord.points.dtype == np.dtype("float64") and
            lon_coord.points.dtype == np.dtype("float64"))
```
Then the `assert` can be done explicitly lower down:
```python
assert is_float64(cube_lat_coord, cube_lon_coord)
```
This has the effect of ensuring the assertion blocks uniformly use `assert` as a keyword.

test/test_um2netcdf.py
```diff
@@ -651,6 +663,355 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+
+
+def assert_has_bounds(lat_coord, lon_coord):
```
truth-quark, line 819: This could also be switched to `has_bounds()` as per the stylistic fix for the float type check.

test/test_um2netcdf.py
```diff
@@ -651,6 +663,355 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+    lon_coord_single = iris.coords.DimCoord(points=np.array([0]),
+                                            standard_name=um2nc.LONGITUDE)
+
+    cube_with_uv_coords = DummyCubeWithCoords(
+        dummy_cube=ua_plev_cube,
+        coords=[lat_coord_single, lon_coord_single]
+    )
+
+    assert not lat_coord_single.has_bounds()
+    assert not lon_coord_single.has_bounds()
```
truth-quark, line 955: If the has bounds check helper is converted to a func returning a `bool`, this assertion can make use of it too.

test/test_um2netcdf.py
```diff
@@ -651,6 +663,355 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+        raise iris.exceptions.CoordinateNotFoundError(coord_name)
+
+    # Replace coord method to raise UnsupportedTimeSeriesError
+    ua_plev_cube.coord = _raise_CoordinateNotFoundError
+
+    with (
+        pytest.raises(um2nc.UnsupportedTimeSeriesError)
+    ):
+        um2nc.fix_latlon_coords(ua_plev_cube, um2nc.GRID_NEW_DYNAMICS,
+                                D_LAT_N96, D_LON_N96)
```
truth-quark, line 1012: This test can be simplified once the `CoordinateNotFoundError` fix from the earlier suggestion:
https://github.com/ACCESS-NRI/um2nc-standalone/pull/63/files#r1749685839

I think simply passing in a dummy cube should cause the required breakage, if the `lat/lon` coords are missing. There might be a duplicate of this test in another PR. I can't recall where, but it might allow this test to be deleted.

test/test_um2netcdf.py
blimlim: Possibly! This function is just to make sure that our tests are set up correctly, i.e. that we are starting off with a blank `var_name` so that the tests check whether the `var_name` actually gets changed. Is it worth having it as `assert coord.var_name is None or coord.var_name == ""` in this case?

test/test_um2netcdf.py
blimlim: Do you think it's clearer to do this, e.g:
```python
def has_bounds(lat_coord, lon_coord):
    return lat_coord.has_bounds and lon_coord.has_bounds()

...
def test_some_function():
...
assert has_bounds(lat_coord, lon_coord)
```
or to instead just go back to what we had previously and directly use the coordinates' `has_bounds` method in the tests:

```python
def test_some_function():
...
assert lat_coord.has_bounds()
assert lon_coord.has_bounds()
```
Let me know what you prefer!

test/test_um2netcdf.py
truth-quark: I did a quick test:
```python
cubes = iris.load("aiihca.paa1jan")
for cube in cubes:
    print({c.var_name for c in cube.coords()})
```
This emitted 193 `{None}`, although I don't know if other files have coordinates with a `var_name`, or if this is part of the API inherited (accidentally) from somewhere else. @MartinDix do you know why `iris` coordinate objects have a `var_name`?

If it's not required, we can leave the `None` assertion as is.

test/test_um2netcdf.py
truth-quark: The first option provides a degree of flexibility. Each test just has to assert the helper func. If the bool check changes, it can be fixed in 1 place & then applies to all dependent tests.

This is one of those fiddlier test cases where the iterations are less clear due to existing code, lib dependencies etc.

test/test_um2netcdf.py
blimlim: My understanding is that `var_name` is the name given to a variable in the output netCDF file, and so the tests here are checking that we've set up our dummy coordinates to start of without a `var_name`, and then check that `fix_latlon_coords` adds in the correct `var_name`.

umpost/um2netcdf.py
blimlim: Yeah good idea, fixed in 6a4c876

test/test_um2netcdf.py
blimlim: Fixed up in 6a4c876

test/test_um2netcdf.py
blimlim: I've added this in 6a4c876

test/test_um2netcdf.py
blimlim: Sounds good! Added in 6a4c876

test/test_um2netcdf.py
blimlim: Yeah this works! Added in 6a4c876

test/test_um2netcdf.py
blimlim: The test definitely becomes simpler after adding the `CoordinateNotFoundError`. Added in 6a4c876

truth-quark (approved):
Wow, that required substantially more effort than the initial skim indicated. Apologies!

This looks good to go & will help with the process of merging all the current PRs. @marc-white do you want to do a final check?

test/test_conversion_driver_esm1p5.py
```diff
@@ -130,29 +131,28 @@ def test_convert_fields_file_list_success(mock_process,
...
-        "TIMESERIES_ERROR"
-    ]
-    mock_process_with_exception(allowed_error_message)
+    # Test that UnsupportedTimeSeriesErrors are excepted.
+
+    # TODO: This test will not catch changes to the exceptions in um2nc. E.g.
+    # if decided to instead raise a RuntimeError in um2nc when timeseries
+    # are encountered, the conversion of ESM1.5 outputs would undesirably
+    # crash, but this test would say everything is fine, since we're
+    # prescribing the error that is raised.
```
marc-white, line 140: You could define the error type raised in this instance as a global variable in `conversion_driver_esm1p5.py`, and then import it for use in the test as the expected error type?

test/test_um2netcdf.py
```diff
@@ -651,6 +663,358 @@ def test_fix_units_do_nothing_no_um_units(ua_plev_cube):
...
+            }
+        else:
+            self.coordinate_dict = {}
+
+    def coord(self, coordinate_name):
+        try:
+            return self.coordinate_dict[coordinate_name]
+        except KeyError:
+            msg = f"{self.__class__}: lacks coord for '{coordinate_name}'"
+            raise iris.exceptions.CoordinateNotFoundError(msg)
```
marc-white, line 803: You should be able to drop the `iris.exceptions` from the front of this statement

test/test_um2netcdf.py
```diff
@@ -12,6 +14,9 @@
 import iris.cube
 import iris.coords
 
+D_LAT_N96 = 1.25
+D_LON_N96 = 1.875
```
marc-white, line 18: Could you comment what these numbers are for clarity?

truth-quark: Are these meant to be `DELTA_LAT_N96` & `DELTA_LON_N96` ?

umpost/um2netcdf.py
```diff
@@ -99,46 +126,193 @@ def convert_proleptic(time):
...
+                f"Expected {LATITUDE}."
+            )
+
+    if is_lon_river_grid(lon_coordinate.points):
+        lon_coordinate.var_name = VAR_NAME_LON_RIVER
+    elif is_lon_u_grid(lon_coordinate.points, grid_type, dlon):
+        lon_coordinate.var_name = VAR_NAME_LON_U
+    else:
+        lon_coordinate.var_name = VAR_NAME_LON_STANDARD
+
```
marc-white, line 185: Given these two functions have an almost identical structure, is it worth considering refactoring them as follows:
- A private helper function that does the actual work;
- Turn the existing functions into wrappers that call the private functions with the expecyed coordinate name, and comparison options for that coordinate?

umpost/um2netcdf.py
```diff
@@ -99,46 +126,193 @@ def convert_proleptic(time):
...
+    min_lon_u_nd_grid = 0.5*dlon
+    min_lon_u_eg_grid = 0
+
+    if grid_type == GRID_END_GAME:
+        return min_longitude == min_lon_u_eg_grid
+    elif grid_type == GRID_NEW_DYNAMICS:
+        return np.allclose(min_lon_u_nd_grid, min_longitude)
     else:
-        lon.var_name = 'lon'
+        return False  # TODO: Is this situation ever valid?
```
marc-white, line 251:
```suggestion
    
    return False  # TODO: Is this situation ever valid?
```

This has the same effect as the original code, but I think it's slightly clearer; it becomes more obvious that this function will return `False` unless once of the conditions above gets satisfied. It also prevents the `False` section of the `if` statement getting accidentally mangled if someone starts to add new conditions, or start thinking a condition should be applied to the `False` return when it's really the fall-back option.

umpost/um2netcdf.py
```diff
@@ -99,46 +126,193 @@ def convert_proleptic(time):
...
+    """
+    min_latitude = latitude_points[0]
+    min_lat_v_nd_grid = -90.+0.5*dlat
+    min_lat_v_eg_grid = -90
+
+    if grid_type == GRID_END_GAME:
+        return min_latitude == min_lat_v_eg_grid
+    elif grid_type == GRID_NEW_DYNAMICS:
+        return np.allclose(min_lat_v_nd_grid, min_latitude)
     else:
```
marc-white, line 228: See suggestion for L251 below

umpost/um2netcdf.py
```diff
@@ -99,46 +126,193 @@ def convert_proleptic(time):
...
+    Parameters
+    ----------
+    cube_coordinate: coordinate object from iris cube.
+    """
+    coordinate_name = cube_coordinate.name()
+    if coordinate_name not in [LONGITUDE, LATITUDE]:
+        raise ValueError(
+                f"Wrong coordinate {coordinate_name} supplied. "
+                f"Expected one of {LONGITUDE}, {LATITUDE}."
+            )
```
marc-white, line 268: Are we ever likely to get coordinate names other than `LATITUDE` and `LONGITUDE`? If so, it might be worth parameterizing out the 'set of coordinate names I expect' for use in this situation. You could also do it using `GLOBAL_COORDS_BOUNDS.keys()`.

marc-white (approved):
The comments I've made are only tweaks that you're free to think about and implement or not at your discretion - good job!

test/test_um2netcdf.py
blimlim: I tried to use similar naming here as in other parts of `um2nc`, eg [here](https://github.com/ACCESS-NRI/um2nc-standalone/blob/92096ce86f3cdd305570e940397b06633ae8777b/umpost/um2netcdf.py#L390) and [here](https://github.com/ACCESS-NRI/um2nc-standalone/blob/92096ce86f3cdd305570e940397b06633ae8777b/umpost/um2netcdf.py#L364)

I'm wondering would it be best to stick to that naming, and add a comment clarifying that these are grid spacings in degrees?

marc-white: Yes just go the comment.

truth-quark:
> The comments I've made are only tweaks that you're free to think about and implement or not at your discretion - good job!

As a general observation, I think staggering the 2 reviews has been beneficial. The later 2nd review brings fresh eyes after a few refinements. I'm pretty sure I inadvertently started to skim the code as it got more familiar.

umpost/um2netcdf.py
blimlim: I added in this error to try to guard against `add_latlon_coord_bounds` being inadvertently used on the wrong coordinates, and so here we should only be expecting `LATITUDE` or `Longitude`.

Would adding a constant eg `HORIZONTAL_COORD_NAMES = [LATITUDE, LONGITUDE]`, and then checking `if coordinate_name not in HORIZONTAL_COORD_NAMES` be a bit cleaner here?

umpost/um2netcdf.py
blimlim: I've put together a trial of this - let me know if it's along the lines of what you were thinking! The first attempt was using a private function:
```python

def _fix_horizontal_coord_name(coordinate, grid_type, grid_spacing, 
                               river_grid_check,
                               river_grid_name, staggered_grid_check,
                               staggered_name, base_name):
    if river_grid_check(coordinate.points):
        coordinate.var_name = river_grid_name
    elif staggered_grid_check(coordinate, grid_type, grid_spacing):
        coordinate.var_name = staggered_name
    else:
        coordinate.var_name = base_name



def fix_lat_coord_name(lat_coordinate, grid_type, dlat):
    _fix_horizontal_coord_name(coordinate = lat_coordinate,
                               grid_type=grid_type,
                               grid_spacing=dlat,
                               river_grid_check=is_lat_river,
                               river_grid_name=VAR_NAME_LAT_RIVER,
                               staggered_grid_check=is_lat_v_grid,
                               staggered_name=VAR_NAME_LAT_V,
                               base_name=VAR_NAME_LAT_STANDARD)
```

And another option was to put together a dictionary holding the specified checks:
```python
HORIZONTAL_GRID_NAMING_DATA = {
    LATITUDE: {
        "river_grid_check": is_lat_river_grid,
        "river_grid_name": VAR_NAME_LAT_RIVER,
        "staggered_name": VAR_NAME_LAT_V,
        "base_name":  VAR_NAME_LAT_STANDARD,
        "staggered_grid_check": is_lat_v_grid
    },
    LONGITUDE: {
        "river_grid_check": is_lon_river_grid,
        "river_grid_name": VAR_NAME_LON_RIVER,
        "staggered_grid_check": is_lon_u_grid,
        "staggered_name": VAR_NAME_LON_U,
        "base_name":  VAR_NAME_LAT_STANDARD,
    }
}


def fix_latlon_coord_name(coordinate, grid_type, grid_spacing):
    coord_name = coordinate.name

    if HORIZONTAL_GRID_NAMING_DATA["river_grid_check"](coordinate.points):
        coordinate.var_name = HORIZONTAL_GRID_NAMING_DATA["river_grid_name"]
    elif HORIZONTAL_GRID_NAMING_DATA["staggered_grid_check"](coordinate.points, grid_type, grid_spacing):
        coordinate.var_name = HORIZONTAL_GRID_NAMING_DATA["staggered_name"]
    else:
        coordinate.var_name = HORIZONTAL_GRID_NAMING_DATA["base_name"]
```
In terms of readability, I think I find the separate tests a bit easier to follow, though it is some amount of duplicate code. Happy to hear other opinions and ideas though!

test/test_um2netcdf.py
blimlim: Good idea! Fixed in 21dc396

test/test_um2netcdf.py
blimlim: Comments added in 21dc396

umpost/um2netcdf.py
blimlim: That's a good idea. I've added this in 21dc396

umpost/um2netcdf.py
blimlim: Also added in 21dc396

umpost/um2netcdf.py
marc-white: Precisely, although you've possibly already got the 'constant' available as `GLOBAL_COORDS_BOUNDS.keys()`.

truth-quark: Is `GLOBAL_COORDS_BOUNDS` likely to change, therefore adding keys that `coord name` should not be?

The advantage of the current line is that the test condition is obvious.

truth-quark: Also a minor formatting thing, lines 266-268 look over indented.

blimlim: Have updated the indentation in 37049c7

umpost/um2netcdf.py
truth-quark: Are these experiments on other branches?

While the code & tests work here (with a bit of duplication), refining the code can be deferred in favour of modularisation & test coverage. @blimlim What are your thoughts on adding an neatening task to issue #27, linking it to this discussion & returning to it later?

blimlim: That sounds like a good idea, I've added a task to #27.  The two experiments are up on the following branches:
https://github.com/ACCESS-NRI/um2nc-standalone/tree/55/refactor-fix_latlon_coord-private-name-func
https://github.com/ACCESS-NRI/um2nc-standalone/tree/55/refactor-fix_latlon_coord-dictionary-name-func

marc-white: Deferred action sounds like a good idea.

My original thought was like your option 2, but even more general - each coordinate could have an iterable of (check, name) tuples that you go through until you hit a match. The last tuple in the iterable would then be `(True, <default name>)`.

blimlim:
Thanks @truth-quark and @marc-white for the very helpful reviews! And apologies it took so much work on your end!

Issue #69, opened 2024-08-13
Refactor coord mocks when #63/lat long fixes is merged


truth-quark:
When #63 is merged, I'll refactor the `test_process_...()` test setup to remove the `coord` mock & ensure the prod code is called.

Removing the `fix_level_coord()` & `cubewrite()` mocks can be handled in separate PRs (working on a model of gradually improving tests as `um2nc` is improved).

_Originally posted by @truth-quark in https://github.com/ACCESS-NRI/um2nc-standalone/pull/58#discussion_r1713185730_

To clarify tasks:

Replace buggy implementation with `DummyCube` with no coord attrs:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/120dc47d4a3c131c18c3f1030167ff29abcb37b3/test/test_um2netcdf.py#L73-L79

Remove these mocks & the tests should pass:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/120dc47d4a3c131c18c3f1030167ff29abcb37b3/test/test_um2netcdf.py#L160

truth-quark:
From exploratory work on branch `69/refactor-cube-mocks`:

Before:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/120dc47d4a3c131c18c3f1030167ff29abcb37b3/test/test_um2netcdf.py#L73-L94

After:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/43a541e11db4d86b75ff2aad05baeddb0669dc07/test/test_um2netcdf.py#L73-L78

Issue #15, opened 2024-06-19
Refactoring: split apply_mask()


truth-quark:
The `apply_mask()` function has 2 modes, depending if the `cube` shapes match or not. To facilitate unit testing, split into a matching & mismatching shape variants for more specific unit testing.

Any calling code will have to check the cube shapes, but this can be placed in a check function to increase readability.

PR #72, opened 2024-08-14
Testing/repairing `apply_mask()`

truth-quark:
This PR is for initial work on retrofitting unit tests to `apply_mask()`.

As of the first few commits, the work is incomplete. Therefore, this PR is intended for discussions around design & retrofitting testing to iteratively improve the retro-fix approach. 

Interlinked problems exist with retro-fitting testing to this codebase:
* Production code depends upon `iris.cube.Cube` objects.
* `iris.cube.Cube`s need to be read from files
* `iris.cube.Cube`s are complex data structures, with production code accessing multiple `Cube` attrs
* Avoiding I/O (e.g. loading UM files) in low level unit tests
* Simulating `Cube`s wtih mocks requires tricky/fragile work to build nested data structures

**Fixes Required:**

`test_apply_mask_with_matching_grid_extents()`:
* the final assertions are weak & incomplete, only checking for data modification. Does this need more detailed tests to ensure the algorithm works as expected?
* Use `np.where()`?

`test_apply_mask_with_non_matching_grid_extents()`:
* relies on fragile mocking to synthesise cube attrs. Is there a way to create/ a synthetic cube from real data & skip mocking?
* also has weak assertions

### DIFF
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -1,4 +1,7 @@
 import unittest.mock as mock
+
+import math
+import copy
 from dataclasses import dataclass
 from collections import namedtuple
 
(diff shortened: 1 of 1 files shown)


truth-quark:
IDEA: instead of mocking `coords`, can the [iris.cube.Coord](https://scitools-iris.readthedocs.io/en/stable/generated/api/iris.coords.html#iris.coords.Coord) subclasses be used directly?

truth-quark:
Testing & cleaning `apply_mask()` needs cubes with coordinates & other attributes. This will be easier to complete when the changes in #111 & #69 are merged (make use of existing test fixtures).

atteggiani:
I think this PR will not be added in the first release, and we might think about ways to restructure `um2netcdf.py` to make itself and testing easier and cleaner.
Therefore, I would close it without merging. 
What do you think @blimlim ?

blimlim:
I think so, let's close this for now – I think testing the `apply_mask()` function would become easier in the future if we swap to using actual iris cubes for the tests (i.e. #150). Let's leave the relevant issue #15 open though

PR #74, opened 2024-08-15
Add conda packaging and deployment to accessnri channel

### LINKED ISSUE #71: Add conda packaging and deployment to accessnri channel
We need conda packaging for `um2nc` to add it to the payu conda environments that are deployed to gadi

https://github.com/ACCESS-NRI/payu-condaenv/issues/34

and it needs to be deployed to the accessnri conda channel.

There is some discussion about what packages are required in this PR

https://github.com/ACCESS-NRI/um2nc-standalone/pull/62

aidanheerdegen:
Fixes #71

I pretty much just copied what @dougiesquire had done on the payu repo.


conda/meta.yml
```diff
@@ -0,0 +1,49 @@
...
+
+test:
+    imports:
+        - umpost
+    requires:
+        - pytest
+    source_files:
+        - test/**
+    commands:
+        - py.test
```
jo-basevi, line 42: Is there a reason to use `py.test` over `pytest`? I hadn't heard of `py.test` before and from what I can tell, it seems like an older version of `pytest`

jo-basevi (commented):
Changes look good to me, just a couple of small things. It could be good to add a CI to test the conda build step? (e.g. in payu:  https://github.com/payu-org/payu/blob/e611bcca30a95be96903821d8a083baf2c8cdde7/.github/workflows/CI.yml#L38-L56)

.github/workflows/CD.yml
```diff
@@ -0,0 +1,40 @@
+name: CD
+
+on:
+  push:
+    branches:
+        - master
```
CodeGat, line 6:
```suggestion
        - main
```
Changed to the `main` branch.
Also the default branch for this repository is `develop` not `main`, so the default pull request target is `develop` also - you might have to change the target!

.pylintrc
```diff
@@ -0,0 +1,5 @@
+[MAIN]
+
+# Files or directories to be skipped. They should be base names, not
+# paths.
+ignore=_version.py
```
CodeGat, line 5: Is the `_version.py` file autogenerated?

conda/meta.yml
```diff
@@ -0,0 +1,49 @@
...
+
+about:
+    home: https://github.com/access-nri/um2nc-standalone
+    license: Apache Software
+    license_family: Apache
```
CodeGat, line 47: In future we should add a `LICENSE` file to the repo.

CodeGat (dismissed):
Mostly just the target branch of this pull request.

.github/workflows/CD.yml
aidanheerdegen: @jo-basevi pointed out we should be deployed based on tagging, which I agree with, so this is now moot, but thanks for spotting.

.pylintrc
aidanheerdegen: Yep. `versioneer` generates it.

conda/meta.yml
aidanheerdegen: Agreed!

aidanheerdegen: Added here and associated metadata.

conda/meta.yml
aidanheerdegen: No reason. I just copied from another config. Thanks.

.github/workflows/CD.yml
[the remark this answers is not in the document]
aidanheerdegen: Good point. Agreed, changed.

aidanheerdegen:
I could change the target to `main`, or keep on `develop` and merge `develop` with `main` ASAP.

aidanheerdegen:
Seems the CI interprets the python version as a number, so "3.10" becomes 3.1.

https://github.com/ACCESS-NRI/um2nc-standalone/actions/runs/10399122598/job/28797544390?pr=74#step:3:4

I've quote the strings, and they seem to work ok in the intake catalogue CI, which looks the same to my eye

https://github.com/ACCESS-NRI/access-nri-intake-catalog/actions/runs/10295656976/job/28495683756#step:3:4

Any suggestions/help would be welcome.

CodeGat (dismissed):
Yeah gotta quote those strings! That should be fine as is

CodeGat:
I'm not sure I understand the failing tests, though...

truth-quark:
The primary dev branch is temporarily set to `develop` while there's a lot of change activity.

jo-basevi:
I am not sure why the conda build is [failing](https://github.com/ACCESS-NRI/um2nc-standalone/actions/runs/10399194469/job/28797724155?pr=74), my initial guess was there's no git tag versions for versioneer so the meta.yaml isn't valid but I think versioneer might give a default? I wonder if it's worth adding a `--debug` flag to the conda build cmd `conda build . --no-anaconda-upload`  and hope it gives some more information?

jo-basevi:
In payu, the filename is `meta.yaml` vs `meta.yml`? The conda docs use `meta.yaml` - https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html. This stack-overflow issue run into issues with conda build ignoring the `meta.yml` file (https://stackoverflow.com/questions/73405205/conda-build-ignoring-meta-yml). I don't know if that would make a difference

aidanheerdegen:
> In payu, the filename is `meta.yaml` vs `meta.yml`?

Heh. Yeah, thanks @jo-basevi. I was sure I'd copied that from somewhere else, but now I'm not so sure. `meta.yml` seems to work in some cases. `conda` certainly complained when I had both files present, which is kinda weird.

So yes that was the proximate error of not being able to find a recipe file.

aidanheerdegen:
> I am not sure why the conda build is [failing](https://github.com/ACCESS-NRI/um2nc-standalone/actions/runs/10399194469/job/28797724155?pr=74), my initial guess was there's no git tag versions for versioneer so the meta.yaml isn't valid but I think versioneer might give a default? I wonder if it's worth adding a `--debug` flag to the conda build cmd `conda build . --no-anaconda-upload` and hope it gives some more information?

Again most excellent advice. I was testing locally with those commands, and yeah I wonder if we should just add the `--debug` flag so there is some more verbose output if something goes awry.

aidanheerdegen:
Now the CI tests complain because `um2nc` isn’t `pip install`-able without the `mule` conda dependency being installed in the environment. 

An elegant solution is to save the package from the `conda-build` stage as an artefact and `conda install` it in the test conda env, which will pull down the `conda` package dependencies.

@CodeGat can you point me to an example I can shamelessly copy to achieve this?

aidanheerdegen:
So all the infra stuff is working, now failing on the tests

https://github.com/ACCESS-NRI/um2nc-standalone/actions/runs/10414392571/job/28843293751?pr=74#step:8:12

Will consult with @truth-quark, but am considering having allowed-fails for these and fix them later when the interface if stabilised.

aidanheerdegen:
There was an import of `um2netcdf4` that is left over crufty transition to new infra. This was fixed/avoided in https://github.com/ACCESS-NRI/um2nc-standalone/pull/57 so I rebased and force-pushed.

CodeGat:
SUCCESS!!!!!!!! 🥳

aidanheerdegen:
Well hello there passing tests, looking pretty good in your slinky green ticks ...

CodeGat (approved):
Excellent stuff

aidanheerdegen:
@truth-quark is going to take a look on Monday, so will wait on that for merging.

aidanheerdegen:
I cleaned up commit history and force-pushed ready for merging on Monday.

truth-quark:
> I cleaned up commit history and force-pushed ready for merging on Monday.

I'm not a CI specialist, but will review what I can...

.github/workflows/CD.yml
```diff
@@ -0,0 +1,40 @@
+name: CD
```
truth-quark, line 1: Should we have a more descriptive name like `um2nc CD`?

.github/workflows/CI.yml
```diff
@@ -0,0 +1,109 @@
+# Run CI tests
+name: CI
```
truth-quark, line 2: Should we have a more descriptive name like `um2nc CI`?

.github/workflows/CI.yml
```diff
@@ -0,0 +1,109 @@
...
+
+# Controls when the action will run.
+on:
+  # Triggers the workflow on push or pull request events but only for the master branch
+  push:
+    branches: [ master, develop ]
```
truth-quark, line 8: This repo has a `main`, but not `master`. Does `main` need to be included in these lists?

.github/workflows/CD.yml
```diff
@@ -0,0 +1,40 @@
...
+          show-channel-urls: true
+
+      - name: Build and upload the conda package
+        uses: uibcdf/action-build-and-upload-conda-packages@c6e7a90ad5e599d6cde76e130db4ee52ad733ecf
+        with:
+          meta_yaml_dir: conda
+          python-version: ${{ env.PY_VERSION }}
+          user: accessnri
+          label: main
+          token: ${{ secrets.anaconda_token }}
```
truth-quark, line 40: Minor thing, GitHub is flagging a missing EOF newline...

.github/workflows/CI.yml
```diff
@@ -0,0 +1,109 @@
...
+        shell: bash -l {0}
+        run: python -m pytest --cov=umpost --cov-report=xml -s test
+
+      - name: Upload code coverage
+        uses: codecov/codecov-action@v4
+        # Only upload once for the installed version
+        if: matrix.python-version == ${{ env.PY_VERSION }}
+        with:
+          token: ${{ secrets.codecov_token }}
+          files: ./coverage.xml
```
truth-quark, line 109: Minor thing, GitHub is flagging a missing EOF newline here too...

.github/workflows/CD.yml
CodeGat: That's probably covered because we're in the `um2nc-standalone` repo, I think

.github/workflows/CI.yml
```diff
@@ -0,0 +1,109 @@
...
+        run: |
+          conda install -c conda-forge -c accessnri -c coecms -c file://${{ env.ARTIFACT_LOCATION }} um2nc
+
+      - name: List installed packages
+        shell: bash -l {0}
+        run: conda list
+
+      - name: Lint
+        shell: bash -l {0}
+        run: pylint --extension-pkg-whitelist=netCDF4 --ignored-modules=um2netcdf4,umpost -E umpost
```
truth-quark, line 97: Given the earlier import problem was with `um2netcdf4` (the `gadi` version), does `umpost` still need to be skipped by the linter?

truth-quark (approved):
I've looked over the config & only found a few minor things. Otherwise, nothing stands out given I'm not a CI/CD expert. It's possible some more nuanced issues could be overlooked...

.github/workflows/CD.yml
aidanheerdegen: I'm not sure it is necessary as `um2nc` is our only "product" being deployed from the repo. In general we've not done this for other repositories.

.github/workflows/CI.yml
aidanheerdegen: I'm not sure it is necessary as `um2nc` is our only "product" being checked in the repo.

Any opinion @CodeGat?

.github/workflows/CI.yml
aidanheerdegen: Yeah that's a typo. Thanks. Fixed.

.github/workflows/CD.yml
aidanheerdegen: Fixed.

.github/workflows/CI.yml
aidanheerdegen: I got an error without specifying it. It's only skipping the `import`. I believe it's still linting all of the code. Happy to be corrected if I'm wrong.

.github/workflows/CI.yml
aidanheerdegen: Fixed.

aidanheerdegen:
@truth-quark can you check you're happy with my responses to your queries. If you're satisfied please mark the unresolved conversations as resolved and if you could provide another approval as I've pushed some changes based on your review.

Thanks!

.github/workflows/CI.yml
aidanheerdegen: Heh. I didn't see your previous comment @CodeGat. I'll assume you're happy  with this too and resolved.

CodeGat (approved):

.github/workflows/CI.yml
truth-quark: I'm not sure from the docs:
https://pylint.pycqa.org/en/latest/user_guide/configuration/all-options.html#ignored-modules

"List of module names for which member attributes should not be checked and will not be imported". From some docs & stack overflow, it looks like the import is prevented, which I assume prevents linting. Has anyone else used pylint in detail?

truth-quark: Another option is to add `# pylint: disable=import-error`, so `conversion_driver_esm1.5.py` would have:

```Python
if hostname.startswith("gadi"):
    import um2netcdf4 # pylint: disable=import-error
```
How does that sound?

.github/workflows/CI.yml
```diff
@@ -0,0 +1,113 @@
...
+
+      - name: Setup conda environment
+        uses: conda-incubator/setup-miniconda@v3
+        with:
+          miniconda-version: "latest"
+          python-version: ${{ matrix.python-version }}
+          environment-file: conda/environment.yml
+          activate-environment: um2nc
+
+      - name: DEBUG - where are things
```
CodeGat, line 83: Whoops, we can remove this DEBUG section now :laughing:

.github/workflows/CI.yml
truth-quark: This is now deferred in #79.

truth-quark:
> @truth-quark can you check you're happy with my responses to your queries. If you're satisfied please mark the unresolved conversations as resolved and if you could provide another approval as I've pushed some changes based on your review.

Good to go! The only outstanding conversation is @CodeGat's comment on the debug section.

.github/workflows/CI.yml
aidanheerdegen: Yes that is another option, but requires code changes. Happy for you to take that route if you want to.

Should probably make that another issue/PR, remove the exception in the CI, make it fail and then add the fix in the code to make it pass again.

aidanheerdegen:
If we want nice entry points, e.g. the ability to just call `um2nc` from the command line, then we'll need `main` or similar functions defined, as entry points need to point at a function.

https://packaging.python.org/en/latest/specifications/pyproject-toml/#entry-points

This looks like this in the `pyproject.toml` script:
```toml
[project.scripts]
um2nc = "umpost.um2netcdf:main"
```

When I try and omit the function specifier
```toml
[project.scripts]
um2nc = "umpost.um2netcdf"
```
it throws this error:
```
ERROR: For req: um2nc==0+untagged.179.g0e246cc.dirty. Invalid script entry point: <ExportEntry um2nc = umpost.um2netcdf:None []> - A callable suffix is required. Cf http
s://packaging.python.org/specifications/entry-points/#use-for-scripts for more information.
```

I've tested `__main__` as the target function, but that also doesn't work with the current set up.

I'm going to remove entry points for this PR and they can be added back in a separate issue when appropriate functions are available as targets.

.github/workflows/CI.yml
aidanheerdegen: Done.

CodeGat (approved):

PR #77, opened 2024-08-19
ESM1.5 conversion driver, delete inputs upon successful conversion

### LINKED ISSUE #49: Driver script - fields file deletion upon successful conversion
In general during an ESM1.5 simulation, we'd like to delete the fields files after they've been converted to NetCDF. The original driver script for CM2 did this according to the value of a variable `REMOVE_FF` read in from the environment. 
```python
        else:
            um2netcdf4.process(input, output, args)
        if REMOVE_FF:
            input.unlink()
```

I think it would make sense for the ESM1.5 conversion driver to delete the fields files by default, perhaps with an optional flag in case anyone wants to keep them. There are a couple other situations when the fields files probably shouldn't be deleted:
 - If an error occurs during conversion.
 - If the conversion doesn't produce an error, but not all of the fields are successfully written to the NetCDF
    - This can happen for example when a field (say `var`) on pressure levels requires masking but the necessary `heaviside..` field isn't present. With the current behaviour of `um2nc`, `var` would be skipped and not written to the NetCDF file. I think it would be bad to delete the fields file when this happens, as its the only place  where `var` is kept.

This will require changes to the driving script, and probably changes to the warnings and exceptions in `um2netcdf`.

blimlim:
This pull request would close #49.

In most simulations of ESM1.5, the original fields files aren't very useful once they've been converted to NetCDF. This pull request adds a command line flag `--delete_ff`, `-d`, to the conversion driver. When included, the conversion driver will delete the input fields files when they are converted.

I've implemented this by adding the deletion as close as possible to the actual conversion:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/379001b7918487ee3e3a1eb05b5c523a13350b8f/umpost/conversion_driver_esm1p5.py#L156-L160

This does bury a bit more I/O in the `convert_fields_file_list`, but I think it seems safe to have it there to try and ensure that the only way a file can be deleted is if the conversion runs without any errors.

I also modified the `convert_fields_file_list` tests to check that file deletion is run when requested, and is not run when it is not requested or if there are any errors.

Let me know if these changes look ok, any feedback/suggestions are welcome!


umpost/conversion_driver_esm1p5.py
```diff
@@ -292,11 +299,15 @@ def convert_esm1p5_output_dir(esm1p5_output_dir):
                             "Otherwise report full stack trace."
                         )
                         )
+    parser.add_argument("--delete_ff", "-d", action="store_true",
```
truth-quark, line 302: This is a good setting as the default operation is safe (non-destructive). This places the onus on users to consider deletion.

The arg should be `--delete-ff` (hyphen) as `argparse` automatically converts the hyphen to an underscore in the args data structure --> `arg.delete_ff`.

blimlim: Ah thanks, I hadn't realised that! I've just fixed this up.

umpost/conversion_driver_esm1p5.py
```diff
@@ -154,6 +156,8 @@ def convert_fields_file_list(fields_file_paths, nc_write_dir):
         try:
             um2netcdf4.process(fields_file_path, nc_write_path, ARG_VALS)
             succeeded.append((fields_file_path, nc_write_path))
+            if delete_ff:
+                os.remove(fields_file_path)
```
truth-quark, line 160: Adding deletion code here adds an additional responsibility for the _conversion_ function. The deletion flag also touches a couple of the workflow function signatures.

I'd recommend removing the `delete_ff` handlers from the code &  using `__main__` for deletion like:
```Python
if args.delete_ff:
    # TODO: probably test that the conversion was OK
    # TODO: only delete successes?
    for path in successful_paths:
        os.remove(path)
```
I think it might be best as the final conversion driver step. This has the effect of running the conversion workflow first. In the event the conversions fail, deletion won't occur during that step & potentially leave ESM outputs in an incomplete state.

Unit testing deletion can also be skipped if there's only 2-3 lines of simple code (we visually confirm it's correct).

blimlim: Sweet, that sounds like a good idea. I've added this in, and added in an extra check to ensure that none of the successful conversion paths also appear in the failed conversions:

https://github.com/ACCESS-NRI/um2nc-standalone/blob/b86c0b2820610e0b6fb6ab04a708dfa79488b1b4/umpost/conversion_driver_esm1p5.py#L310-L322

Let me know if this is looking ok!

blimlim: Just realised I forgot to add the message to the PostProcessingError... Added it in.

blimlim: And just realised I forgot to modify the tests... on a roll today. Will fix them now.

test/test_conversion_driver_esm1p5.py
```diff
@@ -111,20 +116,43 @@ def _mock_process_with_exception(error_message):
...
+                                                 mock_os_remove,
+                                                 input_list):
+    """
+    Test that process is called for each input and that each input is deleted.
+    """
     input_list_paths = [Path(p) for p in input_list]
+
     succeeded, _ = esm1p5_convert.convert_fields_file_list(
-        input_list_paths, "fake_nc_write_dir")
+        input_list_paths, "fake_nc_write_dir", delete_ff=True)
```
truth-quark, line 128: `delete_ff=True` causes a test failure as `convert_fields_file_list()` has no flag arg.

test/test_conversion_driver_esm1p5.py
```diff
@@ -135,30 +163,37 @@ def test_convert_fields_file_list_fail_excepted(mock_process_with_exception):
     fake_file_path = Path("fake_file")
 
     _, failed = esm1p5_convert.convert_fields_file_list(
-        [fake_file_path], "fake_nc_write_dir")
+        [fake_file_path], "fake_nc_write_dir", delete_ff=True)
```
truth-quark, line 166: This also failed with the additional `delete_ff=True` flag arg.

test/test_conversion_driver_esm1p5.py
```diff
@@ -135,30 +163,37 @@ def test_convert_fields_file_list_fail_excepted(mock_process_with_exception):
...
-def test_convert_fields_file_list_fail_critical(mock_process_with_exception):
+def test_convert_fields_file_list_fail_critical(mock_process_with_exception, mock_os_remove):
     # Test that critical exceptions which are not allowed by ALLOWED_UM2NC_EXCEPTION_MESSAGES
     # are raised, and hence lead to the conversion crashing.
     generic_error_message = "Test error"
     mock_process_with_exception(generic_error_message)
     with pytest.raises(Exception) as exc_info:
         esm1p5_convert.convert_fields_file_list(
-            ["fake_file"], "fake_nc_write_dir")
+            ["fake_file"], "fake_nc_write_dir", delete_ff=True)
```
truth-quark: The test failed here and the one 12 lines down too.

test/test_conversion_driver_esm1p5.py
blimlim: Yep, sorry about this – forgot to fix the tests after making the change. Should be fixed now.

test/test_conversion_driver_esm1p5.py
blimlim: This one should be fixed now.

test/test_conversion_driver_esm1p5.py
blimlim: These ones should be fixed now too

test/test_conversion_driver_esm1p5.py
```diff
@@ -111,20 +116,43 @@ def _mock_process_with_exception(error_message):
...
+
     succeeded, _ = esm1p5_convert.convert_fields_file_list(
-        input_list_paths, "fake_nc_write_dir")
+        input_list_paths, "fake_nc_write_dir", delete_ff=True)
 
     assert mock_process.call_count == len(input_list)
+    assert mock_os_remove.call_count == len(input_list)
+
+    for path in input_list_paths:
+        assert mock.call(path) in mock_os_remove.call_args_list
```
truth-quark: Let's remove the mock assertion lines as they breaking the test with the assumption `convert_fields_file_list()` handles file deletion.

blimlim: Good point! the mocks of `os.remove` and the call assertions are removed with the latest changes.

umpost/conversion_driver_esm1p5.py
```diff
@@ -270,6 +271,14 @@ def convert_esm1p5_output_dir(esm1p5_output_dir):
 
     return succeeded, failed
 
+def success_fail_overlap(succeeded, failed):
+    succeeded_inputs = [succeed_path for succeed_path, _ in succeeded]
+    failed_inputs = [fail_path for fail_path, _ in failed]
+
+    overlap = set(succeeded_inputs) & set(failed_inputs)
+
+    return overlap
```
truth-quark, line 280: We can make use of set comprehensions here:
```Python
def success_fail_overlap(succeeded, failed):
    succeeded_inputs = {succeed_path for succeed_path, _ in succeeded}
    failed_inputs = {fail_path for fail_path, _ in failed}
    return succeeded_inputs - failed_inputs  # rely on set difference operation
```
... then add a test like:
```Python
def test_success_fail_overlap():
    # ensure the shared "fail" item is skipped
    a = ... a dummy successes data structure
    b = ... a dummy failures data structure
    res =  success_fail_overlap(a, b)   
    assert "fail" not in res
```
That should cover the overlap check.

The overlap check could also be called `safe_removal()`, so the calling code becomes:
```Python
for path in safe_removal():
    os.remove(path)
```
Although this loses the ability to report on files appearing in both lists.

truth-quark: > Although this loses the ability to report on files appearing in both lists.

This might not be a problem as there is error reporting on the failed files list already...

umpost/conversion_driver_esm1p5.py
truth-quark: That detail is _buried_ somewhere in the `argparse` docs.

umpost/conversion_driver_esm1p5.py
blimlim: Good idea, and it also makes sense to add the test. I've added this in in the latest commit!

blimlim: And then fixed the typos I added in...

truth-quark (approved):
Looks good to go!

Issue #82, opened 2024-08-21
Add `um2netcdf` entry point(s)


truth-quark:
For simplicity, this issue is for `um2netcdf` entry points, separate to the ESM1.5 conversion driver entry points.

Tasks for adding entry points for `um2netcdf` are:
* Move cmd line parsing & `process()` call into a new `main()` function
* Update `conda` (?) config to allow direct calls to `um2netcdf` (via `console-scripts`)
* Add CI test to verify simple command(s) can be called from the env e.g. `un2netcdf --help`

Other general thoughts:
* Moving the parsing & `process()` call to functions provides an interface for testing
  - As of mid Aug 2024, cmd line parsing cannot be tested as it's in `__main__`
* Adding `main()` provides a top level function
  - this facilitates splitting `process()` into workflow sub functions under `main()`
  - helps simplify testing with reduced need for complex test setups & mocking
  - new workflows can be composed from the components (do future ESM versions need custom `um2nc` workflows?)

truth-quark:
Resolved by #86.

truth-quark:
This relates to #81 too.

Issue #81, opened 2024-08-21
Need main functions or similar for entry points in conda


aidanheerdegen:
The conda deployment supports creating nice [entry points](https://packaging.python.org/en/latest/specifications/pyproject-toml/#entry-points), e.g. the ability to just call `um2nc` from the command line. For this to work e'll need main or similar functions defined, as entry points need to point at a function.

See this related PR where entry points were removed because there were no suitable functions:

https://github.com/ACCESS-NRI/um2nc-standalone/pull/74#issuecomment-2297854954

I started using the pattern shown below as a pretty standard approach in case it's useful

https://github.com/ACCESS-NRI/ncigrafana/blob/master/ncigrafana/parse_lquota.py#L146-L174

The corresponding entry point definition is

https://github.com/ACCESS-NRI/ncigrafana/blob/master/setup.cfg#L34

The point of this approach was to allow for the argument parsing to be tested independently of the call to `main`, and also allows for testing `main` directly in `pytest`, without the need for an entrypoint script (though I don't believe I actually had those tests in this case ... )

aidanheerdegen:
@blimlim if you want/need a post-processing script to be accessible as a command you should specify the requirements in this issue.

blimlim:
It would be great to have the post-processing script available as a command. Specifically we'd want an entry point to the `__main__` block in `conversion_driver_esm1p5.py`

https://github.com/ACCESS-NRI/um2nc-standalone/blob/8aa708d4dbeaa950301173475149a4f37c11ed4a/umpost/conversion_driver_esm1p5.py#L298-L328

Which hopefully should be pretty straightforward to adapt with the pattern in your example. Should I add a PR doing this?

aidanheerdegen:
Yep! Thanks.

PR #83, opened 2024-08-21
Split main into separate function and add entry point

blimlim:
This PR addresses the ESM1.5 conversion driver part of #81, splitting the `main` block into separate `main()` and `parse_args()` functions, which allows for an entry point to be added to the `pytproject.toml` file. 

Adding entry points to `um2netcdf` will be addressed separately in #82. 

I'm unsure how to test that the entry point is working, and so suggestions on that would be great too!

### DIFF
--- a/umpost/conversion_driver_esm1p5.py
+++ b/umpost/conversion_driver_esm1p5.py
@@ -12,6 +12,7 @@ https://github.com/ACCESS-NRI/access-cm2-drivers/blob/main/src/run_um2netcdf.py
 
 
 import os
+import sys
 import collections
 import re
 import f90nml
@@ -295,25 +296,36 @@ def safe_removal(succeeded, failed):
     return succeeded_inputs - failed_inputs
 
 
-if __name__ == "__main__":
+def parse_args(args):
+    """
+    Parse arguments given as list (args)
+    """
     parser = argparse.ArgumentParser()
     parser.add_argument(
-        "current_output_dir", help="ESM1.5 output directory to be converted", type=str
+        "current_output_dir", help="ESM1.5 output directory to be converted",
+        type=str
     )
     parser.add_argument("--quiet", "-q", action="store_true",
                         help=(
-                            "Report only final exception type and message for allowed"
-                            "exceptions raised during conversion when flag is included."
-                            "Otherwise report full stack trace."
+                            "Report only final exception type and message for "
+                            "allowed exceptions raised during conversion when "
+                            "flag is included. Otherwise report full "
+                            "stack trace."
                         )
                         )
     parser.add_argument("--delete-ff", "-d", action="store_true",
                         help="Delete fields files upon successful conversion."
                         )
-    args = parser.parse_args()
+
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args(sys.argv[1:])
 
     current_output_dir = args.current_output_dir
 
+    # Run conversion
     successes, failures = convert_esm1p5_output_dir(current_output_dir)
 
     # Report results to user
@@ -326,3 +338,8 @@ if __name__ == "__main__":
         # Remove files that appear only as successful conversions
         for path in safe_removal(successes, failures):
             os.remove(path)
+
+
+if __name__ == "__main__":
+
+    main()
--- a/umpost/conversion_driver_esm1p5.py
+++ b/umpost/conversion_driver_esm1p5.py
@@ -12,7 +12,6 @@ https://github.com/ACCESS-NRI/access-cm2-drivers/blob/main/src/run_um2netcdf.py
 
 
 import os
-import sys
 import collections
 import re
 import f90nml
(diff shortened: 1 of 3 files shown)


umpost/conversion_driver_esm1p5.py
```diff
@@ -295,25 +296,36 @@ def safe_removal(succeeded, failed):
     return succeeded_inputs - failed_inputs
 
 
-if __name__ == "__main__":
+def parse_args(args):
```
truth-quark, line 299: I think the `args` arg is superfluous here as `argparse` probably parses `sys.argv` by default.

umpost/conversion_driver_esm1p5.py
```diff
@@ -295,25 +296,36 @@ def safe_removal(succeeded, failed):
...
     parser.add_argument("--delete-ff", "-d", action="store_true",
                         help="Delete fields files upon successful conversion."
                         )
-    args = parser.parse_args()
+
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args(sys.argv[1:])
```
truth-quark, line 324: The `sys.argv[1:]` arg can be removed with the change to `parse_args()`.

umpost/conversion_driver_esm1p5.py
blimlim: Good point – it works without the extra argument. Have removed this.

umpost/conversion_driver_esm1p5.py
blimlim: Have removed this one too

umpost/conversion_driver_esm1p5.py
```diff
@@ -295,25 +296,36 @@ def safe_removal(succeeded, failed):
...
 
     current_output_dir = args.current_output_dir
 
+    # Run conversion
     successes, failures = convert_esm1p5_output_dir(current_output_dir)
```
truth-quark, line 328: For some minor styling observations:
* `convert_esm1p5_output_dir()` can take `args.current_output_dir` directly (shrinks code slightly)
* the comment duplicates "conversion" in the function name

truth-quark (changes requested):
I think the only thing missing is an entry point CI test to "run" the conversion driver script in a simple fashion.

Do you want to try copying & modifying the old entry point test from the CI.yml here:
https://github.com/ACCESS-NRI/um2nc-standalone/pull/74/commits/2c7b35542072907613767526090108c4f67183f7, and see if it can run the conversion driver script with `--help`?

aidanheerdegen:
@blimlim I debugged the `conda` packaging by copying the commands from the CI to build and install the package locally.

So something like
```
conda build -c accessnri -c conda-forge -c coecms . --no-anaconda-upload --output-folder=./build
conda env create -f conda/environment.yml 
conda activate um2nc 
conda install -c conda-forge -c accessnri -c coecms -c file://$(readlink -f build) um2nc
```

aidanheerdegen:
The problem is with package discovery

https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#custom-discovery

The package isn't discovered with the current settings because there isn't a `__init__.py` file in the `umpost` directory. So it isn't even installed correctly (see I said there would always be some problem!).

Just add one (it can be empty), e.g.
```
touch upmost/__init__.py
```
and it should all work as it is now.
```
(um2nc) $ esm1p5_convert_nc 
usage: esm1p5_convert_nc [-h] [--quiet] [--delete-ff] current_output_dir
esm1p5_convert_nc: error: the following arguments are required: current_output_dir
(um2nc) $ git status -s
?? build/
?? umpost/__init__.py
(um2nc) $ 
```

aidanheerdegen:
Doh. We need to also remove the `accessnri` channel when installing the `um2nc` package from the local directory, otherwise it will likely pick up the version that is already installed there rather the local one.

https://github.com/ACCESS-NRI/um2nc-standalone/pull/83/files#diff-3ab46ee209a127470fce3c2cf106b1a1dbadbb929a4b5b13656a4bc4ce19c0b8L86

*or* make the local file channel the [first channel to ensure it picks that version up](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html).

aidanheerdegen:
Yeah it picked up the accessnri channel version

https://github.com/ACCESS-NRI/um2nc-standalone/actions/runs/10484327176/job/29038506545?pr=83#step:6:213

blimlim:
I moved the local file to the first channel and it's working now!

umpost/conversion_driver_esm1p5.py
blimlim: Good idea, I've added these changes in

aidanheerdegen:
> I moved the local file to the first channel and it's working now!

Quick! Merge before anything changes! ;)

truth-quark (approved):
Looks good to go!

PR #86, opened 2024-08-22
Add `um2netcdf` CI entrypoint

### LINKED ISSUE #82: Add `um2netcdf` entry point(s)

truth-quark:
Closes #82.

This PR adds CI config to:
* Add an entry point for `um2netcdf`
* Run a simple check to call the command

Any comments welcome!

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -93,9 +93,13 @@ jobs:
         shell: bash -l {0}
         run: pylint --extension-pkg-whitelist=netCDF4 --ignored-modules=umpost -E umpost
 
-      - name: Entrypoint test
+      - name: Entrypoint test of driver script
         shell: bash -l {0}
         run: esm1p5_convert_nc --help
+
+      - name: Entrypoint test of um2netcdf script
+        shell: bash -l {0}
+        run: um2nc --help
 
       - name: Run tests
         shell: bash -l {0}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ dependencies = [
 
 [project.scripts]
 esm1p5_convert_nc = "umpost.conversion_driver_esm1p5:main"
+um2nc = "umpost.um2netcdf:main"
 
 [tool.versioneer]
 VCS = "git"


blimlim (approved):
Looks good!

.github/workflows/CI.yml
```diff
@@ -93,10 +93,14 @@ jobs:
         shell: bash -l {0}
         run: pylint --extension-pkg-whitelist=netCDF4 --ignored-modules=umpost -E umpost
 
-      - name: Entrypoint test
+      - name: Entrypoint test of driver script
         shell: bash -l {0}
         run: esm1p5_convert_nc --help
 
+      - name: Entrypoint test of um2netcdf script
+        shell: bash -l {0}
```
CodeGat, line 101: Should we invoke the shell with an `-e` flag so it errors on failure?

truth-quark: Is using `-e` the practice in other ACCESS repos?

CodeGat: I often just use the [default (bash) shell](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell) (no `shell: *` section at all unless required) - which is `bash --noprofile --norc -eo pipefail {0}`

truth-quark: Ok, to clarify, do you simply specify it as this?
```YAML
- name: Entrypoint test of um2netcdf script
  shell: bash
  run: um2nc --help
```

CodeGat: Even simpler, you can do:
```
- name: Entrypoint test of um2netcdf script
  run: um2nc --help
```
But now I wonder if they're specified for a reason...

truth-quark: It may have been from copying the CI config from another repo. @aidanheerdegen @blimlim do either of you recall if the test config was copied in, or was there was a reason for adding `bash -l {0}`?

aidanheerdegen: The login requirement for the shell was a recommendation from the miniforge action docs, otherwise the conda stuff wasn't set up correctly. I did wonder about making it default, and wasn't sure about the `-e`, so feel free to add that at the very least, or set up a default to do so. You might need to make a specific suggestion if you think we should make that change @CodeGat, or park it and make an issue to change it in the future.

aidanheerdegen: https://github.com/conda-incubator/setup-miniconda#use-a-default-shell

truth-quark: @aidanheerdegen @blimlim @CodeGat  To help with getting the recent entry point changes in for the release, are folks OK with moving the alternate `bash` suggestion to another issue & using these working settings for now?

CodeGat: Yep park it :fire:

CodeGat (approved):

truth-quark:
I've added the open question to #87.

PR #90, opened 2024-08-23
Add tests/clean `fix_level_coord()`

### LINKED ISSUE #53: Add tests for `fix_level_coord()`
Tasks for `fix_level_coord()`:

* [ ] Move function near other naming functionality
* [ ] Add unit tests
* [ ] Add constants for magic numbers
* [ ] Replace repeated strings with constants
* [ ] Can the logic be refactored?
* [ ] Refactor `process()` tests to remove level coord mocking

truth-quark:
Closes #53. Note, this PR is merges to `main`.

This PR:
* adds unit tests for `fix_level_coord()`
* moves the function down to call order (i.e. below `process()`)
* Cleans the code slightly

These tests use `iris` coordinate objects to avoid mocks cube attributes. Based on this experiment, this could be a reasonable approach for other tests requiring `iris` coordinates.

Any comments welcome!

### DIFF
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -11,6 +11,7 @@ import mule
 import mule.ff
 import iris.cube
 import iris.coords
+import iris.exceptions
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


test/test_um2netcdf.py
```diff
@@ -674,3 +675,46 @@ def test_fix_cell_methods_keep_weeks():
...
+def test_fix_level_coord_theta(z_sea_rho_data, z_sea_theta_data):
+    cube = FakeCubeCoords()
+    um2nc.fix_level_coord(cube, z_sea_rho_data, z_sea_theta_data)
+
+    assert LEVEL_COORDS[um2nc.MODEL_LEVEL_NUM].var_name == "model_theta_level_number"
+    assert LEVEL_COORDS[um2nc.LEVEL_HEIGHT].var_name == "theta_level_height"
+    assert LEVEL_COORDS[um2nc.SIGMA].var_name == "sigma_theta"
+
+
+def test_fix_level_coord_not_found(z_sea_rho_data, z_sea_theta_data):
```
blimlim, line 717: I'm wondering whether it would be worth having a quick docstring noting that this is checking `fix_level_coord` doesn't do anything if a cube doesn't have the vertical coordinates.

test/test_um2netcdf.py
```diff
@@ -674,3 +675,46 @@ def test_fix_cell_methods_keep_weeks():
     assert mod.method == cm.method
     assert mod.coord_names == cm.coord_names
     assert mod.intervals[0] == "week"
+
+
+LEVEL_HEIGHTS = [20.0003377]  # NB: technically only need [0]
```
blimlim, line 680: I'm wondering whether it's worth noting that this value comes from the `zsea_at_theta` data, so that it's clearer in `test_fix_level_coord_theta` why we are expecting the names to have the `theta` added. This might get a bit confusing though as the same data is used in the `test_fix_level_coord_rho` test, so am ok with either way here.

blimlim (changes requested):
Looks good! The only change is the `fix_level_coord` docstring, while the other two are questions about the clarity for the tests and so aren't completely necessary.

test/test_um2netcdf.py
truth-quark: Yep, that's a good idea as the test function name is too vague & there's no explanation. I'll aim for something where the name briefly states a requirement, then use a docstring for the detail.

test/test_um2netcdf.py
truth-quark: Yes, the provenance should be mentioned. Also, the `z_theta` fixture has `data = np.array([0.0, 20.000337706971997, 80.0...` where the `20.0...` is `data[1]` & not `data[0]` as per the comment. It's small, but indicates forgetting a detail within a few days.

umpost/um2netcdf.py
```diff
@@ -722,6 +699,43 @@ def fix_units(cube, um_var_units, verbose: bool):
...
 
+def fix_level_coord(cube, z_rho, z_theta, tol=1e-6):
+    """
+    Renames model_level_number coordinates to help distinguish rho/theta levels.
+
+    Cubes without 'model_level_number' coordinates are skipped.
+
+    Parameters
+    ----------
+    cube : iris.cube.Cube object to
```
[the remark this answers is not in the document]
truth-quark: It's fixed in https://github.com/ACCESS-NRI/um2nc-standalone/pull/90/commits/4392efd502fa91879eb3470a5e541984c685b782.

test/test_um2netcdf.py
truth-quark: I've made the `[0]` comment clearer. The test funcs need `points[0]`, so the incoming data only needs one point.

truth-quark:
I've refactored the tests with fixtures, now the assertions test against a `cube.coord("name")` function call interface. It has resulted in some 1st class object passing, but it's constrained within a fixture. There's a potential future benefit as the fixture `dataclass` can be called with init args in tests.

@blimlim  How do you find the readability of this change?

blimlim (approved):
Changes look good and clearer to read!

PR #93, opened 2024-08-29
Update ESM1.5 driver file naming.

### LINKED ISSUE #91: ESM1.5 driver output file names
As [discussed here](https://github.com/ACCESS-NRI/access-esm1.5-configs/issues/59#issuecomment-2311624726) the netCDF file names written using the ESM1.5 driver should be modified to include `YYYYMM` and the type of output (daily/monthly) to match the naming style in `/g/data/p73/archive/CMIP6/ACCESS-ESM1-5/`. The `NetCDF` output directory should also be renamed `netCDF`.

blimlim:
This pr attempts to close #91.

It updates the naming of the output netCDF files to include a `YYYYMM` date string, and a suffix stating the data frequency in a file. E.g. the fields file `aiihca.paa1jan` gets converted to `aiihca.pa-010101_mon.nc` following the convention in the `p73` archive. 

Getting this information involves two steps:
1. Reading the file date from the fields file header, via `mule.FixedLengthHeader.from_file`
2. Converting the unit key in the file name (the 9th character) to a frequency suffix, which should be one of `mon`,`dai`,`6hr`, or `3hr`. 

Originally setting the output filepath was handled by `get_nc_write_path()`, which was called from inside `convert_fields_file_list()` and just glued `.nc` onto the name:

https://github.com/ACCESS-NRI/um2nc-standalone/blob/70a890cc9434a501810d549e308668f59d21eb00/umpost/conversion_driver_esm1p5.py#L141-L143

I thought adding changes 1 and 2 to the naming here would involve burying more I/O which might make `convert_fields_file_list` harder to test, and also would make it the function a bit inflexible (e.g. the same function couldn't be used with a different naming convention).

I tried moving the file naming out from the conversion function up to the higher level `convert_esm1p5_output_dir` function. It uses an updated `get_nc_write_path()` to create a list of `(input_path, output_path)` pairs, which gets fed into `convert_fields_file_list` 

https://github.com/ACCESS-NRI/um2nc-standalone/blob/2c5fceab2b7a59780d2398e15aa0c0cdd77a04e1/umpost/conversion_driver_esm1p5.py#L343-L347

`get_nc_write_path` then performs steps 1 and 2, and I tried to make it so that it could be a target for more "mid level" testing. 

I went back and forth a bit on the best way to structure the changes and am still not completely sure, so any suggestions would be welcome!

### DIFF
--- a/test/test_conversion_driver_esm1p5.py
+++ b/test/test_conversion_driver_esm1p5.py
@@ -16,6 +16,7 @@ def test_get_esm1p5_fields_file_pattern_wrong_id_length(run_id):
     with pytest.raises(ValueError):
         esm1p5_convert.get_esm1p5_fields_file_pattern(run_id)
 
+
 @pytest.mark.parametrize("ff_path,ff_year,ff_month,nc_write_dir,expected",
                          [
                             (
@@ -53,12 +54,11 @@ def test_get_nc_write_path_recognized_unit(ff_path, ff_year, ff_month,
     Check that netCDF file naming produces expected file paths for various
     expected unit keys.
     """
-    # Mock get_ff_date to avoid reading header from fields file
-    with mock.patch("umpost.conversion_driver_esm1p5.get_ff_date",
-                    return_value=(ff_year, ff_month, 1)):
-
-        nc_write_path = esm1p5_convert.get_nc_write_path(ff_path,
-                                                         nc_write_dir)
+    nc_write_path = esm1p5_convert.get_nc_write_path(
+                        ff_path,
+                        nc_write_dir,
+                        (ff_year, ff_month, 1)
+                    )
 
     assert nc_write_path == expected
 
@@ -77,12 +77,11 @@ def test_get_nc_write_path_unrecognized_unit():
     nc_write_dir = Path("netCDF")
     expected_nc_write_path = nc_write_dir / f"aiihca.p{unknown_key}-005007.nc"
 
-    # Mock get_ff_date to avoid reading header from fields file
-    with mock.patch("umpost.conversion_driver_esm1p5.get_ff_date",
-                    return_value=(ff_year, ff_month, 1)):
-        with pytest.warns():
-            nc_write_path = esm1p5_convert.get_nc_write_path(Path(ff_name),
-                                                             nc_write_dir)
+    nc_write_path = esm1p5_convert.get_nc_write_path(
+                        Path(ff_name),
+                        nc_write_dir,
+                        (ff_year, ff_month, 1)
+                    )
 
     assert nc_write_path == expected_nc_write_path
 
--- a/test/test_conversion_driver_esm1p5.py
+++ b/test/test_conversion_driver_esm1p5.py
@@ -44,6 +44,12 @@ def test_get_esm1p5_fields_file_pattern_wrong_id_length(run_id):
                                 Path("./netCDF"),
                                 Path("./netCDF/aiihca.pj-012001_6hr.nc")
                             ),
+                            (
+                                Path("abc/aiihca.pjc0jan"),
+                                None,
+                                Path("./netCDF"),
+                                Path("./netCDF/aiihca.pjc0jan.nc")
+                            ),
                          ])
 def test_get_nc_write_path_recognized_unit(ff_path, ff_date,
                                            nc_write_dir, expected):
--- a/test/test_conversion_driver_esm1p5.py
+++ b/test/test_conversion_driver_esm1p5.py
@@ -16,15 +16,75 @@ def test_get_esm1p5_fields_file_pattern_wrong_id_length(run_id):
     with pytest.raises(ValueError):
         esm1p5_convert.get_esm1p5_fields_file_pattern(run_id)
 
+@pytest.mark.parametrize("ff_path,ff_year,ff_month,nc_write_dir,expected",
+                         [
+                            (
+                                Path("/test/aiihca.paa1feb"),
+                                101,
+                                2,
+                                Path("/test/netCDF"),
+                                Path("/test/netCDF/aiihca.pa-010102_mon.nc")
+                            ),
+                            (
+                                Path("aiihca.pe50dec"),
+                                1850,
+                                12,
+                                Path("netCDF"),
+                                Path("netCDF/aiihca.pe-185012_dai.nc")
+                            ),
+                            (
+                                Path("abc/aiihca.pi87jun"),
+                                1887,
+                                6,
+                                Path("./netCDF"),
+                                Path("./netCDF/aiihca.pi-188706_3hr.nc")
+                            ),
+                            (
+                                Path("abc/aiihca.pjc0jan"),
+                                120,
+                                1,
+                                Path("./netCDF"),
+                                Path("./netCDF/aiihca.pj-012001_6hr.nc")
+                            )
+                         ])
+def test_get_nc_write_path_recognized_unit(ff_path, ff_year, ff_month,
+                                           nc_write_dir, expected):
+    """
+    Check that netCDF file naming produces expected file paths for various
+    expected unit keys.
+    """
+    # Mock get_ff_date to avoid reading header from fields file
+    with mock.patch("umpost.conversion_driver_esm1p5.get_ff_date",
+                    return_value=(ff_year, ff_month, 1)):
 
-# def test_get_nc_write_path():
-#     fields_file_path = Path("/test/path/fields_123.file")
-#     nc_write_dir = Path("/test/path/NetCDF")
+        nc_write_path = esm1p5_convert.get_nc_write_path(ff_path,
+                                                         nc_write_dir)
 
-#     nc_write_path = esm1p5_convert.get_nc_write_path(
-#         fields_file_path, nc_write_dir)
+    assert nc_write_path == expected
 
-#     assert nc_write_path == Path("/test/path/NetCDF/fields_123.file.nc")
+
+def test_get_nc_write_path_unrecognized_unit():
+    """
+    Check that netCDF file naming falls back to simpler naming scheme
+    when unit key in fields file name not recognized.
+    """
+    unknown_key = "w"
+    assert unknown_key not in esm1p5_convert.FF_UNIT_SUFFIX.keys()
+
+    ff_name = f"aiihca.p{unknown_key}abcd"
+    ff_year = 50
+    ff_month = 7
+    nc_write_dir = Path("netCDF")
+    expected_nc_write_path = nc_write_dir / f"{ff_name}-005007.nc"
+
+    # Mock get_ff_date to avoid reading header from fields file
+    with mock.patch("umpost.conversion_driver_esm1p5.get_ff_date",
+                    return_value=(ff_year, ff_month, 1)):
+        with pytest.warns():
+            nc_write_path = esm1p5_convert.get_nc_write_path(Path(ff_name),
+                                                             nc_write_dir)
+
+    assert nc_write_path == expected_nc_write_path
 
 
 def test_find_matching_fields_files():
--- a/test/test_conversion_driver_esm1p5.py
+++ b/test/test_conversion_driver_esm1p5.py
@@ -43,12 +43,6 @@ def test_get_esm1p5_fields_file_pattern_wrong_id_length(run_id):
                                 Path("./netCDF"),
                                 Path("./netCDF/aiihca.pj-012001_6hr.nc")
                             ),
-                            (
-                                Path("abc/aiihca.paa1jan"),
-                                None,
-                                Path("./netCDF"),
-                                Path("./netCDF/aiihca.paa1jan.nc")
-                            ),
                          ])
 def test_get_nc_write_path_recognized_unit(ff_path, ff_date,
                                            nc_write_dir, expected):
--- a/test/test_conversion_driver_esm1p5.py
+++ b/test/test_conversion_driver_esm1p5.py
@@ -3,6 +3,7 @@ import umpost.conversion_driver_esm1p5 as esm1p5_convert
 import pytest
 from pathlib import Path
 import unittest.mock as mock
+import umpost.um2netcdf as um2nc
 
 
 def test_get_esm1p5_fields_file_pattern():
@@ -153,11 +154,11 @@ def mock_process(base_mock_process):
 
 @pytest.fixture
 def mock_process_with_exception(mock_process):
-    # Add a generic exception with chosen message to mock_process.
-    # Yield function so that tests of different exception messages
+    # Add a specified exception with chosen message to mock_process.
+    # Yield function so that tests of different exceptions and messages
     # can make use of the same fixture.
-    def _mock_process_with_exception(error_message):
-        mock_process.side_effect = Exception(error_message)
+    def _mock_process_with_exception(error_type, error_message):
+        mock_process.side_effect = error_type(error_message)
 
     yield _mock_process_with_exception
 
@@ -184,13 +185,8 @@ def test_convert_fields_file_list_success(mock_process,
 
 
 def test_convert_fields_file_list_fail_excepted(mock_process_with_exception):
-    # Hopefully this test will be unnecessary with um2nc standalone.
-    # Test that the "Variable can not be processed" error arising from time
-    # series inputs is excepted.
-    allowed_error_message = esm1p5_convert.ALLOWED_UM2NC_EXCEPTION_MESSAGES[
-        "TIMESERIES_ERROR"
-    ]
-    mock_process_with_exception(allowed_error_message)
+    mock_process_with_exception(um2nc.UnsupportedTimeSeriesError,
+                                "timeseries error")
     fake_input_output_paths = [(Path("fake_file"), Path("fake_file.nc"))]
 
     _, failed = esm1p5_convert.convert_fields_file_list(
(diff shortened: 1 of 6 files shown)


truth-quark:
There's a few things here:
* The side by side diff looks more complicated as `find_matching_fields_files()` moved position. I reordered locally & the diff was easier to parse.
* an existing function signature has changed
* the tests need to mock `get_ff_date()`, which signifies an unwanted dependency on I/O or external thing

A few changes can be made to simplify the code & touch less of the existing interface. The `mule`/paths/dates I/O can be moved higher, so the paths can be sorted out at a high level, only passing data to lower level funcs. The fixes change this loop:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/2c5fceab2b7a59780d2398e15aa0c0cdd77a04e1/umpost/conversion_driver_esm1p5.py#L343-L346

I'd recommend the following changes:
* Loop over `atm_dir_fields_files` for paths for creating `input_output_paths`
* Move the `get_ff_date()` call into the loop to get dates for each file
* Tweak `get_nc_write_path(fields_file_path, nc_write_dir)` to `get_nc_write_path(fields_file_path, nc_write_dir, date=None)`
  - If the `date` input is `None`, default to the prior behaviour of adding `.nc`
  - Otherwise return the new date structured filename
* Call `get_nc_write_path()` in this top loop to create `input_output_paths`
* See if removing the test mocks works

Do you want to see if those changes help simplify the naming feature & tests?

truth-quark:
Ah, there's a video about software architecture called "Boundaries", also known as "Functional Core, Imperative Shell" (FCIS). This is the sort of approach guiding `um2nc` dev. I think the latest talk is this one:
https://www.youtube.com/watch?v=yTkzNHF6rMs

blimlim:
Sweet, I've had a go at putting in these changes.

> The side by side diff looks more complicated as find_matching_fields_files() moved position. I reordered locally & the diff was easier to parse.

Oops sorry about that! This should be fixed in aff66bd and hopefully it's a bit easier to read the changes.

I've moved the `get_ff_date` function up into the loop, which then lets us remove that extra mock in the tests.

I have a couple of extra questions/comments, which I've added over the code.

umpost/conversion_driver_esm1p5.py
```diff
@@ -117,14 +130,73 @@ def find_matching_fields_files(dir_contents, fields_file_name_pattern):
...
+
+    Returns
+    -------
+    name: formated netCDF filename for writing output.
+    """
+    if date is None:
+        # If no date information supplied, just append ".nc"
+        # to current file name.
+        return f"{fields_file_name}.nc"
+
```
blimlim, line 152: > Tweak get_nc_write_path(fields_file_path, nc_write_dir) to get_nc_write_path(fields_file_path, nc_write_dir, date=None)
If the date input is None, default to the prior behaviour of adding .nc

I've added the naming logic into the `get_nc_filename` setting, so that the new naming rules are applied when a date is supplied, while it reverts to the old behaviour if a date is not supplied.

@truth-quark do you think it's worth supporting both of these cases? From I've seen, output fields files from ESM1.5 will always contain a date.

umpost/conversion_driver_esm1p5.py
```diff
@@ -259,15 +326,18 @@ def convert_esm1p5_output_dir(esm1p5_output_dir):
...
+            "converted to netCDF."
         )
 
         return [], []  # Don't try to run the conversion
 
-    succeeded, failed = convert_fields_file_list(
-        atm_dir_fields_files,
-        nc_write_dir
-    )
+    input_output_pairs = []
```
blimlim, line 334: @truth-quark I've added a loop here where `get_ff_date` is called, and the list of input/output paths is constructed. Let me know if this structure seems reasonable!

umpost/conversion_driver_esm1p5.py
```diff
@@ -70,27 +82,28 @@ def get_esm1p5_fields_file_pattern(run_id: str):
...
-    nc_write_dir : path to target directory for writing NetCDF files.
+    nc_write_dir : path to target directory for writing netCDF files.
+    date : tuple of form (year, month, day) associated with fields file data.
 
     Returns
     -------
     nc_write_path : path for writing converted fields_file_path file.
     """
     fields_file_path = Path(fields_file_path)
+    fields_file_name = fields_file_path.name
```
truth-quark, line 100: A minor simplification is using `fields_file_path.name` directly in the `get_nc_filename()` call:
```suggestion
    nc_write_dir = Path(nc_write_dir)
    nc_file_name = get_nc_filename(fields_file_path.name, date)
```

truth-quark (commented):
I've listed a few first round fixes (somehow they got bundled into a review instead of individual comments).

The biggest current problem is the failing CI, once that passes we can go over other steps.

umpost/conversion_driver_esm1p5.py
```diff
@@ -117,14 +130,73 @@ def find_matching_fields_files(dir_contents, fields_file_name_pattern):
...
+    fields_file_name: name of fields file to be converted.
+    date: tuple of form (year, month, day) associated with fields file data.
+
+    Returns
+    -------
+    name: formated netCDF filename for writing output.
+    """
+    if date is None:
+        # If no date information supplied, just append ".nc"
+        # to current file name.
```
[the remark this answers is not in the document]
truth-quark: The comment could be moved into the user docs & reworded to indicate `None` is a valid input.

umpost/conversion_driver_esm1p5.py
```diff
@@ -117,14 +130,73 @@ def find_matching_fields_files(dir_contents, fields_file_name_pattern):
...
+    Returns
+    -------
+    date_tuple : tuple of integers (yyyy,mm,dd) containing the fields
+                 file's date.
+    """
+    fields_file_header = mule.FixedLengthHeader.from_file(
+                                            str(fields_file_path))
+    year = fields_file_header.t2_year
+    month = fields_file_header.t2_month
+    day = fields_file_header.t2_day
```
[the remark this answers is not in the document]
blimlim: I've tested this out in the temporary PR here (I'll delete it if we're happy to bring the changes back into this PR): https://github.com/ACCESS-NRI/um2nc-standalone/pull/103/

The `# noqa` comments didn't prevent the `pylint` errors, though a similar pylint syntax: `# pylint: disable=no-member` ended up working.

umpost/conversion_driver_esm1p5.py
```diff
@@ -259,15 +326,18 @@ def convert_esm1p5_output_dir(esm1p5_output_dir):
...
 
-    succeeded, failed = convert_fields_file_list(
-        atm_dir_fields_files,
-        nc_write_dir
-    )
+    input_output_pairs = []
+    for ff_path in atm_dir_fields_files:
+        ff_date = get_ff_date(ff_path)
+        nc_path = get_nc_write_path(ff_path, nc_write_dir, ff_date)
+        input_output_pairs.append((ff_path, nc_path))
```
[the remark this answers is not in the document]
blimlim: Can do! Added in a list comprehension in c040780

umpost/conversion_driver_esm1p5.py
blimlim: Good idea! Added in c040780

umpost/conversion_driver_esm1p5.py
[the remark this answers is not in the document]
blimlim: I've incorporated the `# pylint: disable=no-member` comments into this branch and the CI checks are now passing.

umpost/conversion_driver_esm1p5.py
blimlim: After trialing removing the `date = None` option in [this branch](https://github.com/ACCESS-NRI/um2nc-standalone/blob/afdb0e22e8f6c251b0e6a23bfaa3adda48b76971/umpost/conversion_driver_esm1p5.py#L126-L161) I'm wondering whether it's simplest to actually keep the `date=None` option in?

> I'd wrap `year, month, _ = date` in a `try/except AttributeError` to catch anything passed in without a date interface.

The main part I'm unsure about is the best way for the exceptions to be handled. E.g. if `date=None` was supplied at this point, I think we'd get a `TypeError`, but in the current code I'm not sure if there's a clear way for this to happen. Let me know if the experiment in [this branch](https://github.com/ACCESS-NRI/um2nc-standalone/blob/afdb0e22e8f6c251b0e6a23bfaa3adda48b76971/umpost/conversion_driver_esm1p5.py#L126-L161) looks reasonable/similar to what you had in mind though!

umpost/conversion_driver_esm1p5.py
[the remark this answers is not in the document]
blimlim: Have moved this into the docstring in c040780, and will remove it if we decide to remove `None` as a valid input!

umpost/conversion_driver_esm1p5.py
truth-quark: > After trialing removing the `date = None` option in [this branch](https://github.com/ACCESS-NRI/um2nc-standalone/blob/afdb0e22e8f6c251b0e6a23bfaa3adda48b76971/umpost/conversion_driver_esm1p5.py#L126-L161) I'm wondering whether it's simplest to actually keep the `date=None` option in?

Leaving `date=None` in is fine as the func provides a sensible default result. Let's leave that part in.

truth-quark: > > I'd wrap `year, month, _ = date` in a `try/except AttributeError` to catch anything passed in without a date interface.
> 
> The main part I'm unsure about is the best way for the exceptions to be handled. E.g. if `date=None` was supplied at this point, I think we'd get a `TypeError`, but in the current code I'm not sure if there's a clear way for this to happen. Let me know if the experiment in [this branch](https://github.com/ACCESS-NRI/um2nc-standalone/blob/afdb0e22e8f6c251b0e6a23bfaa3adda48b76971/umpost/conversion_driver_esm1p5.py#L126-L161) looks reasonable/similar to what you had in mind though!

Ah yes, `TypeError` occurs when attempting to unpack a non-iter `None`. Trying `None.foo()` in a REPL raises an `AttributeError` as `None` is an instance of `NoneType`. Both cases can be captured in a multi error type handler:

```Python
try:
    # unpack None or possible non date obj access
except (TypeError, AttributeError)  as err:
    # some kind of handling
```

I'd also include the date value in the error message for debugging assistance.

Also, you could modify the parameter docs to indicate `None` & `date` are the expected args. This leaves users to debug passing in the wrong thing :-)

blimlim: > Also, you could modify the parameter docs to indicate None & date are the expected args

I've updated the docstrings to be a bit more clear about `None` being allowed in e9d0d79. 

I just wanted to check whether I've understood things correctly – since we're supporting the `date=None` case in the following, can we avoid adding the extra error handling?

https://github.com/ACCESS-NRI/um2nc-standalone/blob/e9d0d79627f0a6097991cdf687329d2311a30663/umpost/conversion_driver_esm1p5.py#L142-L143

umpost/conversion_driver_esm1p5.py
```diff
@@ -110,14 +122,73 @@ def find_matching_fields_files(dir_contents, fields_file_name_pattern):
...
+    date_tuple : tuple of integers (yyyy,mm,dd) containing the fields
+                 file's date.
+    """
+    fields_file_header = mule.FixedLengthHeader.from_file(
+                                            str(fields_file_path))
+    year = fields_file_header.t2_year       # pylint: disable=no-member
+    month = fields_file_header.t2_month     # pylint: disable=no-member
+    day = fields_file_header.t2_day         # pylint: disable=no-member
+
+    return (year, month, day)
```
truth-quark, line 181: Minor styling: the brackets can be removed.

umpost/conversion_driver_esm1p5.py
```diff
@@ -110,14 +122,73 @@ def find_matching_fields_files(dir_contents, fields_file_name_pattern):
...
+    """
+    Format a netCDF output filename based on the input fields file name and
+    its date. Assumes fields_file_name follows ESM1.5's naming convention
+    '{5 char run_id}.pa{unit}{date encoding}`.
+
+    Parameters
+    ----------
+    fields_file_name: name of fields file to be converted.
+    date: tuple of form (year, month, day) associated with fields file data,
+          or None. If None, ".nc" will concatenated to the original fields
```
truth-quark, line 135: There's a small typo with "concatenated".

umpost/conversion_driver_esm1p5.py
truth-quark: > I just wanted to check whether I've understood things correctly – since we're supporting the `date=None` case in the following, can we avoid adding the extra error handling?

Oooops, a mental shortcut meant wrongly assuming a `datetime.date` obj was being passed in & thinking about a wrong solution <facepalm>.

Since the date passing is handled internally, let's assume the those functions are internal & not used as an API. Thus, there's no requirement to do the extra Type/AttributeError checks.

umpost/conversion_driver_esm1p5.py
truth-quark, line 181: One idea, we can simplify the code to deduplicate the `noqa` guards:
```suggestion
    header = mule.FixedLengthHeader.from_file(str(fields_file_path))
    return header.t2_year, header.t2_month, header.t2_day  # pylint: disable=no-member
```
This provides the benefit of constraining the guard to 1 line of code.

truth-quark: This can be skipped if the code suggestion for this block is used.

umpost/conversion_driver_esm1p5.py
```diff
@@ -249,15 +315,20 @@ def convert_esm1p5_output_dir(esm1p5_output_dir):
...
-        atm_dir_fields_files,
-        nc_write_dir
-    )
+    input_output_pairs = [
+        (
+            ff_path,
+            get_nc_write_path(ff_path, nc_write_dir, get_ff_date(ff_path))
+        )
+        for ff_path in atm_dir_fields_files
+    ]
```
truth-quark, line 329: We can rework this multi-line comprehension into 2 steps:
```suggestion
    outputs = [get_nc_write_path(path, nc_write_dir, get_ff_date(path)) for path in atm_dir_fields_files]
    input_output_pairs = zip(atm_dir_fields_files, outputs)
```
I find this clearer as the output path & in/out tuple concatenation steps are separate. Also, less brackets are required, which were getting forced onto separate lines by the formatter.

truth-quark (commented):
Adding a review in the event comments are _blocked_ without one.

umpost/conversion_driver_esm1p5.py
blimlim: Good catch! Fixed up in 797a824

umpost/conversion_driver_esm1p5.py
blimlim: Added this in 797a824

umpost/conversion_driver_esm1p5.py
blimlim: Added in 797a824

Just to understand better, is the main reason to use the list comprehension over the manual `append` that it's faster? Probably because I'm still newish to Python I still do find the manual loop + `append` the easiest to read.

truth-quark: Comprehensions should be faster, although they're the recommended language idiom for creating data sequences. The gist is they're more declarative than the imperative style of manually assembling new data structures in a stepwise fashion.
This talk should help explain core idioms & describe data interfaces for composing code, _Transforming Code Into Beautiful, Idiomatic Python | Raymond Hettinger_:
https://www.youtube.com/watch?v=anrOzOapJ2E
List comprehensions are around the 46 minute mark, but early on he discusses moving away from manual looping.
(The speaker is a core Python developer)

blimlim: Awesome, thanks!

truth-quark (approved):
This should be good to go!

blimlim:
@marc-white, I'm just pinging you for a second pair of eyes to look over the changes. Any suggestions or feedback you have are welcome!

umpost/conversion_driver_esm1p5.py
```diff
@@ -110,14 +122,70 @@ def find_matching_fields_files(dir_contents, fields_file_name_pattern):
...
+
+    Returns
+    -------
+    name: formated netCDF filename for writing output.
+    """
+    if date is None:
+        return f"{fields_file_name}.nc"
+
+    stem = fields_file_name[0:FF_UNIT_INDEX + 1]
+    unit = fields_file_name[FF_UNIT_INDEX]
```
marc-white, line 146: This suffices for this use case, but you may want to consider adding a `TODO` to switch over to regular expressions. They're more flexible (especially if this conversion ever needs to be done for other models/filename patterns), and I think they're also a bit clearer as to what you're doing (i.e., it would be more explicit what you're pulling out of the filename pattern and from where).

marc-white: Also forgot to mention that REs would give immediate feedback on whether or not `fields_file_name` conforms to the expected pattern.

test/test_conversion_driver_esm1p5.py
```diff
@@ -18,14 +18,75 @@ def test_get_esm1p5_fields_file_pattern_wrong_id_length(run_id):
...
 
     nc_write_path = esm1p5_convert.get_nc_write_path(
-        fields_file_path, nc_write_dir)
+                        Path(ff_name),
+                        nc_write_dir,
+                        (ff_year, ff_month, 1)
+                    )
 
-    assert nc_write_path == Path("/test/path/NetCDF/fields_123.file.nc")
+    assert nc_write_path == expected_nc_write_path
```
marc-white, line 90: Is there a need to test against what happens if a totally invalid filename gets fired into `get_nc_write_path`? Is that behaviour even defined/consistent?

marc-white (approved):
I think this is ready to go as-is; comments added are things to consider/possible `TODO` items to add.

umpost/conversion_driver_esm1p5.py
blimlim: That's a good idea. I've added a `# TODO` comment and will make a new issue for this.

test/test_conversion_driver_esm1p5.py
blimlim: Good question. Currently the `get_nc_write_path` is only being provided file names that match some sort of expected format provided by the `find_matching_fields_files` function, however it probably makes sense to more explicitly deal with invalid file names being supplied. I think if we swap to using regex following your suggestion, this should be easier!

PR #94, opened 2024-08-30
92/Refactor pressure level modifications

### LINKED ISSUE #92: Refactor: `cubewrite()` pressure level correction to tested function
Tasks:
* [x] Extract this block https://github.com/ACCESS-NRI/um2nc-standalone/blob/main/umpost/um2netcdf.py#L150-L160
* [x] Add unit tests (~2)
* [ ] Move into `process()`

truth-quark:
Resolves #92.

This PR covers a few tasks:
* extracts pressure level modification functionality to a standalone function
* adds partial unit testing
* rearranges the `um2netcdf` module slightly
* partial code cleanup

**Known gaps:**

This work exposes a (possibly one off) case where `iris` funcs return a _new_ cube. Moving the pressure level fixes into `process()` requires modifying the list of cubes being processed to patch the new cube in. This is noted in the code. The tests are also moderately loose, related to the complexity with testing `iris` components.

It's likely addressing this will have to be deferred until the code is restructured & tested, to allow workflow refactoring.

Otherwise, any comments welcome!

### DIFF
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -759,7 +759,7 @@ def test_fix_plevs_no_pressure_coord(get_fake_cube_coords):
     with pytest.raises(iris.exceptions.CoordinateNotFoundError):
         cube.coord("pressure")  # ensure missing 'pressure' coord
 
-    um2nc.fix_plevs(cube)  # should just exit
+    um2nc.fix_pressure_levels(cube)  # should just exit
 
 
 def _add_attrs_points(m_plevs: mock.MagicMock, points):
@@ -775,7 +775,7 @@ def test_fix_plevs_do_rounding(get_fake_cube_coords):
     extra = {"pressure": m_plevs}
     cube = get_fake_cube_coords(extra)
 
-    um2nc.fix_plevs(cube)
+    um2nc.fix_pressure_levels(cube)
 
     # TODO: test flaw, this verifies pressure coord but ignores fix_plevs()
     #       returning a new cube if the pressure is reversed. This is verified
@@ -792,7 +792,7 @@ def test_fix_plevs_reverse_pressure(get_fake_cube_coords):
     cube = get_fake_cube_coords(extra)
 
     with mock.patch("iris.util.reverse"):
-        um2nc.fix_plevs(cube)
+        um2nc.fix_pressure_levels(cube)
 
     # TODO: test flaw, this verifies pressure coord but ignores fix_plevs()
     #       returning a new cube if the pressure is reversed. This is verified
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -745,3 +745,44 @@ def test_fix_level_coord_skipped_if_no_levels(z_sea_rho_data, z_sea_theta_data):
     m_cube = mock.Mock(iris.cube.Cube)
     m_cube.coord.side_effect = iris.exceptions.CoordinateNotFoundError
     um2nc.fix_level_coord(m_cube, z_sea_rho_data, z_sea_theta_data)
+
+
+# fix pressure level tests
+
+def test_fix_plevs_no_pressure_coord(get_fake_cube_coords):
+    cube = get_fake_cube_coords()
+
+    with pytest.raises(KeyError):
+        um2nc.fix_plevs(cube)  # ensure there's no 'pressure' key
+
+
+def _add_attrs_points(m_plevs: mock.MagicMock, points):
+    setattr(m_plevs, "attributes", {"positive": None})
+    setattr(m_plevs, "points", points)
+
+
+def test_fix_plevs_do_rounding(get_fake_cube_coords):
+    m_plevs = mock.Mock()
+    _add_attrs_points(m_plevs, [1.000001, 0.000001])
+    extra = {"pressure": m_plevs}
+    cube = get_fake_cube_coords(extra)
+
+    um2nc.fix_plevs(cube)
+
+    plev = cube.coord('pressure')
+    assert plev.attributes["positive"] == "down"
+    assert all(plev.points == [1.0, 0.0])
+
+
+def test_fix_plevs_reverse_pressure(get_fake_cube_coords):
+    m_plevs = mock.Mock()
+    _add_attrs_points(m_plevs, [0.000001, 1.000001])
+    extra = {"pressure": m_plevs}
+    cube = get_fake_cube_coords(extra)
+
+    with mock.patch("iris.util.reverse"):
+        um2nc.fix_plevs(cube)
+
+    plev = cube.coord('pressure')
+    assert plev.attributes["positive"] == "down"
+    assert all(plev.points == [0.0, 1.0])
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -777,6 +777,9 @@ def test_fix_plevs_do_rounding(get_fake_cube_coords):
 
     um2nc.fix_plevs(cube)
 
+    # TODO: test flaw, this verifies pressure coord but ignores fix_plevs()
+    #       returning a new cube if the pressure is reversed. This is verified
+    #       in command line testing though
     plev = cube.coord('pressure')
     assert plev.attributes["positive"] == "down"
     assert all(plev.points == [1.0, 0.0])
@@ -791,6 +794,9 @@ def test_fix_plevs_reverse_pressure(get_fake_cube_coords):
     with mock.patch("iris.util.reverse"):
         um2nc.fix_plevs(cube)
 
+    # TODO: test flaw, this verifies pressure coord but ignores fix_plevs()
+    #       returning a new cube if the pressure is reversed. This is verified
+    #       in command line testing though
     plev = cube.coord('pressure')
     assert plev.attributes["positive"] == "down"
     assert all(plev.points == [0.0, 1.0])
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -753,7 +753,7 @@ def test_fix_level_coord_skipped_if_no_levels(z_sea_rho_data, z_sea_theta_data):
 
 # tests - fix pressure level data
 
-def test_fix_plevs_no_pressure_coord(get_fake_cube_coords):
+def test_fix_pressure_levels_no_pressure_coord(get_fake_cube_coords):
     cube = get_fake_cube_coords()
 
     with pytest.raises(iris.exceptions.CoordinateNotFoundError):
@@ -769,10 +769,10 @@ def _add_attrs_points(m_plevs: mock.MagicMock, points):
     setattr(m_plevs, "points", points)
 
 
-def test_fix_plevs_do_rounding(get_fake_cube_coords):
-    m_plevs = mock.Mock()
-    _add_attrs_points(m_plevs, [1.000001, 0.000001])
-    extra = {"pressure": m_plevs}
+def test_fix_pressure_levels_do_rounding(get_fake_cube_coords):
+    m_pressure = mock.Mock()
+    _add_attrs_points(m_pressure, [1.000001, 0.000001])
+    extra = {"pressure": m_pressure}
     cube = get_fake_cube_coords(extra)
 
     um2nc.fix_pressure_levels(cube)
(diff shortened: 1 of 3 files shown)


CodeGat (approved):
Looks logical to me.

test/test_um2netcdf.py
```diff
@@ -683,21 +683,31 @@ def level_heights():
...
     return [20.0003377]
 
+
 @pytest.fixture
 def level_coords(level_heights):
     return {um2nc.MODEL_LEVEL_NUM: iris.coords.DimCoord(range(1, 39)),
             um2nc.LEVEL_HEIGHT: iris.coords.DimCoord(level_heights),
             um2nc.SIGMA: iris.coords.AuxCoord(np.array([0.99771646]))}
 
+
```
truth-quark, line 693: The fixture below should be replaceable with the `DummyCubeWithCoords` from the lat/long fixes.

umpost/um2netcdf.py
```diff
@@ -736,6 +727,41 @@ def fix_level_coord(cube, z_rho, z_theta, tol=1e-6):
...
+    pressure.attributes['positive'] = 'down'
+    pressure.convert_units('Pa')
+
+    # Round small fractions otherwise coordinates are off by 1e-10 in ncdump output
+    pressure.points = np.round(pressure.points, decimals)
+
+    if pressure.points[0] < pressure.points[-1]:
+        # Flip to get pressure decreasing as per CMIP6 standard
+        # NOTE: returns a new cube!
+        return iris.util.reverse(cube, 'pressure')
```
marc-white, line 760: Is it possible to ever get a cube where the pressure data isn't monotonic for some reason (including user stupidity)?

test/test_um2netcdf.py
```diff
@@ -740,3 +749,54 @@ def test_fix_level_coord_skipped_if_no_levels(z_sea_rho_data, z_sea_theta_data):
...
+# tests - fix pressure level data
+
+def test_fix_pressure_levels_no_pressure_coord(get_fake_cube_coords):
+    cube = get_fake_cube_coords()
+
+    with pytest.raises(iris.exceptions.CoordinateNotFoundError):
+        cube.coord("pressure")  # ensure missing 'pressure' coord
+
+    um2nc.fix_pressure_levels(cube)  # should just exit
```
marc-white, line 764: Should the `None` return be asserted here?

test/test_um2netcdf.py
```diff
@@ -744,6 +753,57 @@ def test_fix_level_coord_skipped_if_no_levels(z_sea_rho_data, z_sea_theta_data):
...
+    extra = {"pressure": m_pressure}
+    cube = get_fake_cube_coords(extra)
+
+    with mock.patch("iris.util.reverse"):
+        um2nc.fix_pressure_levels(cube)
+
+    # TODO: test flaw, this verifies pressure coord but ignores fix_pressure_levels()
+    #       returning a new cube if the pressure is reversed. This is verified
+    #       in command line testing though
+    c_pressure = cube.coord('pressure')
```
blimlim, line 802: Is it worth checking whether the mocked `iris.util.reverse()` is called here, or does that not really add anything?

blimlim (approved):
Changes look good to me. One small question but not really a necessary change

test/test_um2netcdf.py
blimlim: Sounds like a good idea!

umpost/um2netcdf.py
truth-quark: I'm not sure if the data is always monotonic. 

In the "aiihca.paa1jan" dataset (`gadi:/g/data/tm70/esm15_testing/sample_output/aiihca.paa1jan`), extracting all the pressure coords from 100+ cubes gives 11 copies of these points for `pressure`:

```python
array([   1.,    5.,   10.,   20.,   30.,   50.,   70.,  100.,  150.,
         200.,  250.,  300.,  400.,  500.,  600.,  700.,  850.,  925.,
        1000.])
```
@MartinDix do you know if there is any guarantee that the pressure data is monotonic?

truth-quark: Let's throw out an idea for stepwise simplification. The 1st goal for `um2nc-standalone` is to modularise & retrofit unit tests to the code. Changing/extending behaviour is a secondary step, made easier once tests exist. Therefore, an option for this PR/feature is:

* (Optional) add a `TODO: add monotonic test func` reminder to the code
* Create GitHub issue specifically for addressing the monotonic test (do here, add data preprocessing script?)
* Verify this extracted func matches the existing behaviour
* Merge this smaller work unit

How does that sound?

test/test_um2netcdf.py
truth-quark: Yes it should, since the fix pressure levels calling code is relying on a `cube` or a value to resolve as `False`.

Good catch!

truth-quark: Fixed in dd9c18dd.

umpost/um2netcdf.py
truth-quark: Ooops, a `monotonic` check function is included in `iris`:
https://github.com/SciTools/iris/blob/main/lib/iris/util.py#L615

marc-white: Do we know if that check function gets applied to enforce the data to be monotonic, or is it just something we can use in tests?

test/test_um2netcdf.py
```diff
@@ -744,6 +753,58 @@ def test_fix_level_coord_skipped_if_no_levels(z_sea_rho_data, z_sea_theta_data):
...
+    assert c_pressure.attributes["positive"] == "down"
+    assert all(c_pressure.points == [0.0, 1.0])
```
truth-quark, line 805: NB: this is a BUG, the pressure points should be reversed

umpost/um2netcdf.py
```diff
@@ -734,6 +725,41 @@ def fix_level_coord(cube, z_rho, z_theta, tol=1e-6):
...
```
truth-quark, line 760: I noticed line 760 is using the entire `cube` in reversing the pressure levels:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/dd9c18dd1ad709bd7c3dd839d3eb74bc6ebbfae4/umpost/um2netcdf.py#L757-L760

However, `iris.util.reverse()` takes `cube`s or `array`s:
https://github.com/SciTools/iris/blob/e1f5a8b880e443c58cba7770a1ffd32e067a64f2/lib/iris/util.py#L518-L524 (preview is failing....)

I'm experimenting with modifying the pressure array order in place, in the event the coord can be updated within the `cube`. It's currently breaking...

@MartinDix from your knowledge of `iris` do you know if modifying the `pressure.points` array is recommended practice?

```python
    pressure = cube.coord('pressure')

    # snip processing...

    if pressure.points[0] < pressure.points[-1]:
        # Flip to get pressure decreasing as per CMIP6 standard
        # return iris.util.reverse(cube, 'pressure')  # ORIG
        pressure.points = np.flip(pressure.points)  # TODO: is this valid???
```

umpost/um2netcdf.py
truth-quark: `monotonic()` isn't called directly in `iris.util.reverse()`:
https://github.com/SciTools/iris/blob/e1f5a8b880e443c58cba7770a1ffd32e067a64f2/lib/iris/util.py#L518-L612

umpost/um2netcdf.py
blimlim: Just to jump in with some extra info, let me know if I've misunderstood the question though! My understanding is that the pressure coordinate's points ordering shouldn't change separately from of the rest of the cube's data.

My (rough) understanding of a typical `iris.cube` is that it holds an n-dimensional array as it's `data` attribute, and for each dimension has a separate coordinate array. The points in the coordinate arrays specify the locations corresponding to the main `data`. E.g. a cube describing surface temperature might have a 180x360 array of values, plus a length 180 `latitude` coordinate array, and length 360 longitude coordinate array. 

The temperature given by `cube.data[10,55]` would correspond to the location specified by `cube.coord("latitude")[10]` and `cube.coord("latitude")[55]`. But this means if we flip one of the coordinates independently of the data, the wrong locations will be given for the cube data.

It looks like `iris.util.reverse(cube, 'pressure')` is flipping both the `pressure` coordinate, and the actual cube's `data`, so that the two still line up:

```python
> cube_on_plev = cubes[142] # Get a cube with a pressure coordinate

# extract single latitude and longitude point, so that vertical coordinate pressure is the only coordinate
> single_cell_cube = cube_on_plev.extract(iris.Constraint(latitude=0.625, longitude = 0.9375))

# Print out initial pressure points and data array
> single_cell_cube.coord("pressure").points

array([   1.,    5.,   10.,   20.,   30.,   50.,   70.,  100.,  150.,
        200.,  250.,  300.,  400.,  500.,  600.,  700.,  850.,  925.,
       1000.])

> single_cell_cube.data
array([2.23973939e-04, 1.04892252e-02, 5.33754104e-02, 2.66307022e-01,
       9.27435023e-01, 6.87145581e+00, 3.06664567e+01, 5.38439087e+01,
       5.30945653e+01, 4.83872854e+01, 2.50941701e+01, 1.22997686e+01,
       9.74147291e+00, 2.95951569e+01, 5.42121837e+01, 5.50435870e+01,
       5.61263122e+01, 8.55014709e+01, 7.11272948e+01])

# Reverse cube according to pressure coordinate
> single_cell_cube_reverse = iris.util.reverse(single_cell_cube, single_cell_cube.coord("pressure"))

# Print out new data and pressure coordinate arrays
> single_cell_cube_reverse.coord("pressure").points
array([1000.,  925.,  850.,  700.,  600.,  500.,  400.,  300.,  250.,
        200.,  150.,  100.,   70.,   50.,   30.,   20.,   10.,    5.,
          1.])

>single_cell_cube_reverse.data
array([7.11272948e+01, 8.55014709e+01, 5.61263122e+01, 5.50435870e+01,
       5.42121837e+01, 2.95951569e+01, 9.74147291e+00, 1.22997686e+01,
       2.50941701e+01, 4.83872854e+01, 5.30945653e+01, 5.38439087e+01,
       3.06664567e+01, 6.87145581e+00, 9.27435023e-01, 2.66307022e-01,
       5.33754104e-02, 1.04892252e-02, 2.23973939e-04])

# Both have flipped!
```

umpost/um2netcdf.py
marc-white: The test condition here probably needs to be updated to use `monotonic()` then - until we can confirm the pressure values are guaranteed to be monotonic, the current test doesn't actually confirm the direction of the pressure values.

truth-quark: > Do we know if that check function gets applied to enforce the data to be monotonic, or is it just something we can use in tests?

As a standalone function, we can use it in both cases.

test/test_um2netcdf.py
truth-quark: Changed in 1941c693.

umpost/um2netcdf.py
truth-quark: Added TODO to consider monotonic checking in f6ba005c.

truth-quark: I've deferred this to #107 to help bring the rest of the work into the `main` branch.

test/test_um2netcdf.py
truth-quark: Issue #63 covers this task.

umpost/um2netcdf.py
truth-quark: Ok, that makes sense.

I've skipped the test on account of problems at the um2nc & iris "boundary". Implementing a mocked `reverse()` requires faffing with return data structures OR using the real function & making a complicated custom cube arg. Both options are ugly for testing a _one liner_ call to an external library.

I can leave the skipped test in for future work OR delete it & comment why the pressure reversal isn't tested. Does anyone (@blimlim @marc-white  have a preference either way?

truth-quark:
This one needs a final review before merging. @blimlim @marc-white would either of you mind going over the latest changes & conversation threads?

umpost/um2netcdf.py
marc-white: I think the skipped test is fine for now, but add an issue to deal with it later. It definitely needs to be tested, but data transformations (even ones as basic as flipping) are a massive faff that need to be stepped through very methodically.

marc-white (approved):
I think we're good to go at this point!

umpost/um2netcdf.py
truth-quark: #108 now covers the testing task.

I think deferring is a good step for now, to free effort for modularising & retrofitting tests to other functions.

truth-quark:
> I think we're good to go at this point!

Argh, GitHub seems to need someone else to resolve the conversations before it'll let me merge...

CodeGat (approved):

Issue #16, opened 2024-06-19
Refine command line args


truth-quark:
To help merge runtime option parsing, some `argparse` features can help:

* [ ] Move `include`/`exclude` into mutually exclusive option block & simplify this: https://github.com/ACCESS-NRI/um2nc-standalone/blob/main/src/um2netcdf.py#L294-L297
* [ ] Default the output filename?

PR #96, opened 2024-09-06
Clean, test & refactor 64 to 32 bit conversion step.

### LINKED ISSUE #95: Refactor: `cubewrite()` extract conversion from 64 to 32 bit values
Tasks for this block of code:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/70a890cc9434a501810d549e308668f59d21eb00/umpost/um2netcdf.py#L162-L166

* [x] Extract to function
* [x] Add unit tests
* [X] ~~Does rounding need to be handled explicitly for `float` values?~~ SKIP this, `float32` range is massive
* [ ] Are any `int64` values likely to overflow when converted to `int32`?
  * [X] ~~Can `numpy` be toggled to warn of under/overflows in arrays?~~ (Does not appear to be the case)
* [ ] Move conversion call into `process()`?

truth-quark:
Resolves #95.

This PR is mostly to extract & test 64 to 32 bit down-conversion from `cubewrite()`.

I've added a few things:
* New `data` member to `DummyCube`, so data grids can be attached to imitation cubes (it's simpler to reuse fixtures).
* Unit tests
* Added integer overflow warning (just in case)

I'll defer moving this step into `process()` until `cubewrite()` is deconstructed.

Comments welcome :-)

### DIFF
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -744,6 +744,7 @@ def test_fix_level_coord_skipped_if_no_levels(z_sea_rho_data, z_sea_theta_data):
 
 
 # 64 to 32 bit data conversion tests
+# NB: skip float63 to float32 overflow as float32 min/max is  -/+ 3.40e+38
 
 def test_convert_32_bit_with_int64(ua_plev_cube):
     array = np.array([100, 10, 1, 0, -10], dtype=np.int64)
@@ -774,7 +775,7 @@ def test_convert_32_bit_underflow_with_int64(ua_plev_cube):
     assert ua_plev_cube.data.dtype == np.int32
 
 
-def test_64_to_32_float(ua_plev_cube):
+def test_convert_32_bit_with_float64(ua_plev_cube):
     array = np.array([300.33, 30.456, 3.04, 0.0, -30.667], dtype=np.float64)
     ua_plev_cube.data = array
     um2nc.convert_32_bit(ua_plev_cube)
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -1,6 +1,7 @@
 import unittest.mock as mock
 from dataclasses import dataclass
 from collections import namedtuple
+import operator
 
 import umpost.um2netcdf as um2nc
 
@@ -743,38 +744,23 @@ def test_fix_level_coord_skipped_if_no_levels(z_sea_rho_data, z_sea_theta_data):
     um2nc.fix_level_coord(m_cube, z_sea_rho_data, z_sea_theta_data)
 
 
-# 64 to 32 bit data conversion tests
-# NB: skip float63 to float32 overflow as float32 min/max is  -/+ 3.40e+38
-
-def test_convert_32_bit_with_int64(ua_plev_cube):
-    array = np.array([100, 10, 1, 0, -10], dtype=np.int64)
-    ua_plev_cube.data = array
+# int64 to int32 data conversion tests
+# NB: skip float64 to float32 overflow as float32 min/max is huge: -/+ 3.40e+38
+@pytest.mark.parametrize("array,_operator,bound",
+                         [([100, 10, 1, 0, -10], None, None),
+                          ([3000000000], operator.gt, np.iinfo(np.int32).max),
+                          ([-3000000000], operator.lt, np.iinfo(np.int32).min)])
+def test_convert_32_bit(ua_plev_cube, array, _operator, bound):
+    ua_plev_cube.data = np.array(array, dtype=np.int64)
     um2nc.convert_32_bit(ua_plev_cube)
-    assert ua_plev_cube.data.dtype == np.int32
 
-
-def test_convert_32_bit_overflow_with_int64(ua_plev_cube):
-    array = np.array([3000000000], dtype=np.int64)
-    assert array[0] > np.iinfo(np.int32).max
-    ua_plev_cube.data = array
-
-    with pytest.warns():
-        um2nc.convert_32_bit(ua_plev_cube)
+    if _operator:
+        assert _operator(array[0], bound)
 
     assert ua_plev_cube.data.dtype == np.int32
 
 
-def test_convert_32_bit_underflow_with_int64(ua_plev_cube):
-    array = np.array([-3000000000], dtype=np.int64)
-    assert array[0] < np.iinfo(np.int32).max
-    ua_plev_cube.data = array
-
-    with pytest.warns():
-        um2nc.convert_32_bit(ua_plev_cube)
-
-    assert ua_plev_cube.data.dtype == np.int32
-
-
+# test float conversion separately, otherwise parametrize block is ugly
 def test_convert_32_bit_with_float64(ua_plev_cube):
     array = np.array([300.33, 30.456, 3.04, 0.0, -30.667], dtype=np.float64)
     ua_plev_cube.data = array
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -397,6 +397,7 @@ class DummyCube:
         self.standard_name = None
         self.long_name = None
         self.coord = {}
+        self.data = None
 
     def name(self):
         # mimic iris API
@@ -740,3 +741,19 @@ def test_fix_level_coord_skipped_if_no_levels(z_sea_rho_data, z_sea_theta_data):
     m_cube = mock.Mock(iris.cube.Cube)
     m_cube.coord.side_effect = iris.exceptions.CoordinateNotFoundError
     um2nc.fix_level_coord(m_cube, z_sea_rho_data, z_sea_theta_data)
+
+
+# 64 to 32 bit data conversion tests
+
+def test_64_to_32_int(ua_plev_cube):
+    array = np.array([100, 10, 1, 0, -10], dtype=np.int64)
+    ua_plev_cube.data = array
+    um2nc.convert_32_bit(ua_plev_cube)
+    assert ua_plev_cube.data.dtype == np.int32
+
+
+def test_64_to_32_float(ua_plev_cube):
+    array = np.array([300.33, 30.456, 3.04, 0.0, -30.667], dtype=np.float64)
+    ua_plev_cube.data = array
+    um2nc.convert_32_bit(ua_plev_cube)
+    assert ua_plev_cube.data.dtype == np.float32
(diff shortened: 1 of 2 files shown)


test/test_um2netcdf.py
```diff
@@ -740,3 +741,42 @@ def test_fix_level_coord_skipped_if_no_levels(z_sea_rho_data, z_sea_theta_data):
...
+
+def test_convert_32_bit_underflow_with_int64(ua_plev_cube):
+    array = np.array([-3000000000], dtype=np.int64)
+    assert array[0] < np.iinfo(np.int32).max
+    ua_plev_cube.data = array
+
+    with pytest.warns():
+        um2nc.convert_32_bit(ua_plev_cube)
+
+    assert ua_plev_cube.data.dtype == np.int32
```
marc-white, line 775: You could save some repeated lines by using parametrize to run the same test on both the overflow and underflow array... would need to figure out how to tell it which way around to do the `np.iinfo` comparison though.

test/test_um2netcdf.py
```diff
@@ -740,3 +741,42 @@ def test_fix_level_coord_skipped_if_no_levels(z_sea_rho_data, z_sea_theta_data):
...
+
+
+def test_convert_32_bit_with_float64(ua_plev_cube):
+    array = np.array([300.33, 30.456, 3.04, 0.0, -30.667], dtype=np.float64)
+    ua_plev_cube.data = array
+    um2nc.convert_32_bit(ua_plev_cube)
+    assert ua_plev_cube.data.dtype == np.float32
```
marc-white, line 782: Likewise, you could merge this test with `test_convert_32_bit_with_int64`, and pass through the arrays and data types via parametrize

test/test_um2netcdf.py
marc-white: To answer my own question, you can use the `operator` library to pass things like `>` and `<` as parameters: https://stackoverflow.com/questions/71730346/is-it-possible-to-pass-comparison-operators-as-parameters-into-functions

test/test_um2netcdf.py
```diff
@@ -740,3 +741,42 @@ def test_fix_level_coord_skipped_if_no_levels(z_sea_rho_data, z_sea_theta_data):
...
+
+def test_convert_32_bit_with_int64(ua_plev_cube):
+    array = np.array([100, 10, 1, 0, -10], dtype=np.int64)
+    ua_plev_cube.data = array
+    um2nc.convert_32_bit(ua_plev_cube)
+    assert ua_plev_cube.data.dtype == np.int32
+
+
+def test_convert_32_bit_overflow_with_int64(ua_plev_cube):
+    array = np.array([3000000000], dtype=np.int64)
```
marc-white, line 757: Once parametrized, you can add the test case that is just outside the limit in addition to this one

test/test_um2netcdf.py
```diff
@@ -740,3 +741,42 @@ def test_fix_level_coord_skipped_if_no_levels(z_sea_rho_data, z_sea_theta_data):
...
+    assert array[0] > np.iinfo(np.int32).max
+    ua_plev_cube.data = array
+
+    with pytest.warns():
+        um2nc.convert_32_bit(ua_plev_cube)
+
+    assert ua_plev_cube.data.dtype == np.int32
```
marc-white, line 764: Does the out-of-bounds conversion give a predictable result, and if so, can you compare to ensure that result is what comes out?

umpost/um2netcdf.py
```diff
@@ -736,6 +733,34 @@ def fix_level_coord(cube, z_rho, z_theta, tol=1e-6):
...
+    if cube.data.dtype == 'float64':
+        cube.data = cube.data.astype(np.float32)
+    elif cube.data.dtype == 'int64':
+        _max = np.max(cube.data)
+        _min = np.min(cube.data)
+
+        if _max > MAX_NP_INT32:
+            msg = f"Converting {cube.var_name} causes a 32 bit overflow!"
+            warnings.warn(msg)
+        elif _min < MIN_NP_INT32:
```
marc-white, line 757: Should this be an `if` condition instead of `elif`? Hypothetically, an array could create both an overflow and underflow condition...

umpost/um2netcdf.py
```diff
@@ -736,6 +733,34 @@ def fix_level_coord(cube, z_rho, z_theta, tol=1e-6):
...
+MIN_NP_INT32 = np.iinfo(np.int32).min
+
+
+def convert_32_bit(cube):
+    """
+    Convert 64 bit int/float data to 32 bit (in place).
+
+    Parameters
+    ----------
+    cube : iris.cube object to modify.
```
marc-white, line 746:
```suggestion
    cube : iris.cube object to modify.

Warns
-----
    TBD : If the input cube has values beyond 32-bit limits, causing an overflow.
```

See comment below on the warning to replace TBD.

umpost/um2netcdf.py
```diff
@@ -736,6 +733,34 @@ def fix_level_coord(cube, z_rho, z_theta, tol=1e-6):
...
+    """
+    if cube.data.dtype == 'float64':
+        cube.data = cube.data.astype(np.float32)
+    elif cube.data.dtype == 'int64':
+        _max = np.max(cube.data)
+        _min = np.min(cube.data)
+
+        if _max > MAX_NP_INT32:
+            msg = f"Converting {cube.var_name} causes a 32 bit overflow!"
+            warnings.warn(msg)
```
marc-white, line 756: It would be best to explicitly define which warning class to return. The default `UserWarning` is probably OK, although `RuntimeWarning` may be more correct.

truth-quark: The thought occurred that `numpy` might include an option to toggle overflow warnings. This is based on seeing `numpy 2.x` raise emit more warnings than some code run against `v1.22`. I'll check the docs it'd be great to reuse existing warnings.

truth-quark: A quick search shows auto detecting & warning of integer overflow in arrays is still an open issue:
https://github.com/numpy/numpy/issues/8987

I'll treat this as a case requiring manual checks & emitting a relevant warning. Performance is a possible issue, although I'm deliberately just checking the max value for some efficiency (I'm unsure if `numpy` caches min/max values). Some other ways around this are:
* review min/max values for each data variable
  - Are any vars outside `-/+ 2**31` (apply overflow checks on selected vars)
  - Detect large or odd values during data reads/sanity checking process?
  - Could be a lot of work...
* deliberately skip warnings via a new option

marc-white: I think checking the min/max for now is satisfactory. It would also be good to have the ability to turn off the warnings if you're certain you don't need them/you don't care.

truth-quark: > It would be best to explicitly define which warning class to return. The default `UserWarning` is probably OK, although `RuntimeWarning` may be more correct.

According to https://docs.python.org/3.10/library/exceptions.html#RuntimeWarning, the Runtime one is a _Base class for warnings about dubious runtime behavior_. That's pretty apt to me :-)

umpost/um2netcdf.py
truth-quark: I've fixed this in d282d45a.

test/test_um2netcdf.py
truth-quark: Parameterised tests  are in with b21698af, which chops out a fair bit of repeated logic.

test/test_um2netcdf.py
truth-quark: That's also fixed in b21698af.

umpost/um2netcdf.py
truth-quark: Good idea, it's fixed in 9bd6f270.

umpost/um2netcdf.py
truth-quark: > It would be best to explicitly define which warning class to return. The default `UserWarning` is probably OK, although `RuntimeWarning` may be more correct.

I've fixed this in 15274011 to use `RuntimeWarning`.

truth-quark: > I think checking the min/max for now is satisfactory. It would also be good to have the ability to turn off the warnings if you're certain you don't need them/you don't care.

I've expanded #16 to cover this/fixing command line issues as a whole.

test/test_um2netcdf.py
truth-quark: The under/overflow result should be predictable.

Technically, the under/overflow result is junk data, so I can't think of what is gained by testing the end result. Also, it adds something else to the intent of the test.

One simple solution is updating the under/overflow warnings to recommend the `--64` cmd line option:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/9bd6f270005518060b37d0bb77e4fbf30c4bf672/umpost/um2netcdf.py#L779-L780
That places the onus on the end user. How does that appeal to you?

blimlim (dismissed):
Everything looks good to me!

test/test_um2netcdf.py
marc-white: The only reason you'd test the junk data result is to make sure the junk remains the same, i.e., numpy versions haven't changed how it operates. That being said, I think the `--64` cmd line option does the trick.

truth-quark: I've fixed adding `--64` in 6c02f7d9.

truth-quark:
@marc-white did you want to do a final skim/review of the "end" product?

umpost/um2netcdf.py
truth-quark: This is fixed too.

marc-white:
Looks good to me. I just tried to resolve the last outstanding comment, but Github is being difficult... maybe the VSCode extension can save me...

test/test_um2netcdf.py
marc-white: OK, let's see if I can close this out....

marc-white (approved):
Improvements looks good, should be set!

truth-quark:
> Looks good to me. I just tried to resolve the last outstanding comment, but Github is being difficult... maybe the VSCode extension can save me...

I could close the convo off...

truth-quark:
Awesome, I think closing the open conversation thread off removed review approvals. @marc-white or @blimlim wouldeither of you mind doing another approval to enable a merge?

blimlim (approved):
Looks good!

Issue #98, opened 2024-09-09
Improve warning behaviour in `set_item_codes()`


truth-quark:
Currently `um2netcdf.set_item_codes()` skips setting/overwriting the `stash code` if one already exists. This causes several warnings in the unit tests, which would be good to clean up.

There's several options to fix:
* Simply overwrite the stash code
* Determine which tests reassign stash codes & decide on best action
* Something else?

truth-quark:
Here's a ~ September 2024 variant of `set_item_codes()`:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/c1a4a54404347ccba925a5c18249049482309caa/umpost/um2netcdf.py#L642-L651

Line 650 retrieves & converts _cube sourced stash codes_ to item codes.
Line 651 adds this as an attribute.

I think the warning implementation was an overly defensive approach while early test coverage was sparse & less was known about the codebase. Setting/adding the `item_code` from the cube's _own attributes_ should be safe, it's always setting the same value again unless attributes are modified during runtime (this would have to be external to `um2netcdf.py`). The warning doesn't seem to serve a purpose now.

Proposed fix:
* Delete the check & warning from `set_item_codes()`
* Add `set_item_codes()` docstrings
* In the tests:
  * Delete `def test_set_item_codes()`
  * Delete `test_set_item_codes_avoid_overwrite()`

Tests for `process()` call `set_item_codes()`, therefore I suggest this covers testing item codes within more general test cases.

Are there any comments on this approach or something I've missed?

truth-quark:
An initial implementation is on a new branch https://github.com/ACCESS-NRI/um2nc-standalone/commit/1ce80498714dce8155f2d109ad68298e68aa251f

PR #113, opened 2024-09-26
Refactor cube testing objects/mocking with `DummyCube`

### LINKED ISSUE #69: Refactor coord mocks when #63/lat long fixes is merged

truth-quark:
This closes #69.

This PR covers multiple elements to switch tests to using `DummyCube`:
* Replaces several `Mock` objs masquerading as cubes with `DummyCube`
* Removes old setup cruft
* Cleans some test fixtures
* Adds coords to other test fixtures (fixtures better open for reuse)
* Removes `mock.patch("umpost.um2netcdf.fix_latlon_coords")` in `process()` tests
* BONUS: removes `mock.patch("umpost.um2netcdf.fix_level_coord")` in `process()` tests

Any comments welcome!

### DIFF
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -997,16 +997,16 @@ def test_fix_level_coord_modify_cube_with_rho(level_coords_cube,
     # verify cube renaming with appropriate z_rho data
     cube = level_coords_cube
 
-    assert cube.coord(um2nc.MODEL_LEVEL_NUM).var_name != "model_rho_level_number"
-    assert cube.coord(um2nc.LEVEL_HEIGHT).var_name != "rho_level_height"
-    assert cube.coord(um2nc.SIGMA).var_name != "sigma_rho"
+    assert cube.coord(um2nc.MODEL_LEVEL_NUM).var_name != um2nc.MODEL_RHO_LEVEL
+    assert cube.coord(um2nc.LEVEL_HEIGHT).var_name != um2nc.RHO_LEVEL_HEIGHT
+    assert cube.coord(um2nc.SIGMA).var_name != um2nc.SIGMA_RHO
 
     rho = np.ones(z_sea_theta_data.shape) * level_heights[0]
     um2nc.fix_level_coord(cube, rho, z_sea_theta_data)
 
-    assert cube.coord(um2nc.MODEL_LEVEL_NUM).var_name == "model_rho_level_number"
-    assert cube.coord(um2nc.LEVEL_HEIGHT).var_name == "rho_level_height"
-    assert cube.coord(um2nc.SIGMA).var_name == "sigma_rho"
+    assert cube.coord(um2nc.MODEL_LEVEL_NUM).var_name == um2nc.MODEL_RHO_LEVEL
+    assert cube.coord(um2nc.LEVEL_HEIGHT).var_name == um2nc.RHO_LEVEL_HEIGHT
+    assert cube.coord(um2nc.SIGMA).var_name == um2nc.SIGMA_RHO
 
 
 def test_fix_level_coord_modify_cube_with_theta(level_heights,
@@ -1017,9 +1017,9 @@ def test_fix_level_coord_modify_cube_with_theta(level_heights,
     cube = level_coords_cube
     um2nc.fix_level_coord(cube, z_sea_rho_data, z_sea_theta_data)
 
-    assert cube.coord(um2nc.MODEL_LEVEL_NUM).var_name == "model_theta_level_number"
-    assert cube.coord(um2nc.LEVEL_HEIGHT).var_name == "theta_level_height"
-    assert cube.coord(um2nc.SIGMA).var_name == "sigma_theta"
+    assert cube.coord(um2nc.MODEL_LEVEL_NUM).var_name == um2nc.MODEL_THETA_LEVEL_NUM
+    assert cube.coord(um2nc.LEVEL_HEIGHT).var_name == um2nc.THETA_LEVEL_HEIGHT
+    assert cube.coord(um2nc.SIGMA).var_name == um2nc.SIGMA_THETA
 
 
 def test_fix_level_coord_skipped_if_no_levels(z_sea_rho_data, z_sea_theta_data):
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -314,20 +314,6 @@ def test_stash_code_to_item_code_conversion():
     assert result == 30255
 
 
-@dataclass(frozen=True)
-class DummyStash:
-    """
-    Partial Stash representation for testing.
-    """
-    section: int
-    item: int
-
-
-def add_stash(cube, stash):
-    d = {um2nc.STASH: stash}
-    setattr(cube, "attributes", d)
-
-
 def test_set_item_codes():
     cube0 = DummyCube(1002, "d0", {um2nc.STASH: DummyStash(1, 2)})
     cube1 = DummyCube(3004, "d1", {um2nc.STASH: DummyStash(3, 4)})
@@ -341,6 +327,31 @@ def test_set_item_codes():
 
     assert c0.item_code == 1002
     assert c1.item_code == 3004
+
+
+def test_set_item_codes_avoid_overwrite():
+    item_code = 1007
+    item_code2 = 51006
+
+    cubes = [DummyCube(item_code, "fake_var"), DummyCube(item_code2, "fake_var2")]
+    um2nc.set_item_codes(cubes)
+    assert cubes[0].item_code == item_code
+    assert cubes[1].item_code == item_code2
+
+
+@dataclass(frozen=True)
+class DummyStash:
+    """
+    Partial Stash representation for testing.
+    """
+    section: int
+    item: int
+
+
+def add_stash(cube, stash):
+    d = {um2nc.STASH: stash}
+    setattr(cube, "attributes", d)
+
 
 
 class DummyCube:
@@ -375,16 +386,6 @@ class DummyCube:
         except KeyError:
             msg = f"{self.__class__}[{self.var_name}]: lacks coord for '{_name}'"
             raise CoordinateNotFoundError(msg)
-
-
-def test_set_item_codes_avoid_overwrite():
-    item_code = 1007
-    item_code2 = 51006
-
-    cubes = [DummyCube(item_code, "fake_var"), DummyCube(item_code2, "fake_var2")]
-    um2nc.set_item_codes(cubes)
-    assert cubes[0].item_code == item_code
-    assert cubes[1].item_code == item_code2
 
 
 @pytest.fixture
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -445,10 +445,8 @@ def test_cube_filtering_no_include_exclude(ua_plev_cube, heaviside_uv_cube):
 # cube variable renaming tests
 @pytest.fixture
 def x_wind_cube():
-    x_wind_cube = DummyCube(2, var_name="var_name",
-                            attributes={'STASH': DummyStash(0, 2)})
+    x_wind_cube = DummyCube(2, var_name="var_name")
     x_wind_cube.standard_name = "x_wind"
-    x_wind_cube.cell_methods = []
     return x_wind_cube
 
 
@@ -510,9 +508,8 @@ def test_fix_standard_name_update_x_wind(x_wind_cube):
 def test_fix_standard_name_update_y_wind():
     # test cube wind renaming block only
     # use empty std name to bypass renaming logic
-    m_cube = DummyCube(3, attributes={'STASH': DummyStash(0, 3)})
+    m_cube = DummyCube(3)
     m_cube.standard_name = "y_wind"
-    m_cube.cell_methods = []
 
     um2nc.fix_standard_name(m_cube, "", verbose=False)
     assert m_cube.standard_name == "northward_wind"
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -1041,15 +1041,7 @@ def test_fix_pressure_levels_no_pressure_coord(level_coords_cube):
     assert um2nc.fix_pressure_levels(cube) is None  # should just exit
 
 
-def _add_attrs_points(m_plevs: mock.MagicMock, points):
-    # NB: iris attributes appear to be added via mixins, so it's easier but
-    #     less desirable to rely on mock attrs here
-    setattr(m_plevs, "attributes", {"positive": None})
-    setattr(m_plevs, "points", points)
-
-
 def test_fix_pressure_levels_do_rounding():
-    # init pressure coordinate
     pressure = iris.coords.DimCoord([1.000001, 0.000001],
                                     var_name="pressure",
                                     units="Pa",
@@ -1066,15 +1058,15 @@ def test_fix_pressure_levels_do_rounding():
 
 
 @pytest.mark.skip
-def test_fix_pressure_levels_reverse_pressure(level_coords_cube):
+def test_fix_pressure_levels_reverse_pressure():
     # TODO: test is broken due to fiddly mocking problems (see below)
+    pressure = iris.coords.DimCoord([0.000001, 1.000001],
+                                    var_name="pressure",
+                                    units="Pa",
+                                    attributes={"positive": None})
 
-    m_pressure = mock.Mock()
-    # m_pressure.ndim = 1
-    _add_attrs_points(m_pressure, [0.000001, 1.000001])
-    extra = {"pressure": m_pressure}
-    cube = get_fake_cube_coords(extra)
-    # cube.ndim = 3
+    cube = DummyCube(1, coords=[pressure])
+    cube.ndim = 3
 
     # TODO: testing gets odd here at the um2nc & iris "boundary":
     #   * A mock reverse() needs to flip pressure.points & return a modified cube.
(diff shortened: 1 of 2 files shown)


test/test_um2netcdf.py
```diff
@@ -70,46 +70,28 @@ def mule_vars(z_sea_rho_data, z_sea_theta_data):
...
 @pytest.fixture
-def air_temp_cube():
-    # data copied from aiihca.paa1jan.subset file
-    name = "air_temperature"
-    m_air_temp = mock.NonCallableMagicMock(spec=iris.cube.Cube, name=name)
-    set_default_attrs(m_air_temp, 30204, name)
-    return m_air_temp
+def air_temp_cube(lat_river_coord, lon_river_coord):
+    # details copied from aiihca.paa1jan.subset file
+    air_temp = DummyCube(30204, "air_temperature",
```
marc-white, line 76: Do these codes have a specific meaning, or do you just need a distinct code for each cube type? If the former, it might benefit from using a reference `dict` or similar rather than having magic numbers.

marc-white (approved):
Looks pretty straightforward to me.

My main comment (which only just occurred to me, and clearly is an existing issue) is that a lot of the test `assert` statements don't have an explanatory note after them, e.g.:
```python
assert False, "Well this will obviously fail"
```
It's nice to get an explanation out of `pytest` as to why a test failed, rather than having to dig into the test code itself to figure out what's going on. Up to you whether this is dealt with as part of this PR, or as a new issue.

test/test_um2netcdf.py
truth-quark: The codes are in the `umpost/stashvar_cmip6.py` module & I've copied them across.

A reference/index dict or constants for common stash codes could certainly help...

truth-quark:
> My main comment (which only just occurred to me, and clearly is an existing issue) is that a lot of the test `assert` statements don't have an explanatory note after them, e.g.:
> 
> ```python
> assert False, "Well this will obviously fail"
> ```
> 
> It's nice to get an explanation out of `pytest` as to why a test failed, rather than having to dig into the test code itself to figure out what's going on. Up to you whether this is dealt with as part of this PR, or as a new issue.

Some of this is related to the process of retro-fitting testing & the way the code operates. Quite often, a function takes inputs, modifies a cube & returns None. Error messages could make it clearer if a value hasn't been set properly.

This relates to another element: being unfamiliar with _why_ steps are done, thus knowing _what_ meaningful error msg to provide.

There's also a future refactoring step  where I suspect simplification could result in shrinking the code & tests. At this stage I'm mainly focusing on code coverage to assist with the rework steps.

test/test_um2netcdf.py
```diff
@@ -1017,67 +974,52 @@ def test_fix_cell_methods_keep_weeks():
...
-    cube = get_fake_cube_coords()
+    cube = level_coords_cube
     um2nc.fix_level_coord(cube, z_sea_rho_data, z_sea_theta_data)
 
-    assert cube.coord(um2nc.MODEL_LEVEL_NUM).var_name == "model_theta_level_number"
-    assert cube.coord(um2nc.LEVEL_HEIGHT).var_name == "theta_level_height"
-    assert cube.coord(um2nc.SIGMA).var_name == "sigma_theta"
+    assert cube.coord(um2nc.MODEL_LEVEL_NUM).var_name == um2nc.MODEL_THETA_LEVEL_NUM
+    assert cube.coord(um2nc.LEVEL_HEIGHT).var_name == um2nc.THETA_LEVEL_HEIGHT
+    assert cube.coord(um2nc.SIGMA).var_name == um2nc.SIGMA_THETA
```
blimlim, line 1023: There seen to be some subtle differences between iris cubes and our `DummyCube` which would cause the  `test_fix_level_coord_...` tests to fail if we were using real iris cubes.

The `cube.coord()` method uses the current name for a `DimCoord` as its argument, where the "current name" comes from either the `DimCoord`'s `standard_name`, `long_name`,  or `var_name`, with the following order of priority: `standard_name` > `long_name` > `var_name`.

This means that if a `DimCoord` only has a `var_name` attribute, changing the `var_name` forces us to also change the argument in `cube.coord(...)` used to access it:

A made up example after modifying a loaded cube's coordinate names:
```Python 
>>> cube.coord("pseudo_level").__dict__
{'_metadata_manager': MetadataManager(standard_name=None, long_name=None, var_name='pseudo_level', units=Unit('1'), attributes={}, coord_system=None, climatological=False, circular=False), '_values_dm': DataManager(array([1, 2, 3, 4, 5, 6])), '_bounds_dm': None, '_ignore_axis': False}

>>> cube.coord("pseudo_level").var_name = "new_var_name"

>>> cube.coord("pseudo_level")

  File "<stdin>", line 1, in <module>
  File "/g/data/vk83/apps/payu/1.1.5/lib/python3.10/site-packages/iris/cube.py", line 2304, in coord
    raise iris.exceptions.CoordinateNotFoundError(emsg)
iris.exceptions.CoordinateNotFoundError: "Expected to find exactly 1 'pseudo_level' coordinate, but found none."

>>> cube.coord("new_var_name")
<DimCoord: new_var_name / (1)  [1, 2, ..., 5, 6]  shape(6,)>
```
In these tests however, we're changing the coordinate's `var_names` but not the `cube.coord(...)` arguments used to access them.

Given that the tests are working and it's a fairly subtly difference, should we defer this to a separate issue?

test/test_um2netcdf.py
```diff
@@ -70,46 +70,28 @@ def mule_vars(z_sea_rho_data, z_sea_theta_data):
...
+    # details copied from aiihca.paa1jan.subset file
+    air_temp = DummyCube(30204, "air_temperature",
+                         coords=[lat_river_coord, lon_river_coord])
+    return air_temp
 
 
 @pytest.fixture
-def precipitation_flux_cube():
+def precipitation_flux_cube(lat_river_coord, lon_river_coord):
     # copied from aiihca.paa1jan.subset file
```
blimlim, line 83: This one's bordering on pedantry so it's not a problem if you leave these as is, but the `lat_river` and `lon_river` coordinates aren't used for very many variables, and the `air_temp_cube`, `precipitation_flux_cube`, `geopotential_cube`, heaviside_uv_cube`, `heaviside_t_cube`, `ta_plev_cube` are more likely to use the `lon/lat_standard`, or `lon_u/lat_v` coordinates.

If you think it's worth changing these fixtures, I can look into some model output to see which coordinates these variables use, otherwise it's probably ok to leave as is!

blimlim (changes requested):

test/test_um2netcdf.py
truth-quark: I've added this as an idea to #27.

truth-quark:
> My main comment (which only just occurred to me, and clearly is an existing issue) is that a lot of the test `assert` statements don't have an explanatory note after them, e.g.:
> 
> ```python
> assert False, "Well this will obviously fail"
> ```
> 
> It's nice to get an explanation out of `pytest` as to why a test failed, rather than having to dig into the test code itself to figure out what's going on. Up to you whether this is dealt with as part of this PR, or as a new issue.

I've added this to #27 for feature fixes too.

test/test_um2netcdf.py
```diff
@@ -486,13 +445,11 @@ def test_cube_filtering_no_include_exclude(ua_plev_cube, heaviside_uv_cube):
...
-    fake_cube = PartialCube("var_name", {'STASH': DummyStash(0, 2)}, "x_wind")
-    fake_cube.cell_methods = []
-    return fake_cube
-
-
-# UMStash = namedtuple("UMStash",
-#                      "long_name, name, units, standard_name, uniquename")
+    x_wind_cube = DummyCube(2, var_name="var_name",
+                          attributes={'STASH': DummyStash(0, 2)})
+    x_wind_cube.standard_name = "x_wind"
```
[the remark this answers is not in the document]
truth-quark: Nice, I've fixed this in f7fe5d94.

test/test_um2netcdf.py
```diff
@@ -553,7 +510,8 @@ def test_fix_standard_name_update_x_wind(x_wind_cube):
 def test_fix_standard_name_update_y_wind():
     # test cube wind renaming block only
     # use empty std name to bypass renaming logic
-    m_cube = PartialCube("var_name", {'STASH': DummyStash(0, 3)}, "y_wind")
+    m_cube = DummyCube(3, attributes={'STASH': DummyStash(0, 3)})
+    m_cube.standard_name = "y_wind"
     m_cube.cell_methods = []
```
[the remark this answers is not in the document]
truth-quark: Also fixed this in f7fe5d94. A single test fixture is proving handy.

test/test_um2netcdf.py
truth-quark: The tests are a form of documentation & should aim to reflect real data.

If it only takes a short while to dig out the grid types, swapping fixtures is a matter of a few renames. Otherwise, if it's fiddly/time consuming, I can add a comment to indicate the data is invented & not to copy it. Comments can go unread, so it's not the "best" option.

blimlim: It should be pretty easy to get the grids! I'll do that now

test/test_um2netcdf.py
```diff
@@ -368,22 +329,13 @@ def add_stash(cube, stash):
...
+    cube0 = DummyCube(1002, "d0", {um2nc.STASH: DummyStash(1, 2)})
+    cube1 = DummyCube(3004, "d1", {um2nc.STASH: DummyStash(3, 4)})
     cubes = [cube0, cube1]
 
     for cube in cubes:
-        assert not hasattr(cube, um2nc.ITEM_CODE)
+        assert hasattr(cube, um2nc.ITEM_CODE)
 
     um2nc.set_item_codes(cubes)
     c0, c1 = cubes
```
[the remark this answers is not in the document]
truth-quark: Good catch! The tests have evolved over time, which this bit of refactoring has uncovered.

Issue #98 covers `set_item_codes()`, let's move this thread over there.

test/test_um2netcdf.py
blimlim: I've looked at some output for ESM1.5, and I think we can use:
`heaviside_uv_cube`, `ta_plev_cube`, `air_temp_cube`, and `geopotential_cube` all would use `lat_v_nd` and `lon_u_nd` as their coordinates.

`precipitation_flux_cube` would use `lat_standard_nd` and `lon_standard_nd`.

The same variables from CM2 would use different coordinates, but if we are happy to use examples from ESM1.5 then the above should be ok.

Our ESM1.5 configs don't save the `heaviside_t` variable, but for CM2 `heaviside_t_cube` would use the `lat_standard_eg` and `lon_standard_eg`.

I noticed a couple of small typos in the stash codes. I think `geopotential_cube`  should have stash code `30207` instead of `30297`, and `ta_plev`'s stash code should be `30204` instead of `30294`. Changing the `ta_plev_cube`'s stash code makes it a duplicate of `air_temp_cube` though (other than the name), and so we may be able to remove one of the two.

truth-quark: Several fixes have gone in:
* 3a774ed2 swaps to the other coordinate fixtures & fixes the item codes
* 8cf02c84 removes `air_temp_cube` & replaces with `ta_plev` --> tests still pass.
* Additional code ordering, moving test fixtures near the module's head (bring code block order closer to top to bottom "reading" order)

test/test_um2netcdf.py
truth-quark: Yep, let's defer this to a "cleanup issue" to help focus on finishing off modularisation & the test retrofit.

I'll sort out the new issue...

test/test_um2netcdf.py
```diff
@@ -70,46 +70,86 @@ def mule_vars(z_sea_rho_data, z_sea_theta_data):
...
-@pytest.fixture
-def air_temp_cube():
-    # data copied from aiihca.paa1jan.subset file
-    name = "air_temperature"
-    m_air_temp = mock.NonCallableMagicMock(spec=iris.cube.Cube, name=name)
-    set_default_attrs(m_air_temp, 30204, name)
-    return m_air_temp
+    def __init__(self, item_code, var_name=None, attributes=None,
+                 units=None, coords=None):
+        self.item_code = item_code
```
truth-quark, line 89: Iris cube objects don't have `item_code`. This is a custom attribute added during `um2nc` processing, highlighting differences between real & fake objects in testing. Technically, there's a period where `item_code` does not exist. Do any tests need a review to determine if the `item_code` attr should be present?

This also highlights unavoidable problems with iris: cube objs are large, complex, span multiple modules & have layers of inheritance (lacks modularity). This forces us to use an alternative.

test/test_um2netcdf.py
```diff
@@ -70,46 +70,86 @@ def mule_vars(z_sea_rho_data, z_sea_theta_data):
...
+        self.var_name = var_name or "unknown_var"
+        self.attributes = attributes
+        self.units = units
+        self.standard_name = None
+        self.long_name = ""
+        self.cell_methods = []
+        self.data = None
```
truth-quark, line 96: TODO: reorder the attribute block, gather name attrs etc

test/test_um2netcdf.py
```diff
@@ -70,46 +70,86 @@ def mule_vars(z_sea_rho_data, z_sea_theta_data):
...
-    # data copied from aiihca.paa1jan.subset file
-    name = "air_temperature"
-    m_air_temp = mock.NonCallableMagicMock(spec=iris.cube.Cube, name=name)
-    set_default_attrs(m_air_temp, 30204, name)
-    return m_air_temp
+    def __init__(self, item_code, var_name=None, attributes=None,
+                 units=None, coords=None):
+        self.item_code = item_code
+        self.var_name = var_name or "unknown_var"
+        self.attributes = attributes
```
truth-quark, line 91: TODO: add the attribute fixes (e.g. `self.attrs = attrs or {}`

test/test_um2netcdf.py
blimlim: Apologies, my comment about the typo's in the STASH codes was incorrect!

CM2 does in fact output a `geopotential height` variable with stash code `30297`. The difference between CM2's geopotential height with  stash code `30297` vs ESM1.5's geopotential height with stash code `30207` is the horizontal grid they use. I.e:
- CM2's geopotential height cube with code `30297` would use `lat_standard_eg_coord` and `lon_standard_eg_coord`. [This variable requires the `heaviside_t` cube for masking.](https://github.com/ACCESS-NRI/um2nc-standalone/blob/223c50b5d738d5f455d0e16a9592f8fda81d678f/umpost/um2netcdf.py#L697-L699)
- ESM1.5's geopotential height cube with code `30207` uses `lat_v_nd_coord` and `lon_u_eg_coord`. [This cube requires the `heaviside_uv` cube for masking.](https://github.com/ACCESS-NRI/um2nc-standalone/blob/223c50b5d738d5f455d0e16a9592f8fda81d678f/umpost/um2netcdf.py#L687-L689)


Similar for the `air_temp_cube` vs `ta_plev_cube`:
- CM2 outputs an air temperature cube with stash code `30294` and coordinates `lat_standard_eg_coord`, `lon_standard_eg_coord`. [This cube requires the `heaviside_t` cube for masking.](https://github.com/ACCESS-NRI/um2nc-standalone/blob/223c50b5d738d5f455d0e16a9592f8fda81d678f/umpost/um2netcdf.py#L697-L699)
- ESM1.5 outputs an air temperature cube with code `30204` with coordinates `lat_v_nd_coord` and `lon_u_eg_coord`. [This cube requires the `heaviside_uv` cube for masking](https://github.com/ACCESS-NRI/um2nc-standalone/blob/223c50b5d738d5f455d0e16a9592f8fda81d678f/umpost/um2netcdf.py#L687-L689).

--- 

Because they require different masking cubes (`heaviside_t` vs `heaviside_uv`), the choice of the stash codes in the fixtures impacts the code executed by the `test_process...` tests. https://github.com/ACCESS-NRI/um2nc-standalone/blob/223c50b5d738d5f455d0e16a9592f8fda81d678f/test/test_um2netcdf.py#L181-L296

Based on this, I think it makes sense to:
- Keep `ta_plev_cube` with stash code `30204`, (which requires the `heaviside_uv_cube` mask) as the only air temperature cube. 
- Revert `geopotential_height_cube` to use stash code `30297`, so that it requires the `heaviside_t_cube`.

Which will make sure both the `need_heaviside_uv` and `need_heaviside_t` branches are tested in the process/filtering tests.

Again sorry about this! Let me know if these suggestions sound ok.

truth-quark: All good, mistakes happen. This highlights problems with the stash code module containing duplicate names under different item codes --> it's a source of confusion.

I've reverted the `geo_potential_cube` to its original `item_code` & commented on the cubes in a84ba48f.

test/test_um2netcdf.py
truth-quark: I've raised this in #120 to defer `DummyCube` refactoring to a separate block of work (as this PR is deduplicating cube like test objs).

test/test_um2netcdf.py
truth-quark: Deferred to #120.

test/test_um2netcdf.py
blimlim: Awesome, those changes look good.

Are you happy to also swap the coordinates on the `geopotential_cube` to `lat_standard_eg_coord` and `lon_standard_eg_coord`? (My mistake in the earlier comment too).

truth-quark: Yep, the change is in 8b244b1.

blimlim (approved):
Nice! Looks good to me!

test/test_um2netcdf.py
blimlim: Awesome, thanks!

PR #114, opened 2024-09-27
Extract `cubewrite` fill value modifications

### LINKED ISSUE #99: Extract `cubewrite()` fill value to function
Continuing with the overarching task of splitting `cubewrite()` (#12), fill value functionality can be extracted as a function:

https://github.com/ACCESS-NRI/um2nc-standalone/blob/92096ce86f3cdd305570e940397b06633ae8777b/umpost/um2netcdf.py#L166-L174

Tasks:
* [ ] Extract code block to separate function
* [ ] Add unit tests
* [ ] Convert `1e20` fill value to a constant
* [ ] Add arg to function signature for custom fill values

For now, let's keep `cubewrite()`'s steps the same, merging into `process()` can be a separate refactoring issue.

blimlim:
This pr closes #99.

It extract's the `cubewrite` fill value modifications into a seperate function, adds a `custom_fill_val` optional argument, and adds unit tests.

Any suggestions or ideas are welcome!


blimlim:
There's a slight amount of messiness with the added `custom_fill_val` argument, and so any ideas or suggestions would be great! Specifically, the final line of `fix_fill_value` forces the fill value to have the same type as the `cube.data`, which is [required by the netCDF conventions](https://docs.unidata.ucar.edu/netcdf-c/current/attribute_conventions.html#:~:text=same%20type%20as%20the%20variable.)

To prevent e.g. a float `custom_fill_val` from silently being rounded to an integer, I thought it would be good to raise an error when the types don't match: https://github.com/ACCESS-NRI/um2nc-standalone/blob/f3fbd44dde5b1074e88609ce2940e2c80c021915/umpost/um2netcdf.py#L334-L340

This will also complain e.g. if a `float` was going to be rounded down to a `float32`, however this is stricter than what's applied to the default fill value for floats, which is would get rounded down to a `float32` if the the cube's data is `float32`: https://github.com/ACCESS-NRI/um2nc-standalone/blob/f3fbd44dde5b1074e88609ce2940e2c80c021915/umpost/um2netcdf.py#L342-L350

I don't think it's too important, but mainly wanted to check whether we're happy for the type requirements for the `custom_fill_val` to be stricter than what's applied to the default values?

blimlim:
Ah oops, didn't notice the additional places the `fill_value` is required. Will fix this now

umpost/um2netcdf.py
```diff
@@ -319,6 +321,63 @@ def fix_latlon_coords(cube, grid_type, dlat, dlon):
...
+    cube: Iris cube object (modified in place).
+    custom_fill_value: Fill value to use in place of defaults.
+    Type must match cube data's type.
+    """
+    if custom_fill_value is not None:
+        fill_value = custom_fill_value
+    else:
+        fill_value = get_default_fill_value(cube)
+
+    if type(fill_value) == cube.data.dtype:
```
blimlim, line 366: My linter (flake8) complains here, suggesting `do not compare types, for exact checks use "is" / "is not", for instance checks use "isinstance()"`

I've had a play around, however neither `is` or `isinstance` work as desired when using `cube.data.dtype` in the comparison. Would be keen to gather any suggestions!

truth-quark: > ... neither `is` or `isinstance` work as desired when using `cube.data.dtype` in the comparison. Would be keen to gather any suggestions!

`is` will fail if the instances are not the same object.

The second aspect is `numpy` using internal types in place of Python's built in primitives:

```python
>>> type(cube.data.dtype)
numpy.dtypes.Float64DType
>>> type(1.0)
<class 'float'>
```
That affects comparison operations (more on that in another comment).

blimlim:
Ok, I've fixed my problem where I omitted the return value, and tried to clean up the structure. Rather than combining both of the following in a single function:
1. getting the default fill value for a cube data's type
2. Setting the cube's `missing_data` attribute to the fill value

I separated the first part into a `get_default_fill_value()` function, which is then called by the overall `fix_fill_value()` function that completes the other steps. I think it's ready for review, and any suggestions or ideas are welcome!

---
There are a couple of things to note:

The behaviour _on the surface_ is not exactly the same as prior to this PR. 

In this PR, we apply the type conversion to `fill_value` before it is supplied to either the `missing_value` or `_FillValue` attributes.
https://github.com/ACCESS-NRI/um2nc-standalone/blob/e0048d475e68a0ad66a9878827f6c537561274b7/umpost/um2netcdf.py#L354

Previously, the `missing_value` attribute was converted to the cube's data type, but the `_FillValue` was not. However `_FillValue` [appears to be converted externally](https://github.com/ACCESS-NRI/um2nc-standalone/issues/116#issuecomment-2387395658), and so I don't believe the additional conversion added in this PR has any impacts. I've added a comment to try and clarify this https://github.com/ACCESS-NRI/um2nc-standalone/blob/e3b98f38efa2010d1b287497f2c7fc1acee1a6d7/umpost/um2netcdf.py#L344-L351

Converting an ESM1.5 monthly output file results in identical `missing_value` and `_FillValue` types before and after this PR.

umpost/um2netcdf.py
```diff
@@ -319,6 +321,66 @@ def fix_latlon_coords(cube, grid_type, dlat, dlon):
...
+
+    Returns
+    -------
+    fill_value: numpy scalar with type matching cube.data
+    """
+    if cube.data.dtype.kind == 'f':
+        fill_value = DEFAULT_FILL_VAL_FLOAT
+
+    else:
+        fill_value = default_fillvals[
```
truth-quark, line 340: Minor thing to improve the orig code, the top level `from netCDF import ...` can be changed to 'import netCDF'. Then `default_fillvals` can be accessed via the module name to indicate what module it belongs to.

That follows the "best" example here:
https://docs.python-guide.org/writing/structure/#modules

umpost/um2netcdf.py
truth-quark: Quick idea, do you want to try:
* set an IDE breakpoint at the check type
* running a fill val test in your IDE, in debug mode
* seeing what values both sides of `==` evaluate to?

blimlim: Ok sweet, I've had a go at that and during the tests we get:
```
type(fill_value)
<class 'numpy.float32'>
cube.data.dtype
dtype('float32')
.
type(fill_value)
<class 'numpy.float64'>
cube.data.dtype
dtype('float64')
.
type(fill_value)
<class 'numpy.complex64'>
cube.data.dtype
dtype('complex64')
.
type(fill_value)
<class 'numpy.int32'>
cube.data.dtype
dtype('int32')
.
type(fill_value)
<class 'numpy.int64'>
cube.data.dtype
dtype('int64')
.
type(fill_value)
<class 'numpy.float64'>
cube.data.dtype
dtype('float64')
.
# this is the one that's meant to fail
type(fill_value)
<class 'numpy.float32'>
cube.data.dtype
dtype('int32')
```

I just realised, after splitting it into two functions `get_default_fill_value()` and `fix_fill_value()`, we're essentially ensuring that the fill value provided to `fix_fill_value()` uses a numpy datatype, apart from when a `custom_fill_value` argument is supplied.

Would it be better to replace the if statement with `if fill_value.dtype == cube.data.dtype:`  and require that any `custom_fill_value` arguments have a numpy datatype?

umpost/um2netcdf.py
```diff
@@ -319,6 +321,66 @@ def fix_latlon_coords(cube, grid_type, dlat, dlon):
...
+        ]
+
+    # NB: the `_FillValue` attribute appears to be converted to match the
+    # cube data's type externally (likely in the netCDF4 library). It's not
+    # strictly necessary to do the conversion here. However, we need
+    # the separate `missing_value` attribute to have the correct type
+    # and so it's cleaner to set the type for both here.
+
+    # TODO: Is there a cleaner way do do the following conversion?
+    return np.array([fill_value], dtype=cube.data.dtype)[0]
```
truth-quark, line 351: A cleaner way might be possible with numpy's type "constructors":

```python
>>> import numpy as np
>>> x = np.int32(5)  # x is a numpy int
>>> type(x)
numpy.int32
```

This would need a lookup mapping a dtype key to a constructor function, like:
```python
lk = {"int32": np.int32, ...}
# other code

return lk[str(fill_value.dtype)](fill_value)
```
It's a bit more code, but doesn't involve creating a `np` array to convert a single primitive value.

umpost/um2netcdf.py
truth-quark: To clarify something, was there a requirement elsewhere to add a custom fill value?

blimlim: No external requirement, though it was included as an addition in the original issue #99 

If it's preferable we could remove the custom fill value completely?

truth-quark: > No external requirement, though it was included as an addition in the original issue #99
> 
> If it's preferable we could remove the custom fill value completely?

Whoops, I think I threw the custom value thing into the issue without considering the details.

Let's simplify by raising a NotImplementedError if a custom value is provided & comment out the type checks in a comment. Working in the function space has been useful to uncover complexity!

umpost/um2netcdf.py
blimlim: Good idea, I've added this in 656d699

umpost/um2netcdf.py
blimlim: That sound's good. I've swapped it to raise a `NotImplementedError` when the `custom_fill_value` argument is used, and added in a comment noting that type comparisons of some sort should be added if custom fill values are eventually implemented.

umpost/um2netcdf.py
blimlim: We've opted to leave the code as is, as other approaches require us to know all the data types that could appear in a cube. I've added a comment in 656d699 noting that any cleaner approaches should be implemented if they're found, and noted that a similar approach is currently used in the [`netCDF4` package](https://github.com/Unidata/netcdf4-python/blob/188cc6c03a8eef0add64c544667b14a0a7b3c85d/src/netCDF4/_netCDF4.pyx#L4413)

test/test_um2netcdf.py
```diff
@@ -1207,3 +1209,36 @@ def test_convert_32_bit_with_float64(ua_plev_cube):
...
+        (np.array([1, 2], dtype="int64"),
+         np.int64(default_fillvals["i8"]))
+    ]
+)
+def test_fix_fill_value_defaults(cube_data, expected_fill_val):
+    """
+    Check that correct default fill values are found based
+    on a cube's data's type.
+    """
+    fake_cube = DummyCube(12345, "fake_var", attributes={})
```
truth-quark, line 1234: It might possible to drop `attributes={}` when this is de-conflicted with the newer code in main.

test/test_um2netcdf.py
```diff
@@ -1207,3 +1209,36 @@ def test_convert_32_bit_with_float64(ua_plev_cube):
...
+    fake_cube.data = cube_data
+
+    fill_value = um2nc.fix_fill_value(fake_cube)
+
+    assert fill_value == expected_fill_val
+    # Check new fill value type matches cube's data's type
+    assert fill_value.dtype == cube_data.dtype
```
truth-quark, line 1241: Can the type check be skipped now?

blimlim:
Oh no... I broke the branch. I've set up a clean one in a new PR #147

test/test_um2netcdf.py
blimlim: Will keep these ones in, as it would be good to check that the `get_default_fill_value()` function correctly converts the default fill values to match the cube types.

test/test_um2netcdf.py
blimlim: Good catch. Fixed in the new PR #147

Issue #37, opened 2024-06-26
Add unit tests for convert_proleptic()


truth-quark:
Tasks:
* [ ] Add unit tests
* [ ] Fix some non-Python styling
* [ ] Extract modifier functionality to new function

Issue #8, opened 2024-06-17
Feature Requirement: calendar fixes


truth-quark:
`TODO`: what is the requirement for calendar fixes?
`TODO`: add unit tests

Are these the relevant code sections?
* https://github.com/ACCESS-NRI/um2nc-standalone/blob/main/src/um2netcdf.py#L24-L46
* https://github.com/ACCESS-NRI/um2nc-standalone/blob/main/src/um2netcdf.py#L138-L160
* https://github.com/ACCESS-NRI/um2nc-standalone/blob/main/src/um2netcdf.py#L169-L185

MartinDix:
Control runs often use nominal dates (or "model years"), starting from year 1. ESM1.5 uses a proleptic Gregorian calendar. The UM and iris only distinguish between 360 day and standard, so the calendar attribute isn't set correctly. This can causes date offsets with some post-processing software that's strict about the proleptic Gregorian vs standard=Julian/Gregorian distinction.

```
import iris
pr = iris.load_cube('/g/data/tm70/esm15_testing/sample_output/aiihca.paa1jan', 'rainfall_flux')
print(pr.coord('time'))

DimCoord :  time / (hours since 1970-01-01 00:00:00, standard calendar)
    points: [0101-01-16 12:00:00]
    bounds: [[0101-01-01 00:00:00, 0101-02-01 00:00:00]]
    shape: (1,)  bounds(1, 2)
    dtype: float64
    standard_name: 'time'
```

aidanheerdegen:
Note that MOM5 also has calendar issues, in that it claims to be gregorian (but also says it doesn't support gregorian in the code comments), but the calendar logic is for proleptic gregorian

https://github.com/COSIMA/access-om2/issues/117

We could/should fix MOM5.

aidanheerdegen:
This comment:

https://github.com/ACCESS-NRI/um2nc-standalone/pull/118#discussion_r1794360157

Suggests there are only 3 supported calendar types. If that is the case the code should explicitly NOT support other types, and return an error to users.

PR #118, opened 2024-10-01
Extract forecast reference time

### LINKED ISSUE #100: Extract `cubewrite()` forecast reference time functionality
For splitting `cubewrite()` (#12), forecast reference time functionality should be extracted to a separate function:

https://github.com/ACCESS-NRI/um2nc-standalone/blob/92096ce86f3cdd305570e940397b06633ae8777b/umpost/um2netcdf.py#L178-L202

Tasks:
* [ ] Examine test data for idea of real values
* [ ] Extract code block
* [ ] Add unit tests using 
* [ ] Move `except` block to the coord lookups
* [ ] Convert any strings to constants?
  * [ ] `forecast_reference_time`
  * [ ] `proleptic_gregorian`

For now, let's keep `cubewrite()`'s steps the same, merging into `process()` can be a separate refactoring issue.

truth-quark:
First step to doing #100.

This PR is intended for facilitating testing functionality for fixing the forecast reference time.

As of late Sept 2024, work up to commit fd664c7b only provides a skeleton of unit tests. It doesn't meaningfully test any cube time changes. This PR includes some skipped tests as the testing required is unclear.

Work on this PR needs to answer & address the following:
* [ ] Is the `time.units.calendar == 'gregorian'` code branch & testing required?
* [ ] Is the `time.units.calendar == 'proleptic_gregorian'` code branch & testing required?
* [ ] What inputs should be tested?
* [ ] What outputs should be tested?
* [ ] What cube state/attrs should be tested?
* [ ] Is the code branch covered by `test_fix_forecast_reference_time_standard()` required?

These questions will help determine what work is required. Any comments/ideas welcome!

Fixing & testing this function is specified in #141.

### DIFF
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -82,6 +82,7 @@ class DummyStash:
     item: int
 
 
+# TODO: would making this a dataclass provide any benefit?
 class DummyCube:
     """
     Imitation iris Cube for unit testing.
@@ -119,13 +120,19 @@ class DummyCube:
             msg = f"{self.__class__}[{self.var_name}]: lacks coord for '{_name}'"
             raise CoordinateNotFoundError(msg)
 
-    def update_coords(self, coords):
+    def remove_coord(self, coord):
+        del self._coordinates[coord]
+
+    # Auxiliary methods
+    # These methods DO NOT exist in the iris cube API, these are helper methods
+    # to configure DummyCubes for testing.
+    #
+    # All methods should see an `aux_` prefix
+
+    def aux_update_coords(self, coords):
         # Mimic a coordinate dictionary keys for iris coordinate names. This
         # ensures the access key for coord() matches the coordinate's name
         self._coordinates = {c.name(): c for c in coords} if coords else {}
-
-    def remove_coord(self, coord):
-        del self._coordinates[coord]
 
 
 # NB: these cube fixtures have been chosen to mimic cubes for testing key parts
@@ -1178,7 +1185,7 @@ def test_fix_forecast_reference_time_exit_on_missing_ref_time(forecast_cube):
 def test_fix_forecast_reference_time_exit_on_missing_time(forecast_cube,
                                                           forecast_ref_time_coord):
     # verify fix_forecast_ref_time() exits early if the coord is missing
-    forecast_cube.update_coords([forecast_ref_time_coord])
+    forecast_cube.aux_update_coords([forecast_ref_time_coord])
 
     with pytest.raises(iris.exceptions.CoordinateNotFoundError):
         forecast_cube.coord(um2nc.TIME)
@@ -1196,9 +1203,9 @@ def test_fix_forecast_reference_time_standard(forecast_cube,
     forecast_period = iris.coords.DimCoord([372.0],
                                            standard_name=um2nc.FORECAST_PERIOD)
 
-    forecast_cube.update_coords([forecast_ref_time_coord,
-                                 time_coord,
-                                 forecast_period])
+    forecast_cube.aux_update_coords([forecast_ref_time_coord,
+                                     time_coord,
+                                     forecast_period])
 
     assert um2nc.fix_forecast_reference_time(forecast_cube) is None
 
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -1182,3 +1182,34 @@ def test_convert_32_bit_with_float64(ua_plev_cube):
     ua_plev_cube.data = array
     um2nc.convert_32_bit(ua_plev_cube)
     assert ua_plev_cube.data.dtype == np.float32
+
+
+# fix forecast reference time tests
+@pytest.fixture
+def forecast_cube():
+    return DummyCube(item_code=999)
+
+
+@pytest.fixture
+def forecast_ref_time_coord():
+    # data ripped from aiihca data file
+    ref_time = iris.coords.DimCoord([-16383336.],
+                                    standard_name=um2nc.FORECAST_REFERENCE_TIME)
+    return ref_time
+
+
+def test_fix_forecast_reference_time_exit_on_missing_ref_time(forecast_cube):
+    with pytest.raises(iris.exceptions.CoordinateNotFoundError):
+        forecast_cube.coord(um2nc.FORECAST_REFERENCE_TIME)
+
+    um2nc.fix_forecast_reference_time(forecast_cube)
+
+
+def test_fix_forecast_reference_time_exit_on_missing_time(forecast_cube,
+                                                          forecast_ref_time_coord):
+    forecast_cube.update_coords([forecast_ref_time_coord])
+
+    with pytest.raises(iris.exceptions.CoordinateNotFoundError):
+        forecast_cube.coord(um2nc.TIME)
+
+    um2nc.fix_forecast_reference_time(forecast_cube)
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -406,10 +406,10 @@ class DummyCube:
         self.standard_name = None
         self.long_name = None
         self.data = None
+        self._coordinates = {}
 
-        # Mimic a coordinate dictionary keys for iris coordinate names. This
-        # ensures the access key for coord() matches the coordinate's name
-        self._coordinates = {c.name(): c for c in coords} if coords else {}
+        if coords:
+            self.update_coords(coords)
 
     def name(self):
         return self.var_name
@@ -420,6 +420,11 @@ class DummyCube:
         except KeyError:
             msg = f"{self.__class__}: lacks coord for '{name}'"
             raise CoordinateNotFoundError(msg)
+
+    def update_coords(self, coords):
+        # Mimic a coordinate dictionary keys for iris coordinate names. This
+        # ensures the access key for coord() matches the coordinate's name
+        self._coordinates = {c.name(): c for c in coords} if coords else {}
 
 
 def test_set_item_codes_avoid_overwrite():
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -1192,28 +1192,38 @@ def test_convert_32_bit_with_float64(ua_plev_cube):
 # fix forecast reference time tests
 @pytest.fixture
 def forecast_cube():
+    # NB: using a non-existent item code for fake forecast cube
     return DummyCube(item_code=999)
 
 
 @pytest.fixture
-def forecast_ref_time_coord():
-    # units data ripped from aiihca data file
+def time_points():
+    """Use for cube.coord('time').points attribute."""
+    return [-16382964.]
+
+
+@pytest.fixture
+def forecast_ref_time_coord(time_points):
+    # units & point data ripped from aiihca.paa1jan data file:
+    # cubes = iris.load("aiihca.paa1jan")
+    # cubes[0].long_name --> 'atmosphere_optical_thickness_due_to_sulphate_ambient_aerosol'
+    # cubes[0].coord("time").points --> array([-16382964.])
     unit = cf_units.Unit(unit="hours since 1970-01-01 00:00:00")
     assert unit.calendar == um2nc.STANDARD
 
-    return iris.coords.DimCoord([-16383336.],
+    return iris.coords.DimCoord(time_points,
                                 standard_name=um2nc.FORECAST_REFERENCE_TIME,
                                 units=unit)
 
 
 @pytest.fixture
-def time_coord():
+def time_coord(time_points):
     # units data ripped from aiihca data file
     unit = cf_units.Unit(unit="hours since 1970-01-01 00:00:00",
                          calendar=um2nc.GREGORIAN)
     assert unit.calendar == um2nc.STANDARD
 
-    return iris.coords.DimCoord([-16382964.],
+    return iris.coords.DimCoord(time_points,
                                 standard_name=um2nc.TIME,
                                 units=unit)
 
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -1209,7 +1209,7 @@ def forecast_ref_time_coord(time_points):
     # cubes[0].long_name --> 'atmosphere_optical_thickness_due_to_sulphate_ambient_aerosol'
     # cubes[0].coord("time").points --> array([-16382964.])
     unit = cf_units.Unit(unit="hours since 1970-01-01 00:00:00")
-    assert unit.calendar == um2nc.STANDARD
+    assert unit.calendar == cf_units.CALENDAR_STANDARD
 
     return iris.coords.DimCoord(time_points,
                                 standard_name=um2nc.FORECAST_REFERENCE_TIME,
@@ -1220,8 +1220,8 @@ def forecast_ref_time_coord(time_points):
 def time_coord(time_points):
     # units data ripped from aiihca data file
     unit = cf_units.Unit(unit="hours since 1970-01-01 00:00:00",
-                         calendar=um2nc.GREGORIAN)
-    assert unit.calendar == um2nc.STANDARD
+                         calendar=cf_units.CALENDAR_GREGORIAN)
+    assert unit.calendar == cf_units.CALENDAR_STANDARD
 
     return iris.coords.DimCoord(time_points,
                                 standard_name=um2nc.TIME,
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -1134,7 +1134,13 @@ def time_points():
 
 
 @pytest.fixture
-def forecast_ref_time_coord(time_points):
+def forecast_ref_time_points():
+    """Use for cube.coord('forecast_reference_time').points attribute."""
+    return [-16383336.]
+
+
+@pytest.fixture
+def forecast_ref_time_coord(forecast_ref_time_points):
     # units & point data ripped from aiihca.paa1jan data file:
     # cubes = iris.load("aiihca.paa1jan")
     # cubes[0].long_name --> 'atmosphere_optical_thickness_due_to_sulphate_ambient_aerosol'
@@ -1142,7 +1148,7 @@ def forecast_ref_time_coord(time_points):
     unit = cf_units.Unit(unit="hours since 1970-01-01 00:00:00")
     assert unit.calendar == cf_units.CALENDAR_STANDARD
 
-    return iris.coords.DimCoord(time_points,
+    return iris.coords.DimCoord(forecast_ref_time_points,
                                 standard_name=um2nc.FORECAST_REFERENCE_TIME,
                                 units=unit)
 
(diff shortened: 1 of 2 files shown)


truth-quark:
Adding Martin as a reviewer to help clarify what is required.

umpost/um2netcdf.py
```diff
@@ -406,6 +360,64 @@ def cubewrite(cube, sman, compression, use64bit, verbose):
...
+def fix_forecast_reference_time(cube):
+    # If reference date is before 1600 use proleptic gregorian
+    # calendar and change units from hours to days
+    try:
+        reftime = cube.coord('forecast_reference_time')
+        time = cube.coord('time')
+        refdate = reftime.units.num2date(reftime.points[0])
+        assert time.units.origin == 'hours since 1970-01-01 00:00:00'
+
+        if time.units.calendar == 'proleptic_gregorian' and refdate.year < 1600:
```
truth-quark, line 372: `1600` should be a configurable parameter?

test/test_um2netcdf.py
```diff
@@ -1177,3 +1187,79 @@ def test_convert_32_bit_with_float64(ua_plev_cube):
...
+    return DummyCube(item_code=999)
+
+
+@pytest.fixture
+def forecast_ref_time_coord():
+    # units data ripped from aiihca data file
+    unit = cf_units.Unit(unit="hours since 1970-01-01 00:00:00")
+    assert unit.calendar == um2nc.STANDARD
+
+    return iris.coords.DimCoord([-16383336.],
```
CodeGat, line 1204: Whats the significance of this number..?

truth-quark: It looks like it's ~100 years before 1 Jan 1970:
```python
>>> -16382964.0 / 24 / 365  # convert to days & years
-1870.2013698630137
```
I've moved the value into a test fixture where it can be documented.

CodeGat (approved):
This looks good to me. But note that I'm missing some of the domain knowledge needed to give a more 'conceptual' approval, so probably wait for others to give their review :D

blimlim:
Pulling this section out into a separate function sounds like a good idea!

I'm still trying to understand the original version of the code, but have a couple of comments and questions so far:

It's a bit tricky to follow I think because of the dynamic overwrite of the `PPfile.calendar` method: https://github.com/ACCESS-NRI/um2nc-standalone/blob/d7117296314ce3ab34f0423767dd9e798be92875/umpost/um2netcdf.py#L95-L110

When I load in an ESM1.5 output file independently, e.g. `/g/data/tm70/sw6175/nc_conversion/output_dirs_for_testing/output000/atmosphere/aiihca.paa1jan`, I get `cube.coord("time").units.calendar` is `standard`. However, when I run `um2nc` on it, the ` if time.units.calendar == 'proleptic_gregorian' and refdate.year < 1600:` branch is still executed because of the above overwrite.


>  Is the time.units.calendar == 'gregorian' code branch & testing required?

Because of the `PPfile.calendar` overwrite, I suspect we will never have a cube with a `gregorian` or `standard` calendar, and so I'm wondering if we can remove this branch entirely, or at least skip tests for it?



>  Is the time.units.calendar == 'proleptic_gregorian' code branch & testing required?

I think it would be good to test this branch eventually. Paleoclimate simulations usually use years before `1600`, and so this branch would be used quite regularly.

---

@MartinDix, do you know why the `forecast_reference_time` coordinate is used to determine if the date is before 1600, rather than using the `time` coordinate directly? https://github.com/ACCESS-NRI/um2nc-standalone/blob/d7117296314ce3ab34f0423767dd9e798be92875/umpost/um2netcdf.py#L367-L369

I'm wondering if it's to ensure that the calendar of dump files isn't changed, e.g. https://github.com/ACCESS-NRI/um2nc-standalone/blob/d7117296314ce3ab34f0423767dd9e798be92875/umpost/um2netcdf.py#L388-L390

but thought it would be good to double check!

umpost/um2netcdf.py
```diff
@@ -70,6 +70,14 @@
...
 SIGMA = "sigma"
 
+# forecast reference time constants
+FORECAST_REFERENCE_TIME = "forecast_reference_time"
+TIME = "time"
+PROLEPTIC_GREGORIAN = "proleptic_gregorian"
+GREGORIAN = "gregorian"
+FORECAST_PERIOD = "forecast_period"
+STANDARD = "standard"
+
```
blimlim, line 81: We might be able to use the `cf_units` calendar constants that are used by `def pg_calendar` if it's preferable. E.g. `cf_units.CALENDAR_PROLEPTIC_GREGORIAN`, `cf_units.CALENDAR_GREGORIAN`, and `cf_units.CALENDAR_STANDARD`.

umpost/um2netcdf.py
```diff
@@ -406,6 +360,64 @@ def cubewrite(cube, sman, compression, use64bit, verbose):
...
+        time = cube.coord('time')
+        refdate = reftime.units.num2date(reftime.points[0])
+        assert time.units.origin == 'hours since 1970-01-01 00:00:00'
+
+        if time.units.calendar == 'proleptic_gregorian' and refdate.year < 1600:
+            convert_proleptic(time)
+        else:
+            if time.units.calendar == 'gregorian':
+                new_calendar = 'proleptic_gregorian'
+            else:
```
blimlim, line 377: Just noticed this isn't using the added constants (e.g. `PROLEPTIC_GREGORIAN`, or otherwise the `cf_units` alternatives)

truth-quark:
I'll link in some previous work on `convert_propleptic()` in issue #37.
There's [part of a unit test here](https://github.com/ACCESS-NRI/um2nc-standalone/commit/e50a33a6b3f19050643d02fa5d305ee13d160d8b) which might be useful...

umpost/um2netcdf.py
truth-quark: Indeed, using the `cf_units` constants is preferable to duplicating them. The fixes are in with 7ee59a23.

umpost/um2netcdf.py
truth-quark: This is updated in 16a3768a.

truth-quark:
Update: following the 9/10/2024 handover meeting, work on this will be split into 2 parts:
1. Extract the time.calendar functions & provide **basic** unit test scaffolding
2. Retrofit testing in another issue & second PR.

umpost/um2netcdf.py
truth-quark: Defer this to #141.

umpost/um2netcdf.py
```diff
@@ -414,6 +365,65 @@ def cubewrite(cube, sman, compression, use64bit, verbose):
...
+        else:
+            if time.units.calendar == cf_units.CALENDAR_GREGORIAN:
+                new_calendar = cf_units.CALENDAR_PROLEPTIC_GREGORIAN
+            else:
+                new_calendar = time.units.calendar
+
+            time.units = cf_units.Unit("days since 1970-01-01 00:00", calendar=new_calendar)
+            time.points = time.points / 24.
+
+            if time.bounds is not None:
```
aidanheerdegen, line 388: I wonder if it wouldn't be better to just check if `cube` has a FORECAST_REFERENCE_TIME coord outside this routine and only call it if so?

Then you could dump the `try`/`except` block.

As it is removing `FORECAST_PERIOD` too then I guess that either needs to be checked for or put in a smaller `try`/`except`

truth-quark:
> First pass, didn't have time to check the tests. Can take another look tomorrow.

This PR has given us several open questions & highlighted that the requirements are not really known. This function partly blocks work to fix structural/architectural repairs too.  To address all these, how is the following?

* Add code comments for all open questions
* Merge the comments & structural only changes via this PR
* Update #100 with details from this PR & leave it open
* Clarify requirements, what UM provides, what ACCESS NRI needs --> add to issue docs
* Use new PR for modifying the timing repair based on #100

How does that sound?

umpost/um2netcdf.py
```diff
@@ -414,6 +365,65 @@ def cubewrite(cube, sman, compression, use64bit, verbose):
...
+    newunits = cf_units.Unit("days since 0001-01-01 00:00",
+                             calendar=cf_units.CALENDAR_PROLEPTIC_GREGORIAN)
+    tvals = np.array(time.points)  # Need a copy because can't assign to time.points[i]
+    tbnds = np.array(time.bounds) if time.bounds is not None else None
+
+    for i in range(len(time.points)):
+        date = time.units.num2date(tvals[i])
+        newdate = cftime.DatetimeProlepticGregorian(date.year, date.month, date.day,
+                                                    date.hour, date.minute, date.second)
+        tvals[i] = newunits.date2num(newdate)
```
aidanheerdegen, line 410: I don't think it is necessary to loop through the whole time array. I think it should be ok to just do the `num2date > newdate > date2num` operation on just the first element, work out the difference between this and the original value and apply that offset to the rest of the `tvals` and `tbnds` in two (vectorised) operations.

umpost/um2netcdf.py
```diff
@@ -414,6 +365,65 @@ def cubewrite(cube, sman, compression, use64bit, verbose):
...
 
+def fix_forecast_reference_time(cube):
+    # If reference date is before 1600 use proleptic gregorian
+    # calendar and change units from hours to days
+    try:
+        reftime = cube.coord(FORECAST_REFERENCE_TIME)
+        time = cube.coord(TIME)
+        refdate = reftime.units.num2date(reftime.points[0])
+        assert time.units.origin == 'hours since 1970-01-01 00:00:00'
+
```
aidanheerdegen, line 376: There is a logic error here. It doesn't correctly deal with `cf_units.CALENDAR_GREGORIAN and refdate.year < 1600`. In this case it will silently convert to proleptic gregorian, which is not correct behaviour if the time is actually gregorian.

umpost/um2netcdf.py
```diff
@@ -414,6 +365,65 @@ def cubewrite(cube, sman, compression, use64bit, verbose):
...
+            else:
+                new_calendar = time.units.calendar
+
+            time.units = cf_units.Unit("days since 1970-01-01 00:00", calendar=new_calendar)
+            time.points = time.points / 24.
+
+            if time.bounds is not None:
+                time.bounds = time.bounds / 24.
+
+        cube.remove_coord(FORECAST_PERIOD)
```
aidanheerdegen, line 391: If the calendar is passed as a function argument all the conversion from hours to days could be done in this function, simplifying the logic.

blimlim (changes requested):
I think the modularisation step looks good, and looks like it will behave exactly the same as before.

Setting up the tests and fixtures for this seems very tricky and intricate though, especially due to the `pg_calendar` overwrite. I'm wondering whether it would be easiest/fastest for this PR to just do the extraction into a separate function, and defer all of the tests to #141?

umpost/um2netcdf.py
blimlim: This one's also affected by the overwrite of the `PPField.calendar()` method, which I think means the only calendars the rest of the code ever encounters is `cf_units.CALENDAR_PROLEPTIC_GREGORIAN`, `cf_units.CALENDAR_365_DAY` or `cf_units.CALENDAR_360_DAY`. 

https://github.com/ACCESS-NRI/um2nc-standalone/blob/2eaa205a48a5b2d1fba9f994dec0099a48d9fe97/umpost/um2netcdf.py#L100-L115

Possibly these are the only three calendars that the UM ever works with – @MartinDix is that correct?

test/test_um2netcdf.py
```diff
@@ -1107,3 +1117,89 @@ def test_convert_32_bit_with_float64(ua_plev_cube):
...
+    return iris.coords.DimCoord(time_points,
+                                standard_name=um2nc.TIME,
+                                units=unit)
+
+
+def test_fix_forecast_reference_time_exit_on_missing_ref_time(forecast_cube):
+    with pytest.raises(iris.exceptions.CoordinateNotFoundError):
+        forecast_cube.coord(um2nc.FORECAST_REFERENCE_TIME)
+
+    assert um2nc.fix_forecast_reference_time(forecast_cube) is None
```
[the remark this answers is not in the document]
truth-quark: This is a silly test executing the exception handler block. It's missing an assertion to ensure the cube isn't modified, but it might be simpler to delete it. I can't recall why I added it...

test/test_um2netcdf.py
```diff
@@ -1107,3 +1117,89 @@ def test_convert_32_bit_with_float64(ua_plev_cube):
...
+
+
+def test_fix_forecast_reference_time_exit_on_missing_time(forecast_cube,
+                                                          forecast_ref_time_coord):
```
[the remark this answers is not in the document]
truth-quark: This was testing failure of the 2nd coord lookup `time = cube.coord(TIME)`.

The tests in this PR are mostly foundations that still need to be filled out. I'll add comments to flag they need to be expanded.

umpost/um2netcdf.py
truth-quark: > As it is removing `FORECAST_PERIOD` too then I guess that either needs to be checked for or put in a smaller `try`/`except`

The `iris` function `remove_coord()` does a `coord()` lookup, which raises an exception if the forecast period is missing:
https://github.com/SciTools/iris/blob/cc60a60508b741709ace45b8ea4fc05e5cd5e8a6/lib/iris/cube.py#L1813-L1830

I think we should consider swapping the 2 `remove_coord()` calls. Remove the forecast reference time first, as it's from to exist from the `try` test. then attempt to remove the forecast period. The current code could fail on removing the forecast period, leaving behind the forecast reference time, HOWEVER this might be the desired behaviour.

test/test_um2netcdf.py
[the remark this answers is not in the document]
blimlim: Ah I see. Otherwise I think leaving it in with some extra comments could be ok!

test/test_um2netcdf.py
```diff
@@ -116,6 +118,14 @@ def coord(self, _name):
             msg = f"{self.__class__}[{self.var_name}]: lacks coord for '{_name}'"
             raise CoordinateNotFoundError(msg)
 
+    def update_coords(self, coords):
+        # Mimic a coordinate dictionary keys for iris coordinate names. This
```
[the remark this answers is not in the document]
truth-quark: This is where the `Mock` class was useful, it could be passed an object & dynamically make an interface from that.

I can prefix helper funcs with `aux_` & add comments to clarify non-iris interface elements. Would that help?

test/test_um2netcdf.py
```diff
@@ -1107,3 +1117,89 @@ def test_convert_32_bit_with_float64(ua_plev_cube):
...
+
+@pytest.fixture
+def time_points():
+    """Use for cube.coord('time').points attribute."""
+    return [-16382964.]
+
+
+@pytest.fixture
+def forecast_ref_time_coord(time_points):
+    # units & point data ripped from aiihca.paa1jan data file:
```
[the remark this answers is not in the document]
truth-quark: Yep, that correctness fix is in 0925960.

test/test_um2netcdf.py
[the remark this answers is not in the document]
truth-quark: I've added comments in 6e8b5e2f.

test/test_um2netcdf.py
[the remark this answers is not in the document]
truth-quark: Comments have gone in with 6e8b5e2f.

umpost/um2netcdf.py
truth-quark: The looping code is old style, conflating logic with manual index fiddling etc. I did some work on #37 to use iteration, but didn't fully solve the problem:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/e50a33a6b3f19050643d02fa5d305ee13d160d8b/umpost/um2netcdf.py#L67-L82

test/test_um2netcdf.py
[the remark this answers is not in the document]
blimlim: I think that would be helpful!

umpost/um2netcdf.py
```diff
@@ -414,6 +365,65 @@ def cubewrite(cube, sman, compression, use64bit, verbose):
         sman.write(cube, zlib=True, complevel=compression, fill_value=fill_value)
 
 
+def fix_forecast_reference_time(cube):
+    # If reference date is before 1600 use proleptic gregorian
```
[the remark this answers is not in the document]
truth-quark: I've flagged this as a TODO in 43e3ba17.

test/test_um2netcdf.py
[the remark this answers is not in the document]
truth-quark: The fix is in d3164e3d.

truth-quark:
@blimlim ^ @aidanheerdegen do you want to check over the recent sets of changes & confirm if they cover the discussion threads here? I'll define "cover" as implements smaller fixes OR makes a note for a future fix.

blimlim:
> confirm if they cover the discussion threads here?

Changes look good

Would you be happy to add comments to the `forecast_ref_time_coord` and `time_coord` fixtures to note that their `unit.calendar` attributes will likely have to be changed in the future, since `um2nc` only ever uses the `cf_units.CALENDAR_PROLEPTIC_GREGORIAN`, `cf_units.CALENDAR_360_DAY`, and `cf_units.CALENDAR_365_DAY` calendars?

umpost/um2netcdf.py
aidanheerdegen: I've added a link to the above comment in [this issue](https://github.com/ACCESS-NRI/um2nc-standalone/issues/8#issuecomment-2404062840). This should be dealt with separately taking into account other calendar related issues.

truth-quark:
> Would you be happy to add comments to the `forecast_ref_time_coord` and `time_coord` fixtures to note that their `unit.calendar` attributes will likely have to be changed in the future <snip>

That's gone in with 38205281.

aidanheerdegen (approved):
LFTM

test/test_um2netcdf.py
```diff
@@ -1107,3 +1117,89 @@ def test_convert_32_bit_with_float64(ua_plev_cube):
...
+    # cubes = iris.load("aiihca.paa1jan")
+    # cubes[0].long_name --> 'atmosphere_optical_thickness_due_to_sulphate_ambient_aerosol'
+    # cubes[0].coord("time").points --> array([-16382964.])
+    unit = cf_units.Unit(unit="hours since 1970-01-01 00:00:00")
+    assert unit.calendar == cf_units.CALENDAR_STANDARD
```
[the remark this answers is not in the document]
blimlim: Fixme added in [3820528](https://github.com/ACCESS-NRI/um2nc-standalone/pull/118/commits/38205281c3f167b99360fa76bcc6a80905fc8c2d)

blimlim (approved):
Looks good to me

truth-quark:
> Setting up the tests and fixtures for this seems very tricky and intricate though, especially due to the `pg_calendar` overwrite. I'm wondering whether it would be easiest/fastest for this PR to just do the extraction into a separate function, and defer all of the tests to #141?

Yep, initial attempts at testing were fiddly & nasty to set up the data structures. Approaching in small, logical steps, extracting helper funcs etc should make things better with gradual logic reduction.

PR #127, opened 2024-10-04
Prevent mule missing STASHmaster warnings

### LINKED ISSUE #126: Irrelevant `STASHmaster` warning from mule
`process` produces the following warning each time it is run on gadi:
```
/g/data/vk83/apps/payu/1.1.5/lib/python3.10/site-packages/mule/stashmaster.py:259: UserWarning: 
Unable to load STASHmaster from version string, path does not exist
Path: $UMDIR/vn7.3/ctldata/STASHmaster/STASHmaster_A
Please check that the value of mule.stashmaster.STASHMASTER_PATH_PATTERN is correct for your site/configuration
  warnings.warn(msg)
```

Which is raised each time this line is run: https://github.com/ACCESS-NRI/um2nc-standalone/blob/33d0dc5ee9830c46f86bbd478fd75698832d34a8/umpost/um2netcdf.py#L468

The warning fills up the stderr when doing several conversions, and can appear alarming/confusing to the user. @MartinDix mentioned that `um2nc`'s use of `mule` doesn't use any `STASHmaster` information (using `stashvar.py` instead to get the variables' information), meaning that this warning is harmless.

Would it be worth filtering out this warning in order to clean up the output?

blimlim:
Closes #126 

This PR adds a warning filter around the `mule.load_umfile()` call in `process`, in order to hide warnings about missing STASHmaster files. The information in these files is not used in `um2nc` and hence these warnings aren't too relevant.

I aded the filter directly in `process` to try and keep it localised, however this does add slightly more code to `process`. Any suggested changes or ideas are welcome!

### DIFF
--- a/umpost/um2netcdf.py
+++ b/umpost/um2netcdf.py
@@ -465,7 +465,14 @@ def apply_mask(c, heaviside, hcrit):
 
 
 def process(infile, outfile, args):
-    ff = mule.load_umfile(str(infile))
+
+    with warnings.catch_warnings():
+        # NB: Information from STASHmaster file is not required by `process`.
+        # Hence supress missing STASHmaster warnings.
+        warnings.filterwarnings(action="ignore", category=UserWarning,
+                                message=r"\sUnable to load STASHmaster")
+        ff = mule.load_umfile(str(infile))
+
     mv = process_mule_vars(ff)
 
     cubes = iris.load(infile)


truth-quark:
> I aded the filter directly in `process` to try and keep it localised, however this does add slightly more code to `process`. Any suggested changes or ideas are welcome!

Localising the context manager to 1 function call is ideal & keeps the scope small. Nice one! Command line test output is the cleanest it has ever been for this project.

This one is good to go!

PR #132, opened 2024-10-08
Capture 32 bit overflow warnings in tests.

### LINKED ISSUE #128: Supress `test_convert_32_bit()` warnings
In `tests/test_um2netcdf.py`, warnings are emitted by `test_convert_32_bit()`.

Capturing the warnings helps clean the test output, especially on calling the test runner from the command line.

truth-quark:
Closes #128.

This PR covers a few things:
* Decouples the simple int conversion case from the parameter generated overflow testing
* Moves the basic condition to its own test
* Captures overflow warnings to keep test output clean

Any comments welcome!

### DIFF
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -1071,6 +1071,7 @@ def test_fix_pressure_levels_reverse_pressure():
 # NB: skip float64 to float32 overflow as float32 min/max is huge: -/+ 3.40e+38
 
 def test_convert_32_bit_safe(ua_plev_cube):
+    # simple baseline test of down conversion int64 to int32 without warnings
     data = [1e6, 200, 100, 10, 1, 0, -10]
     ua_plev_cube.data = np.array(data, dtype=np.int64)
     um2nc.convert_32_bit(ua_plev_cube)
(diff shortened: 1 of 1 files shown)


test/test_um2netcdf.py
```diff
@@ -1069,13 +1069,29 @@ def test_fix_pressure_levels_reverse_pressure():
...
+
+    with pytest.warns(RuntimeWarning) as record:
+        um2nc.convert_32_bit(ua_plev_cube)
+
+        if not record:
+            msg = f"No overflow warning with {array} {_operator} {bound}"
+            pytest.fail(msg)
 
     if _operator:
         assert _operator(array[0], bound)
```
blimlim, line 1097: I think the separation of the test into the `safe` and `overflow` cases helps make it easier to read! Now that there's no `_operator == None` case in this test, is it worth removing the `if _operator` check?

I'm also wondering, is this check included to make sure the test setup is correct? I'm wondering if would be clearest to have it near the start of the test (e.g. around line 1088)?

blimlim (approved):
Changes look good to me! I like how the warnings/failures are handled.

Just had a tiny style comment/question though it's not crucial.

test/test_um2netcdf.py
truth-quark: I've fixed both in 85ef8b6f & cleaned the error messages.

There's a safety net If test params are somehow munged in other commits.

truth-quark:
This PR needs a re-review as to unblock merging.

blimlim (approved):
Awesome, changes look good!

PR #133, opened 2024-10-08
Guard against ESM1.5 file naming collisions

### LINKED ISSUE #124: ESM1.5 driver: guard against file name collisions
The [updated netCDF naming scheme for ESM1.5 ](https://github.com/ACCESS-NRI/um2nc-standalone/pull/93) takes the first part of the fields file name, and adds on the year, month, and frequency as a suffix. E.g. `aiihca.paa1jan` -> `aiihca.pa-010101_mon.nc`.

In some specialised cases, researchers might configure the model to write output files at higher frequencies. Rather than the standard setup which splits output into monthly segments, it's possible for example to split the output into daily segments. There's a risk that the netCDF file naming will produce collisions in this situation, since the file names use year and month.

Currently, naming collisions would cause the same netCDF file to be silently rewritten. This is especially an issue given the automatic fields file deletion in the ESM1.5 configurations.

@MartinDix mentioned that we don't need to fully support these custom setups, but that we should still warn when filename collisions would occur, and prevent files from silently being overwritten. 

A couple of options are:
- Raise and exception from `process` if a file already exists at the supplied output path. This would be the simplest to implement. It might result in messy behaviour from the ESM1.5 driver: If fields files `file_a`, `file_b`, and `file_c` were all going to be rewritten to the same file `newfile.nc`, the driver would convert `file_a`, and then error out once it reached `file_b`. However, given that we're not aiming to explicitly support this situation, this may be sufficient. 
- All the output paths are calculated in the ESM1.5 driver before any conversions are done. We could check whether any collisions are set to occur, and either:
   - (a) Error out completely if any collisions are set to occur, or
   - (b) Convert the files that don't have any collisions, and ignore the ones that do. While the idea is simple, I think the way the conversion driver is [currently set up](https://github.com/ACCESS-NRI/um2nc-standalone/blob/e2a39a7fa9fa25b30d0c5a0414f17245798e3d5d/umpost/conversion_driver_esm1p5.py#L271-L325) might require some refactoring to implement this. I think this could be the most informative option though.
   - (c) Revert to the output name being `input_name + .nc` for files where collisions would occur, and otherwise convert as usual with deletion.

@MartinDix, I'm wondering which behaviour you think would be best for handling naming collisions?

blimlim:
This PR closes #124 

It adds a new step to the `convert_esm1p5_output_dir` function, where pairs of `(input_path, output_path)` tuples are removed prior to the conversion step if their `output_path`s are not unique, and warns the user when this occurs.

### DIFF
--- a/umpost/conversion_driver_esm1p5.py
+++ b/umpost/conversion_driver_esm1p5.py
@@ -282,7 +282,7 @@ def filter_name_collisions(input_output_pairs):
 
     Parameters
     ----------
-    input_ouptut_pairs: iterator of of tuples (input_path, output_path).
+    input_ouptut_pairs: iterator of tuples (input_path, output_path).
 
     Yields
     -------
--- a/umpost/conversion_driver_esm1p5.py
+++ b/umpost/conversion_driver_esm1p5.py
@@ -268,6 +268,45 @@ def format_failures(failed, quiet):
             yield failure_report
 
 
+def _resolve_path(path):
+    """
+    Resolve path for use in comparison. Ensure that symlinks, relative paths,
+    and home directories are expanded.
+    """
+    return os.path.realpath(os.path.expanduser(path))
+
+
+def filter_name_collisions(input_output_pairs):
+    """
+    Remove input/output pairs which have overlapping output paths.
+
+    Parameters
+    ----------
+    input_ouptut_pairs: iterator of of tuples (input_path, output_path).
+
+    Yields
+    -------
+    filtered_pairs: (input_path, output_path) tuples with unique
+        output_path values.
+    """
+    # Convert to list to allow repeated traversal.
+    input_output_pairs = list(input_output_pairs)
+
+    output_paths = [_resolve_path(output) for _, output in input_output_pairs]
+    output_counts = collections.Counter(output_paths)
+
+    for input_path, output_path in input_output_pairs:
+        if output_counts[_resolve_path(output_path)] != 1:
+            msg = (
+                f"Multiple inputs have same output path {output_path}.\n"
+                f"{input_path} will not be converted."
+            )
+            warnings.warn(msg)
+            continue
+
+        yield input_path, output_path
+
+
 def convert_esm1p5_output_dir(esm1p5_output_dir):
     """
     Driver function for converting ESM1.5 atmospheric outputs during a simulation.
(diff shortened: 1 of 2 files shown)


umpost/conversion_driver_esm1p5.py
```diff
@@ -268,6 +268,45 @@ def format_failures(failed, quiet):
             yield failure_report
 
 
+def _resolve_path(path):
+    """
+    Resolve path for use in comparison. Ensure that symlinks, relative paths,
+    and home directories are expanded.
+    """
+    return os.path.realpath(os.path.expanduser(path))
+
```
blimlim, line 277: I'm wondering if this is overkill compared to just using `pathlib`'s `Path.resolve()` method. My main concern is making sure different representations of the same path are properly handled. `Path.resolve()` doesn't treat e.g. `~/a/b` and `/<home-directory>/a/b` the same, hence the use of `os.path.expanduser()`.

This is unlikely to come up in practice however, as the output paths are created elsewhere in the conversion driver, and I don't think there would be any way for one output to be given `~/a/...` and another to be given `/<home-directory>/a/...`.

aidanheerdegen: I don't mind being thorough if it's not an expensive operation (not being done thousands of times).

umpost/conversion_driver_esm1p5.py
```diff
@@ -268,6 +268,45 @@ def format_failures(failed, quiet):
...
+
+def filter_name_collisions(input_output_pairs):
+    """
+    Remove input/output pairs which have overlapping output paths.
+
+    Parameters
+    ----------
+    input_ouptut_pairs: iterator of of tuples (input_path, output_path).
```
aidanheerdegen, line 285:
```suggestion
    input_ouptut_pairs: iterator of tuples (input_path, output_path).
```

aidanheerdegen (approved):
One tiny typo, but otherwise looks good to me.

This option seems like a decent balancing act, and it doesn't remove the original files, right?

umpost/conversion_driver_esm1p5.py
blimlim: Good catch! I've added this in

blimlim:
> This option seems like a decent balancing act, and it doesn't remove the original files if they aren't converted, right? 

Yep! Only files that get pushed through the conversion are candidates for deletion, and so any files with clashing outputs won't get deleted.

blimlim:
I've just pinged for a re-review – the only change is the typo fix.

aidanheerdegen (approved):
Looks good to merge to me.

PR #134, opened 2024-10-08
Create integration test utility to binary diff original & um2nc dev sourced netCDF data

### LINKED ISSUE #131: Add script & instructions for initial binary compatibility testing
Tasks:

* [x] Add @truth-quark's binary diff check script to repo
* [x] Document its use
* [ ] Upload test netCDF files to the same `gadi` dir as the original
* [ ] Should it be included in the `gadi` project test?
* [ ] Should it be included in the GitHub CI tests?
  * [ ] Does GH CI provide enough drive space to keep a binary file?
  * [ ] Is `aiihca.paa1jan.subset` public data?
* [ ] Add tests to check the `--64` option?
* [ ] Add test to convert 512mb `aiihca.paa1jan` (convert 100+ cubes)

truth-quark:
Closes #131.

This PR adds a utility to compare netcdf data between different commit versions of `um2nc-standalone`. This is a loose check to determine if basic binary compatibility is maintained between this repo's 1st commit `um2nc-standalone` and the version a user has checked out.

**NOTE:** This is not exhaustive & some options are not tested, therefore may not be backwards binary equivalent!

**NOTE 2:** This complements unit testing to ensure the CLI works & default runs are binary equivalent.


integration/binary_diff.sh
```diff
@@ -0,0 +1,87 @@
...
+fi
+
+if [ ! -d "$UM2NC_TEST_DATA" ]; then
+  echo "ERROR: UM2NC_TEST_DATA dir does not exist";
+  exit
+fi
+
+function diff_warn {
+  # compare & warn if binary args don't match
+  diff -q "$1" "$2" | grep -E "^Files [a-zA-Z0-9. ]+differ"
```
blimlim, line 45:
```suggestion
  diff -q "$1" "$2" | grep -E "^Files .+differ"
```
on gadi the current grep command misses the output when the files differ, resulting in no output when it was supposed to warn:

```
Files /g/data/tm70/sw6175/development/um2nc-standalone/integration/test_data/aiihca.paa1jan.subset.nomask.orig.nc and /g/data/tm70/sw6175/development/um2nc-standalone/integration/test_data/nomask.nc differ
```

The suggested change works for me, but let me know if you think it looks alright.

blimlim (changes requested):

integration/binary_diff.sh
truth-quark: That's a simpler more flexible regex. My orig regex might fail on "special" characters in the filename???

I've updated in 4769dc3a.

blimlim (approved):
Looks very useful!

truth-quark:
Argh, it needs another review as I made it a non-draft....

PR #139, opened 2024-10-09
Capture filename key warning in test

### LINKED ISSUE #137: Capture/test warnings in `test_get_nc_write_path_unrecognized_unit()`
`test_get_nc_write_path_unrecognized_unit()` test currently produces the warning 
```
test/test_conversion_driver_esm1p5.py::test_get_nc_write_path_unrecognized_unit
  /g/data/tm70/sw6175/development/um2nc-standalone/umpost/conversion_driver_esm1p5.py:153: UserWarning: Unit code 'w' from filename faiihca.pwabcd not recognized. Frequency information will not be added to the netCDF filename.
```
This warning is desired in this situation, hence it is worth modifying the test to capture and test the warning, which will also simplify the test output.

blimlim:
Closes #137 

This PR captures the unknown file name key warning produced by `get_nc_filename()` and tests that it occurs in `test_get_nc_write_path_unrecognized_unit()`.

### DIFF
--- a/test/test_conversion_driver_esm1p5.py
+++ b/test/test_conversion_driver_esm1p5.py
@@ -80,11 +80,16 @@ def test_get_nc_write_path_unrecognized_unit():
     nc_write_dir = Path("netCDF")
     expected_nc_write_path = nc_write_dir / f"aiihca.p{unknown_key}-005007.nc"
 
-    nc_write_path = esm1p5_convert.get_nc_write_path(
-                        Path(ff_name),
-                        nc_write_dir,
-                        (ff_year, ff_month, 1)
-                    )
+    with pytest.warns(RuntimeWarning) as record:
+        nc_write_path = esm1p5_convert.get_nc_write_path(
+                            Path(ff_name),
+                            nc_write_dir,
+                            (ff_year, ff_month, 1)
+                        )
+        if not record:
+            msg = ("No warning with unknown filename key "
+                   f"'{unknown_key}' in '{ff_name}'")
+            pytest.fail(msg)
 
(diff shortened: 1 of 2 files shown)


test/test_conversion_driver_esm1p5.py
```diff
@@ -80,11 +80,16 @@ def test_get_nc_write_path_unrecognized_unit():
...
+    with pytest.warns(RuntimeWarning) as record:
+        nc_write_path = esm1p5_convert.get_nc_write_path(
+                            Path(ff_name),
+                            nc_write_dir,
+                            (ff_year, ff_month, 1)
+                        )
+        if not record:
+            msg = ("No warning with unknown filename key "
+                   f"'{unknown_key}' in '{ff_name}'")
+            pytest.fail(msg)
```
atteggiani, line 92: What is the purpose of L89-92?
the 
```py
with pytest.warns(....):
    ...
```
block will already fail if the warning is not raised. 
I think this could be simplified to:
```suggestion
    with pytest.warns(RuntimeWarning):
        nc_write_path = esm1p5_convert.get_nc_write_path(
                            Path(ff_name),
                            nc_write_dir,
                            (ff_year, ff_month, 1)
                        )
```

blimlim: I think this suggestion makes sense. This was based on a previous example, where an additional message was used to provide useful detail to the user on why a test failed. In this case, the test is probably simple enough that its self explanatory and doesn't need additional information.

atteggiani (approved):
LGTM!
Thank you @blimlim

Issue #143, opened 2024-10-09
Retrofit unit testing to time dimension fix function


truth-quark:
Add unit testing for the function extraction specified in #142.

PR #144, opened 2024-10-09
Extract time coord dimension fix

### LINKED ISSUE #142: Extract `cubewrite()` time dimension fix function
This issue has several parts relating to the time coord dimension modification:

https://github.com/ACCESS-NRI/um2nc-standalone/blob/1c61fd9523f2efdd9625fbca44c2808bc5851224/umpost/um2netcdf.py#L383-L414

1. Decouple `sman.write()` from the time conversion logic
2. Extract the time dimension fix to a function
3. Merge all writes to a single function call

Unit testing is covered in #143.

truth-quark:
Closes #142.

This PR covers refactoring: extracting the time coord fix function & decoupling from the file I/O write process. Tests are intended for work on #143. I've tested this with the [binary diff script](https://github.com/ACCESS-NRI/um2nc-standalone/blob/main/integration/binary_diff.sh).

This is a step closer to combining `process()` and `cubewrite()` for a single workflow function to rule them all.

Any comments welcome!

### DIFF
--- a/umpost/um2netcdf.py
+++ b/umpost/um2netcdf.py
@@ -395,19 +395,14 @@ def cubewrite(cube, sman, compression, use64bit, verbose):
                     print("Transpose to", neworder)
 
                 cube.transpose(neworder)
-
-            sman.write(cube,
-                       zlib=True,
-                       complevel=compression,
-                       unlimited_dimensions=['time'],
-                       fill_value=fill_value)
         else:
-            tmp = iris.util.new_axis(cube, cube.coord('time'))
-            sman.write(tmp,
-                       zlib=True,
-                       complevel=compression,
-                       unlimited_dimensions=['time'],
-                       fill_value=fill_value)
+            cube = iris.util.new_axis(cube, cube.coord('time'))
+
+        sman.write(cube,
+                   zlib=True,
+                   complevel=compression,
+                   unlimited_dimensions=['time'],
+                   fill_value=fill_value)
 
     except iris.exceptions.CoordinateNotFoundError:
         # No time dimension (probably ancillary file)
--- a/umpost/um2netcdf.py
+++ b/umpost/um2netcdf.py
@@ -380,7 +380,7 @@ def cubewrite(cube, sman, compression, use64bit, verbose):
         if coord.points.dtype == np.int64:
             coord.points = coord.points.astype(np.int32)
 
-    cube, unlimited_dimension = fix_time_coord(cube, verbose)
+    cube, unlimited_dimensions = fix_time_coord(cube, verbose)
 
     # TODO: refactor & move to end of process()
     # TODO: refactor cubewrite() to return (cube, unlimited dims, fill value)
@@ -388,7 +388,7 @@ def cubewrite(cube, sman, compression, use64bit, verbose):
     sman.write(cube,
                zlib=True,
                complevel=compression,
-               unlimited_dimensions=unlimited_dimension,
+               unlimited_dimensions=unlimited_dimensions,
                fill_value=fill_value)
 
 
--- a/umpost/um2netcdf.py
+++ b/umpost/um2netcdf.py
@@ -965,6 +965,23 @@ def convert_32_bit(cube):
 
 
 def fix_time_coord(cube, verbose):
+    """
+    Ensures cube has a 'time' coordinate dimension.
+
+    Coordinate dimensions are reordered to ensure 'time' is the first dimension.
+    Cubes are modified in place, although it is possible iris will return new
+    copies of cubes. UM ancillary files are ignored.
+
+    Parameters
+    ----------
+    cube : iris Cube object
+    verbose : True to display information on stdout
+
+    Returns
+    -------
+    A (cube, unlimited_dimensions) tuple. Unlimited dimensions is None for
+    ancillary files.
+    """
     try:
         # If time is a dimension but not a coordinate dimension, coord_dims('time')
         # returns empty tuple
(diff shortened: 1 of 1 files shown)


blimlim (dismissed):
Changes look good to me! Is it worth adding a short docstring to the new function?

Would something along the lines of "Ensures 'time' is a coordinate dimension, and appears first in the dimension order when present" make sense?

truth-quark:
> Ensures 'time' is a coordinate dimension, and appears first in the dimension order when present"

Good catch, I've added docs in 7a123924.

truth-quark:
@aidanheerdegen & @blimlim this one is ready for another, _hopefully quick_ review.

umpost/um2netcdf.py
```diff
@@ -965,8 +965,26 @@ def convert_32_bit(cube):
 
 
 def fix_time_coord(cube, verbose):
+    """
+    Ensures cube has a 'time' coordinate dimension.
+
+    Coordinate dimensions are reordered to ensure 'time' is the first dimension.
+    Cubes are modified in place, although it is possible iris will return new
+    copies of cubes. UM ancillary files are ignored.
```
blimlim, line 973: Just to be 100% accurate, it might be worth changing it to "UM ancillary files with no time dimension are ignored", as it is possible for ancillary files to not have any time dimension

umpost/um2netcdf.py
```diff
@@ -965,8 +965,26 @@ def convert_32_bit(cube):
...
+
+    Parameters
+    ----------
+    cube : iris Cube object
+    verbose : True to display information on stdout
+
+    Returns
+    -------
+    A (cube, unlimited_dimensions) tuple. Unlimited dimensions is None for
+    ancillary files.
```
blimlim, line 983: Likewise might be worth saying "None for ancillary files with no time dimension"

blimlim (changes requested):
Just a tiny suggested change to the comments for accuracy with different types of ancillaries.

umpost/um2netcdf.py
truth-quark: OK, it's in dff0b7d7.

umpost/um2netcdf.py
truth-quark: This is also fixed in dff0b7d7.

blimlim (dismissed):
Looks good to me!

umpost/um2netcdf.py
```diff
@@ -986,6 +964,54 @@ def convert_32_bit(cube):
...
+    ancillary files with no time dimension.
+    """
+    try:
+        # If time is a dimension but not a coordinate dimension, coord_dims('time')
+        # returns empty tuple
+        if tdim := cube.coord_dims('time'):
+            # For fields with a pseudo-level, time may not be the first dimension
+            if tdim != (0,):
+                tdim = tdim[0]
+                neworder = list(range(cube.ndim))
```
aidanheerdegen, line 992:
```suggestion
        if (tdim,) := cube.coord_dims('time'):
            # For fields with a pseudo-level, time may not be the first dimension
            if tdim != 0:
                neworder = list(range(cube.ndim))
```
Not sure if I've got the tuple assignment correct, but the general idea is to simplify
the code to remove having to carry around a redundant part of the tuple which makes it 
messy.

Also worth making the time dimension a variable?

umpost/um2netcdf.py
```diff
@@ -380,38 +380,16 @@ def cubewrite(cube, sman, compression, use64bit, verbose):
...
-        # No time dimension (probably ancillary file)
-        sman.write(cube, zlib=True, complevel=compression, fill_value=fill_value)
+    # TODO: refactor & move to end of process()
+    # TODO: refactor cubewrite() to return (cube, unlimited dims, fill value)
+    #       then move above steps into process() / remove cubewrite()
+    sman.write(cube,
+               zlib=True,
+               complevel=compression,
+               unlimited_dimensions=unlimited_dimension,
+               fill_value=fill_value)
```
aidanheerdegen, line 393:
```suggestion
    cube, unlimited_dimensions = fix_time_coord(cube, verbose)

    # TODO: refactor & move to end of process()
    # TODO: refactor cubewrite() to return (cube, unlimited dims, fill value)
    #       then move above steps into process() / remove cubewrite()
    sman.write(cube,
               zlib=True,
               complevel=compression,
               unlimited_dimensions=unlimited_dimensions,
               fill_value=fill_value)

```

Small change, but I suggest using the plural. It's an array, and in `fix_time_coord` it is referred to as a plural too. Just less confusing.

aidanheerdegen (approved):
A couple of suggestions. Don't mind either way.

umpost/um2netcdf.py
aidanheerdegen: GitHub seems to have done something quite odd with this (including the original code). It is a small change: just making the variable name a plural for consistency.

truth-quark: That's made it to e31c5425.

umpost/um2netcdf.py
truth-quark: The suggested tuple change breaks the binary test script around unpacking.

Then checking the aiihca.paa1jan, all `coord_dims()` are empty tuples:

```python
>>> cubes = iris.load("large_data/aiihca.paa1jan")
>>> set([c.coord_dims('time') for c in cubes])
{()}
>>> cubes[0].coord_dims('pseudo_level')  # what do other coord dims look like?
(0,)
```
I'll skip any changes around these tuple mods.

blimlim (approved):
Looks good to me!

PR #146, opened 2024-10-09
Update conda environments, conda package recipe and documentation

### LINKED ISSUE #60: Add `virtualenv` / `conda` setup docs for local env creation
**Tasks:**

* [x] Snarf the conda setup docs from `amami`
* [x] Add commands docs for using `virtualenv`

**Notes**

`conda` is missing some dependencies such as `mule` (specifically installing `mule` from a GitHub URL). For the time being, we'll have to install directly with `pip` until the dependencies are sorted out.

### LINKED ISSUE #106: Create `conda`/`micromamba` development environment for um2nc-standalone
Currently there is no `conda`/`micromamba` `development` environment for `um2nc-standalone`.

It would be good to create such an environment to start doing some tests on actual files, and to make the release of a `release` environment easier whenever we have the first (`v1.0.0`) `um2nc-standalone` release.

### LINKED ISSUE #129: Avoid creating environment before installing package in CI
In the [CI.yml](https://github.com/ACCESS-NRI/um2nc-standalone/blob/main/.github/workflows/CD.yml), before testing that the conda package installs without problems, I would avoid creating an environment using the `environment.yml` file. 
Instead, I would create a blank `conda` environment, so we can be sure the um2nc conda package has all the requirements necessary to install from scratch.

atteggiani:
Fixes #106.
Fixes #129.
Fixes #60.

Superseeds #62.

- [x] Set up build and dev conda environment files. 
- [x] Modified pyproject.toml to reflect modifications in the conda environment files. Modified meta.yaml for conda packaging.
- [x] Modified CI.yml for conda-build testing.
- [x] Modified README.md with instructions on how to install (Gadi, locally and dev)

The conda environments and the steps in the documentation have been all tested locally.


README.md
```diff
@@ -1,37 +1,98 @@
...
-## User documentation
+### Local installation
+`um2nc` is available as a `conda` package in the [access-nri conda channel](https://anaconda.org/accessnri/um2nc).
+To install it run:
+```
+conda install accessnri::um2nc
+```
 
-TODO
+### Installing Dependencies
```
atteggiani, line 29: Are these steps required? 

`udunits` can be installed as a conda package. Why should it be installed in other ways?
Also, is it not already installed with some of the dependencies (for example `iris`)?
@truth-quark @blimlim

.github/workflows/CI.yml
```diff
@@ -8,82 +8,93 @@ on:
     branches: [ main, develop ]
```
atteggiani, line 8: This is triggering on PRs and pushed to `main` and `develop`.
Do we really need them for the `develop` branch? 
Is the `develop` branch currently used as a "staging" branch? 
@blimlim @truth-quark

blimlim: I think the `develop` branch has been deprecated and PRs are now done directly to `main`, so it should be save to remove the `develop` branch here.

README.md
blimlim: Great question. I have to admit my understanding of setting up conda packages and installations is very poor, so most of this goes over my head. I just tried following the install instructions on my mac omitting the `udunits` step (and made sure to uninstall my separate homebrew installation of `udunits`).

Everything worked out ok for me, and I'm able to run the tests and perform conversions. Conda installed `udunits2` probably via [`cf-units`](https://github.com/SciTools/cf-units/blob/main/requirements/cf-units.yml). 

@truth-quark do you know if this separate `udunits` installation is necessary?

.github/workflows/CI.yml
atteggiani: Removed `develop` in [cdac87bff05ba6f6e83d99a462bb1d9f81003406](https://github.com/ACCESS-NRI/um2nc-standalone/pull/146/commits/cdac87bff05ba6f6e83d99a462bb1d9f81003406)

README.md
marc-white: I think I managed to get it running earlier with the `conda` version, but I can confirm that.

marc-white: Doing `brew uninstall udunits` on my system doesn't break the tests, so I think the `conda` `udunits` has us covered.

atteggiani: I will remove it for now.

atteggiani: Removed in [4bc1add33f5e3890227cd5c38c17ac4be458b497](https://github.com/ACCESS-NRI/um2nc-standalone/pull/146/commits/4bc1add33f5e3890227cd5c38c17ac4be458b497)

truth-quark: > @truth-quark do you know if this separate `udunits` installation is necessary?

I used it specifically for a `virtualenv` install, which `conda` seems to handle itself???

atteggiani: Yes `conda` packages that need `udunits` (like `iris`) will install [udunits2](https://anaconda.org/conda-forge/udunits2) automatically.
We encourage people who want a local installation to use `conda` (`micromamba` even better).

.github/workflows/CI.yml
```diff
@@ -106,9 +118,9 @@ jobs:
         run: python -m pytest --cov=umpost --cov-report=xml -s test
 
       - name: Upload code coverage
-        uses: codecov/codecov-action@v4
         # Only upload once for the installed version
-        if: matrix.python-version == ${{ env.PY_VERSION }}
+        if: ${{ matrix.python-version == env.PY_VERSION }}
+        uses: codecov/codecov-action@b9fd7d16f6d7d1b5d2bec1a2887e65ceed900238 #v4.6.0
         with:
           token: ${{ secrets.codecov_token }}
```
atteggiani, line 125: I don't have access to the Repo settings, but this action throws a warning and does not upload the coverage file, which means the `secrets.codecov_token` might be expired or not set.

README.md
```diff
@@ -1,37 +1,72 @@
...
+The `um2nc-standalone` project is an [ACCESS-NRI](https://www.access-nri.org.au/) initiative to merge multiple versions of Unified Model NetCDF conversion tool to a single, canonical project. 
 
 ## Installation
 
-TODO
+### Gadi
 
-* `virtualenv` instructions
-* `conda`/`miniconda`/`micromamba?` instructions
+On Gadi, `um2nc` is available within the `vk83` `payu` environment.
```
marc-white, line 13: Are there any other environments the package is/will be available on? If so, we should probably provide a list (saves a lot of exploring).

atteggiani: For now, I don't think there are other environments where the package is available.
In the future, I am not sure, but I would lean towards no anyway.

pyproject.toml
```diff
@@ -14,27 +15,24 @@ readme = "README.md"
...
 dependencies = [
-  "numpy",
-  "mule",
-  "cftime",
-  "netCDF4",
-  "scitools-iris",
-  "cf-units",
-  "PyYAML",
-  "f90nml",
+    "python >=3.10,<=3.12",
```
marc-white, line 18: Do we want to lock the max Python version to 3.12? I note that 3.13 was released about a week ago.

(And, in general, do we want to lock a max version of Python now, or leave it open, and only come back and set a max once we know there's a new Python that isn't compatible?)

atteggiani: That's a good question. 
I am not sure what's the usual approach between:
- Constraining Python to an upper version and, whenever new Python versions get released, test them and, in case they are compatible, add the version to the upper constraint
- Only constraining for minimum Python versions and, whenever new Python versions get released, test them and, in case they are not compatible, add the version to the constraint

I don't have strong preferences here, but the 2nd approach might pose less burden on maintainance (providing we are lucky and newer versions of Python are compatible)

atteggiani: On a second thought, when a new version is released we would still need to update the CI to use the new version as well (currently the matrix uses until py3.12).
Therefore, even from the maintenance burden standpoint, the two approaches are very similar.

marc-white: Might as well keep the upper constraint there, if for no other reason that to remind us/future devs where the testing has been completed to.

atteggiani: Based on my comments above, my preference would then be the 1st approach (current version), to always ensure compatibility.

.conda/env_dev.yml
```diff
@@ -0,0 +1,25 @@
...
+    - f90nml
+    - ipykernel
+    - mule
+    - netcdf4
+    - numpy
+    - pip
+    - pytest
+    - pytest-cov
+    - pylint #TODO: remove this after setting up linting with ruff
+    - ruff
```
marc-white, line 23: Are we in a position to definitively cast `pylint` aside in favour of `ruff`? My understanding (from a very fast and superficial Google search) is that `ruff` is a lot faster, and is seen as the 'sexier' option, but `pylint` still gives more complete coverage.

Given this is the development dependency list anyway, I'm wondering if leaving both options available might keep more people happy.

marc-white (approved):
Looks pretty sensible & workable to me.

.conda/env_dev.yml
atteggiani: I am a huge fan of Ruff, and [this comment](https://github.com/ACCESS-NRI/um2nc-standalone/issues/26#issuecomment-2398034872) in #26 sets the plan for the setup of ruff pre-commit linting/formatting.

I think it will be the next step after this PR. 
For now we will leave it like this and change it when there it the PR about pre-commit hooks.

I am ok leaving `pylint` there (even though I am not sure what use that will do).

blimlim:
I've just been testing the dev environment out and ran into a couple of things:

Last week `iris` [added support for numpy version 2](https://github.com/SciTools/iris/blame/8753cfb9c30f9bb71b6878e4d29000771b6fea9e/requirements/py310.yml#L21), and as a result creating an environment from `env_dev.yml` uses `numpy v2`. `mule` uses methods that are removed in `numpy v2` and so converting a fields file using the development environment produces the following error:
```
  File "/Users/u1161001/miniconda3/envs/um2nc_base_PR/lib/python3.12/site-packages/numpy/__init__.py", line 414, in __getattr__
    raise AttributeError("module {!r} has no attribute "
AttributeError: module 'numpy' has no attribute 'product
```
On my laptop, conda installed `mule 2019.01.01` which doesn't restrict the numpy version, hence leading to the above error. The payu environment on gadi uses `mule 2022.07.01` which does restrict `numpy >=1.21.6,<2.0a0` but I don't think this is  available on all platforms.

I'm wondering, will we need to restrict the `numpy` version to `numpy<2` in our `env_...yaml` files to ensure mule works?

---

Installing using `env_dev.yml` also installed `hdf5 1.14.4` for me, while the version in the `payu 1.1.5` environment is `1.14.3`. This change in `hdf5` version [breaks the bitwise reproducibility integration tests](https://github.com/ACCESS-NRI/um2nc-standalone/pull/134).

Running the integration tests:
- Using an environment with the new `hdf5 1.14.4`:
   ```
   $ diff mask.nc ../../aiihca.paa1jan.subset.orig.nc 
   > Binary files mask.nc and ../../aiihca.paa1jan.subset.orig.nc differ
   ```
- Using an environment were we restrict `hdf5 1.14.3`:
   ```
   $ diff mask.nc ../../aiihca.paa1jan.subset.orig.nc
   $
   ```
In the above `aiihca.paa1jan.subset.orig.nc` was produced by converting `/g/data/tm70/esm15_testing/sample_output` with the original `um2netcdf4` script.

There's no clear differences in the resulting netCDF files though. Using `nccmp` to compare the data, global attributes, encodings and metadata:
```
$ nccmp -Bdgehs mask.nc ../../aiihca.paa1jan.subset.orig.nc mask.nc
> Files "mask.nc" and "../../aiihca.paa1jan.subset.orig.nc" are identical.
```
I'll make a separate issue for discussion on this, as I'm not sure how strictly we want to adhere to bitwise reproducibility.

blimlim:
> Why you think it is not available on all platforms?
> I think the best solution would be to restrict mule dependency to use version 2022.07.01.

Definitely agree – since it's `mule` that requires `numpy<2` I think it would make the most sense for that dependency to be handled by mule. When I try and build the environment on my mac with `mule=2022.07.1` (I had a typo in my original message, it should be `2022.07.1` instead of `2022.07.01`), I get the error:
```
PackagesNotFoundError: The following packages are not available from current channels:

  - mule=2022.07.1*
```
Building on gadi with the same restriction works without any issues. On the coecms Anaconda page for `mule`, does the `linux-64` prefix before the `2022.07.01` version mean that it's not available on all architectures? 

The `2020.01.1` version also doesn't restrict numpy, and I had a separate error trying to install that one:
```
└─ mule 2020.01.1**  is not installable because it requires
   └─ python_abi 3.6.* *_cp36m, which requires
      └─ python 3.6.* , which does not exist (perhaps a missing channel).
```

> On a general case, I think we might want to think about keeping bitwise reproducibility in the first release of the um2nc conda package, to ensure there is at least a version which is fully backward-compatible with the old script, and therefore can fully reproduce its results.

I think this sounds like a great idea. I've made an issue #149 to continue discussion.

atteggiani:
> On the coecms Anaconda page for `mule`, does the `linux-64` prefix before the `2022.07.01` version mean that it's not available on all architectures?

Yeah, I had not seen that. 
The `2022.07.01` from the `coecms` conda channel is only available on Linux systems.

This is a problem at the moment. 
I am not sure why there is no `mule 2022.07.01` with`noarch` specification. 

For the moment probably we will have to restrict `numpy` to `numpy<2`.

atteggiani:
Added `mule<2` restriction in `pyproject.toml` in [b5aa60b](https://github.com/ACCESS-NRI/um2nc-standalone/pull/146/commits/b5aa60b8937ae642df9e25fbf4fab2d16ff7c72e)

blimlim:
Just noticed with the version implementation, we need to add `__version__` to the history attributes in `um2netcdf.py`:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/e1d78fd8e46f02d58b207be56cd8e6a84f01b5d4/umpost/um2netcdf.py#L761-L768

Would the best way to do this be importing `umpost` and then changing the above to:
```python
 def add_global_history(infile, iris_out): 
     version = umpost.__version__
     t = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S") 
     um2nc_path = os.path.abspath(__file__) 
     history = f"File {infile} converted with {um2nc_path} {version} at {t}" 
  
     iris_out.update_global_attributes({'history': history}) 

```
(I was hoping to add a review code suggestion for this, but don't think it lets you review unchanged files!)

atteggiani:
> Just noticed with the version implementation, we need to add `__version__` to the history attributes in `um2netcdf.py`:
> 
> https://github.com/ACCESS-NRI/um2nc-standalone/blob/e1d78fd8e46f02d58b207be56cd8e6a84f01b5d4/umpost/um2netcdf.py#L761-L768
> 
> Would the best way to do this be importing `umpost` and then changing the above to:
> 
> ```python
>  def add_global_history(infile, iris_out): 
>      version = umpost.__version__
>      t = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S") 
>      um2nc_path = os.path.abspath(__file__) 
>      history = f"File {infile} converted with {um2nc_path} {version} at {t}" 
>   
>      iris_out.update_global_attributes({'history': history}) 
> ```
> 
> (I was hoping to add a review code suggestion for this, but don't think it lets you review unchanged files!)

Good point!

I added your suggestion in [0512700d8fdbd1aa0bb85d129118b1539d15aa23](https://github.com/ACCESS-NRI/um2nc-standalone/pull/146/commits/0512700d8fdbd1aa0bb85d129118b1539d15aa23)

blimlim (dismissed):
Changes look great! Thanks for putting this together!

blimlim (approved):
Changes look good!

atteggiani:
Thank you @blimlim and @marc-white for your reviews.
Merged.

PR #163, opened 2024-11-25
Rename `umpost` to `um2nc`. Added dependency.

### LINKED ISSUE #161: Add nccmp to conda development environment dependencies
To perform the regression tests, the library `nccmp` is required.
To provide an easy way to install `nccmp` it would be good to add it to the dependencies of the conda development environment file [env_dev.yml](https://github.com/ACCESS-NRI/um2nc-standalone/blob/main/.conda/env_dev.yml).

### LINKED ISSUE #162: Rename `umpost` to `um2nc`
Since the package is going to be called `um2nc`, it is confusing having the top-level module named `umpost`. 

The top-level module would be better named `um2nc` and the whole source code moved into the top-level `src` directory.

atteggiani:
Fixes #162.
Fixes #161.

In this PR the following changes have been carried out:
- [x] top-level `umpost` folder has been renamed to `um2nc`, to match the name of the package.
- [x] top-level (newly-renamed) `um2nc` folder has been moved into `src` folder, to reflect classic/suggested structure of Python projects
- [x] `nccmp` has been added to the list of "development" dependencies in the `env_dev.yml` file.
- [x] Python version requirements have been "loosened-up" by keeping the upper limit free (only limiting Python `>=3.10`). This is usually better than restricing the package only to specific versions of Python, when the package might as well be compatible with versions of Python not yet released (for example `py3.14` etc.). This scenario is more likely than the package not being compatible with future versions of Python. In general, if there is such a compatibility issue, future release might restrict the Python version as needed.


blimlim (approved):
Thanks for putting these changes together!

They all look good to me

atteggiani:
Thank you for the review @blimlim!
Merged.

PR #178, opened 2025-01-16
Update paths to regression test data

### LINKED ISSUE #117: Fix subtle naming differences between `DummyCube` & iris cubes
There seen to be some subtle differences between iris cubes and our `DummyCube` which would cause the  `test_fix_level_coord_...` tests to fail if we were using real iris cubes.

The `cube.coord()` method uses the current name for a `DimCoord` as its argument, where the "current name" comes from either the `DimCoord`'s `standard_name`, `long_name`,  or `var_name`, with the following order of priority: `standard_name` > `long_name` > `var_name`.

This means that if a `DimCoord` only has a `var_name` attribute, changing the `var_name` forces us to also change the argument in `cube.coord(...)` used to access it:

A made up example after modifying a loaded cube's coordinate names:
```Python 
>>> cube.coord("pseudo_level").__dict__
{'_metadata_manager': MetadataManager(standard_name=None, long_name=None, var_name='pseudo_level', units=Unit('1'), attributes={}, coord_system=None, climatological=False, circular=False), '_values_dm': DataManager(array([1, 2, 3, 4, 5, 6])), '_bounds_dm': None, '_ignore_axis': False}

>>> cube.coord("pseudo_level").var_name = "new_var_name"

>>> cube.coord("pseudo_level")

  File "<stdin>", line 1, in <module>
  File "/g/data/vk83/apps/payu/1.1.5/lib/python3.10/site-packages/iris/cube.py", line 2304, in coord
    raise iris.exceptions.CoordinateNotFoundError(emsg)
iris.exceptions.CoordinateNotFoundError: "Expected to find exactly 1 'pseudo_level' coordinate, but found none."

>>> cube.coord("new_var_name")
<DimCoord: new_var_name / (1)  [1, 2, ..., 5, 6]  shape(6,)>
```
In these tests however, we're changing the coordinate's `var_names` but not the `cube.coord(...)` arguments used to access them.

~~Given that the tests are working and it's a fairly subtly difference, should we defer this to a separate issue?~~

_Originally posted by @blimlim in https://github.com/ACCESS-NRI/um2nc-standalone/pull/113#discussion_r1780210448_

blimlim:
Closes #117. This pull request updates the paths for the regression test data to match its location on gadi.

### DIFF
--- a/integration/INTEGRATION_README.md
+++ b/integration/INTEGRATION_README.md
@@ -78,4 +78,4 @@ prior to the development of `um2nc`: https://github.com/ACCESS-NRI/um2nc-standal
 The conversion was performed within the following `payu1.1.5` environment, active on Gadi:
 https://github.com/ACCESS-NRI/payu-condaenv/releases/tag/1.1.5
 
-All test data is located in `/g/data/vk83/testing/um2nc/integration-tests`.
+All test data is located in `/g/data/vk83/testing/data/um2nc/integration-tests`.
--- a/integration/regression_tests.sh
+++ b/integration/regression_tests.sh
@@ -28,7 +28,7 @@ EOF
 
 
 
-TEST_DATA_PARENT_DIR=/g/data/vk83/testing/um2nc/integration-tests
+TEST_DATA_PARENT_DIR=/g/data/vk83/testing/data/um2nc/integration-tests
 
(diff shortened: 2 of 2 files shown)


atteggiani (approved):
LGTM!

blimlim:
Thanks for the review @atteggiani!

PR #179, opened 2025-01-19
Updated CD

atteggiani:
Updated `CD.yml` to build and upload to Anaconda.org packages for any platform. Reintroduced the use of the updated ['uibcdf/action-build-and-upload-conda-packages'](https://github.com/uibcdf/action-build-and-upload-conda-packages) action for building and uploading conda packages.

### DIFF
--- a/.conda/meta.yaml
+++ b/.conda/meta.yaml
@@ -6,7 +6,6 @@ package:
     version: "{{ version }}"
 
 build:
-    noarch: python
     number: 0
     script: "python3 -m pip install . -vv"
     entry_points:
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -27,9 +27,6 @@ jobs:
     name: Build with conda and upload
     runs-on: ubuntu-latest
     needs: get-package-name
-    env:
-      BUILD_FOLDER: ${{github.workspace}}/build
-      PACKAGE_PATH: ${{github.workspace}}/build/noarch/*.tar.bz2
     permissions:
       contents: write
     steps:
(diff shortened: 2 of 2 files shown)


blimlim (dismissed):
Thanks @atteggiani, looks good to me!

jo-basevi (approved):
I was just about to ask a question about whether it was going to deploy to different platforms but the added `platform_all: true` in `action-build-and-upload-conda-packages` clears it up for me! Looks good to me 👍 

Would it be worth testing the build without the upload using this action in the CI? https://github.com/uibcdf/action-build-and-upload-conda-packages?tab=readme-ov-file#testing-build-correctness

atteggiani:
> I was just about to ask a question about whether it was going to deploy to different platforms but the added `platform_all: true` in `action-build-and-upload-conda-packages` clears it up for me! Looks good to me 👍
> 
> Would it be worth testing the build without the upload using this action in the CI? https://github.com/uibcdf/action-build-and-upload-conda-packages?tab=readme-ov-file#testing-build-correctness

Yes, we had a chat with @blimlim about it. 
I have never used the `platform_all: True` option with the [uibcdf/action-build-and-upload-conda-packages](https://github.com/uibcdf/action-build-and-upload-conda-packages) action, but reading from its REAMDE and looking at the action.yml, it seems what it does is simply running `conda convert --platform ...` on the built packages. 
This means that "pure Python" packages and packages whose build can be converted to different platforms without needing a re-build could be converted. In this case, however, I am not sure if it will work, because `mo_pack` is a Python binding to a `C` library and its conda package has [2 different distributions](https://anaconda.org/conda-forge/mo_pack) (`linux-64` and `osx-64`). 
We will try this way, otherwise I think we could release `um2nc` only as `linux-64` distribution, because most of the users would have either Windows or Mac (arm64), which are not available as `mo_pack` distributions.

blimlim (approved):
Looks good to me!

PR #182, opened 2025-01-21
Fix errors in the build

atteggiani:
- [x] Added channels to `~/.condarc` to fix errors in the build
- [x] Changed conda-build python version to `3.10` because of incompatibility with `mule` and `py3.11`

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -67,7 +67,7 @@ jobs:
         uses: uibcdf/action-build-and-upload-conda-packages@b06165145a25b9c8bcb2d2b24682ad0d8e494ce7 #v1.4.0
         with:
             meta_yaml_dir: .conda
-            python-version: 3.10 # mule is incompatible with py3.11
+            python-version: ${{ vars.PY_VERSION }}
             user: ${{ secrets.ANACONDA_USER_NAME }}
             token: ${{ secrets.ANACONDA_TOKEN }}
             platform_all: true
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -53,6 +53,11 @@ jobs:
             cat > ~/.condarc << EOF
             conda-build:
                 pkg_format: .tar.bz2
+            channels:
+                - accessnri
+                - conda-forge
+                - coecms
+                - nodefaults
             EOF
       
       - name: Build and upload conda package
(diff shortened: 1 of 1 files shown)


blimlim (dismissed):
Looks good to me! Only suggestion I could think of is mentioning in the 
`# We create a `~/.condarc` file with the correct options to enforce the use of `.tar.bz2` packages` comment that it's a temporary workaround.

atteggiani:
> Looks good to me! Only suggestion I could think of is mentioning in the `# We create a `~/.condarc`file with the correct options to enforce the use of`.tar.bz2` packages` comment that it's a temporary workaround.

Makes sense.
Added a bit more explanation in [92d5dc56cc076fc110417bd20b2a9c9b516b2398](https://github.com/ACCESS-NRI/um2nc-standalone/pull/182/commits/92d5dc56cc076fc110417bd20b2a9c9b516b2398)

blimlim (approved):
Looks good!

.github/workflows/CD.yml
```diff
@@ -47,20 +47,27 @@ jobs:
...
+                - nodefaults
             EOF
       
       - name: Build and upload conda package
         id: build-and-upload
         uses: uibcdf/action-build-and-upload-conda-packages@b06165145a25b9c8bcb2d2b24682ad0d8e494ce7 #v1.4.0
         with:
             meta_yaml_dir: .conda
-            python-version: ${{ vars.PY_VERSION }}
+            python-version: 3.10 # mule is incompatible with py3.11
```
jo-basevi, line 70: Did `vars.PY_VERSION` recently get updated to py3.11? Should `vars.PY_VERSION` be set to 3.10 while mule is incompatible as it's used in the CI and also in the step above in setting up the conda environment `conda-incubator/setup-miniconda`?

jo-basevi (approved):
Looks good!

.github/workflows/CD.yml
jo-basevi: Sorry I think I'm a bit confused by what you mean by different python versions for different packages? 

I just thought it was useful for when there is an update to CI/CD to use a later python version in a repository, to only only update it one place

atteggiani: Well it surely is useful if we want to update the python version in one place, but it depends what that python version is used for (where it is used in a workflow file). 
Here, we are using it in to give the python version in the `conda-incubator/setup-miniconda` action to set the python version in the created conda environment. 
Then we are using it to set the python version used by `conda build` in the `uibcdf/action-build-and-upload-conda-packages`. This is because we are not trying to create separate builds for different python versions (even though I think we should start, because it is in general recommended).
If we want to run different builds for different python versions, we would have to specify different python versions and `PY_VERSION` would not have a use anymore.

atteggiani:
Thank you @jo-basevi and @blimlim!
Merged.

.github/workflows/CD.yml
jo-basevi: Thanks for the explanation! I see what you mean now :) Yeah maybe we should start doing multiple python version conda builds if it's recommended

PR #183, opened 2025-01-21
Added label to the 'uibcdf/action-build-and-upload-conda-packages' action

atteggiani:
Another fix for the `CD.yml` release workflow:
- [x] Added `label` as an input to the `uibcdf/action-build-and-upload-conda-packages` action to fix [the error](https://github.com/ACCESS-NRI/um2nc-standalone/actions/runs/12880928576/job/35910680233#step:5:700) (again, it's a bug of the action but this is a work-around that would also work long-term)


blimlim (approved):
Looks good to me!

atteggiani:
THank you @jo-basevi and @blimlim!

PR #184, opened 2025-01-22
Fix CD release

atteggiani:
- [x] Added step to re-format paths in the `CD.yml` for the correct release
- [x] Removed conversion in the `uibcdf/action-build-and-upload-conda-packages` step, because `um2nc` can only be installed on platforms where [`mo_pack`](https://anaconda.org/conda-forge/mo_pack) is also present.

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -29,8 +29,6 @@ jobs:
     needs: get-package-name
     permissions:
       contents: write
-    outputs:
-        newline-separated-paths: ${{ steps.reformat-paths.outputs.paths }}
     steps:
       - name: Checkout source
         uses: actions/checkout@v4
@@ -79,7 +77,7 @@ jobs:
         # Needed to have the correct newline-separated files format for the following release step
         run: |
             paths=$(tr ' ' '\n' <<< "${{steps.build-and-upload.outputs.paths}}")
-            echo "paths=$paths" >> $GITHUB_OUTPUT
+            echo "newline-separated-paths=$paths" >> $GITHUB_OUTPUT
 
       - name: Create Release
         uses: softprops/action-gh-release@c062e08bd532815e2082a85e87e3ef29c3e6d191 #v2.0.8
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -29,6 +29,8 @@ jobs:
     needs: get-package-name
     permissions:
       contents: write
+    outputs:
+        newline-separated-paths: ${{ steps.reformat-paths.outputs.paths }}
     steps:
       - name: Checkout source
         uses: actions/checkout@v4
@@ -72,7 +74,14 @@ jobs:
             token: ${{ secrets.ANACONDA_TOKEN }}
             label: main
             platform_all: true
-          
+      
+      - name: Re-format output paths
+        id: reformat-paths
+        # Needed to have the correct newline-separated files format for the following release step
+        run: |
+            paths=$(tr ' ' '\n' <<< "${{steps.build-and-upload.outputs.paths}}")
+            echo "paths=$paths" >> $GITHUB_OUTPUT
+
       - name: Create Release
         uses: softprops/action-gh-release@c062e08bd532815e2082a85e87e3ef29c3e6d191 #v2.0.8
         with:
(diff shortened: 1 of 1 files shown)


blimlim (approved):
Looks good, thanks @atteggiani!

.github/workflows/CD.yml
```diff
@@ -71,13 +73,19 @@ jobs:
...
       - name: Create Release
         uses: softprops/action-gh-release@c062e08bd532815e2082a85e87e3ef29c3e6d191 #v2.0.8
         with:
             tag_name: ${{ github.ref_name }}
             name: ${{needs.get-package-name.outputs.package-name}} ${{ github.ref_name }}
             generate_release_notes: true
             fail_on_unmatched_files: true
-            files: ${{steps.build-and-upload.outputs.paths}}
\ No newline at end of file
+            files: ${{steps.reformat-paths.outputs.newline-separated-paths}}
```
jo-basevi, line 91: Should it be `steps.reformat-paths.outputs.paths` here as `paths` is the output of the above step (`echo "paths=$paths" >> $GITHUB_OUTPUT`)

.github/workflows/CD.yml
```diff
@@ -29,6 +29,8 @@ jobs:
     needs: get-package-name
     permissions:
       contents: write
+    outputs:
+        newline-separated-paths: ${{ steps.reformat-paths.outputs.paths }}
```
jo-basevi, line 33: This job-level output  `newline-separated-paths` isn't used - is it needed?

jo-basevi (commented):
Sorry for the late review! Just have a couple small questions

atteggiani:
I'm just looking at the [GitHub context functions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions/#functions) and there is this [`join` function](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#join) that (based on my understanding of it), could be used in the `Create Release` step (and so we could get rid of the `Re-format output paths` step) and would look like:
```diff
 - name: Create Release
        uses: softprops/action-gh-release@c062e08bd532815e2082a85e87e3ef29c3e6d191 #v2.0.8
        with:
            tag_name: ${{ github.ref_name }}
            name: ${{needs.get-package-name.outputs.package-name}} ${{ github.ref_name }}
            generate_release_notes: true
            fail_on_unmatched_files: true
-           files: ${{steps.build-and-upload.outputs.paths}}
+           files: ${{ join(steps.build-and-upload.outputs.paths, '\n') }}
```

But I am not completely sure as I have never used `join`.
What do you think @jo-basevi?

jo-basevi:
> I'm just looking at the [GitHub context functions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions/#functions) and there is this [join function](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#join) that (based on my understanding of it), could be used in the Create Release step (and so we could get rid of the Re-format output paths step)

Join would be a really neat solution! I tried running a couple tests with it as I haven't used it either. I tested having a `steps.path.outputs.space_separated_string` which was set to `"path1 path2 path3"`, but `"${{ join(steps.paths.outputs.space_separated_string, '\n') }}"` would leave it unchanged. I only had much luck with getting a string separated with `\n` when:
-  using outputs from multiples steps with `*` - `${{ join(steps.*.outputs.list, '\n') }}`
- using fromJson with a formatted json string - `"${{ join(fromJson(steps.paths.outputs.list), '\n') }}"` where `steps.paths.outputs.list` was `'["path1", "path2", "path3"]'`.

But the docs say the input can be string - so maybe something was wrong with my test setup..

jo-basevi:
@CodeGat Do you know much on how to use `join()?

CodeGat:
I haven't had any luck with it being a string, either. I confirmed @jo-basevi s tests, as well. I suspect that the docs are wrong, or mean "JSON array as string", rather than space-separated string. I think the solutions are either doing it as @jo-basevi mentioned above, or doing it in bash via `tr ' ' '\n' <<< "${{ steps.build-and-upload.outputs.paths }}"`...

atteggiani:
@jo-basevi @CodeGat 
Thank you for your tests. 

Alright, I removed the commit where I updated that part of the code.

jo-basevi (approved):
Thanks @atteggiani, looks good to me

atteggiani:
Thanks @jo-basevi, @blimlim and @CodeGat.
Merged.

PR #211, opened 2025-12-08
Update versioning management

### LINKED ISSUE #209: Add option to print out version
## Overview
Currently there is no `--version/-V` option that prints out the version of `um2nc` installed. 
The only way to get current version is to run something like:
```
python -c "from um2nc import _version; print(_version.get_versions()['version'])"
```

## Possible implementation
The quickest implementation would probably be adding the `--version/-V` option to the [parse_args](https://github.com/ACCESS-NRI/um2nc-standalone/blob/4616f430e9254f2c1c1c63ff58d8312848e2a3c2/src/um2nc/um2netcdf.py#L1105) function, which prints `um2nc.__version__`:
```
parser.add_argument(
    "--version",
    action="version",
    version=um2nc.__version__,
)
```

### LINKED ISSUE #210: Switch to `setuptools_scm` for version management instead of `versioneer`
## Overview
Versioneer is not well-maintained anymore and its compatibility stops to Python 3.11. 

To be able to support Python versions >= 3.12, the suggestion is to move to the built-in [setuptools-scm](https://github.com/pypa/setuptools-scm) as a way to manage versioning directly from `git` tags.

atteggiani:
Fixes #209 
Fixes #210
 
- [x] Removed `versioneer` dependency and switched in `setuptools-scm`
- [x] Removed `coecms` from the channel as it's not needed anymore. `mule` is now available in the `accessnri` channel.
~- [x] Updated conda recipe, CI and CD workflows~

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,3 +50,10 @@ pythonpath = "src"
 [tool.setuptools_scm]
 version_scheme = "no-guess-dev"
 local_scheme = "node-and-date"
+
+[tool.coverage.run]
+branch = true
+
+[tool.coverage.html]
+title = "Coverage report: um2nc-standalone"
+directory = "coverage_html"
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,9 +1,0 @@
-[run]
-branch = True
-
-# skip coverage testing for external source files
-omit = um2nc/_version.py
-
-[html]
-title = Coverage report: um2nc-standalone
-directory = coverage_html
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,5 +1,0 @@
-[MAIN]
-
-# Files or directories to be skipped. They should be base names, not
-# paths.
-ignore=_version.py


pyproject.toml
```diff
@@ -38,21 +36,17 @@ Repository = "https://github.com/ACCESS-NRI/um2nc-standalone"
...
-parentdir_prefix = "um2nc-"
-
 [tool.pytest.ini_options]
-pythonpath = "src"
\ No newline at end of file
+pythonpath = "src"
+
+[tool.setuptools_scm]
+version_scheme = "no-guess-dev"
+local_scheme = "no-local-version"
```
blimlim: I believe the old versioning system added a git hash for the local version, is there a preference either way?

pyproject.toml
```diff
@@ -38,21 +36,17 @@ Repository = "https://github.com/ACCESS-NRI/um2nc-standalone"
 [build-system]
 build-backend = "setuptools.build_meta"
 requires = [
-    "setuptools>64", 
-    "versioneer[toml]"
+    "setuptools>=80", 
+    "setuptools-scm[simple]>=8",
```
blimlim: Just wanted to double check about including `simple` option. 

In the [simplified activation](https://setuptools-scm.readthedocs.io/en/latest/usage/#simplified-activation-new) and [explicit configuration](https://setuptools-scm.readthedocs.io/en/latest/usage/#explicit-configuration-full-control) sections of the docs, it seems like they remove the `simple` flag when adding customisation via a `tool.setuptools_scm` section for customisation.

It's not 100% clear to me whether the `simple` flag should or shouldn't be removed when including a `tool.setuptools_scm` section. That said, the current options in the PR are definitely working and it might be ok as is!

Did you have any thoughts on this?

blimlim (commented):
Thanks @atteggiani, I think this is looking really good. I've just added a couple of quick questions!

Also, can the following references to `_version.py` be safely removed: https://github.com/ACCESS-NRI/um2nc-standalone/blob/4616f430e9254f2c1c1c63ff58d8312848e2a3c2/.coveragerc#L5


https://github.com/ACCESS-NRI/um2nc-standalone/blob/4ccc0b6a051d18f69eff0f3e5d0bc0896f154fe2/.pylintrc#L5

atteggiani:
> Thanks @atteggiani, I think this is looking really good. I've just added a couple of quick questions!
> 
> Also, can the following references to `_version.py` be safely removed:
> 
> https://github.com/ACCESS-NRI/um2nc-standalone/blob/4616f430e9254f2c1c1c63ff58d8312848e2a3c2/.coveragerc#L5
> 
> https://github.com/ACCESS-NRI/um2nc-standalone/blob/4ccc0b6a051d18f69eff0f3e5d0bc0896f154fe2/.pylintrc#L5

Yes they can definitely be removed. 

Actually, I think the whole `.pylintrc` file could be removed as we're not using pylint. We might set up pre-commit hooks for linting, but I would use Ruff instead of Pylint anyway.
Also the `.coveragerc` file can be removed. If coverage configuration is needed, I would prefer [including it in the pyproject.toml](https://coverage.readthedocs.io/en/latest/config.html#toml-syntax) instead.

pyproject.toml
atteggiani: I had the same doubt when reading documentation (I don't think it's clear enough on this).
However, I remember getting an error without the `[simple]` option (can't remember exactly the error though). 
And, as you said, everything seems to be working as expected, so I presume the `[simple]` option can be used together with `tool.setuptools_scm` customisations.

pyproject.toml
blimlim: Can do! I've added this change in [e803c68](https://github.com/ACCESS-NRI/um2nc-standalone/pull/211/commits/e803c68cab7462437e0cb20de3fdbcca64f68e51), where I've also tried removing `setuptools` from the `.conda/*.yml` dependencies

PR #214, opened 2025-12-11
Add common driver functions file

### LINKED ISSUE #213: Separate common driver functions
To prepare for a separate ESM1.6 conversion driver, it will be helpful to pull common functions out of the ESM1.5 conversion driver. 

I imagine that the general steps performed by the the drivers would be:
1. Find UM output files to convert
2. Generate filepaths for the output netcdfs
3. Perform the conversion
4. Report on successes/failures

and the drivers would differ in the details of how/where they look for input files, how they combine input files together, and how they organise/name the output files. This leaves a lot of room for common functionality accross the drivers. In particular things like finding files via regex matching: https://github.com/ACCESS-NRI/um2nc-standalone/blob/4616f430e9254f2c1c1c63ff58d8312848e2a3c2/src/um2nc/conversion_driver_esm1p5.py#L103

Extracting information from fields files: https://github.com/ACCESS-NRI/um2nc-standalone/blob/4616f430e9254f2c1c1c63ff58d8312848e2a3c2/src/um2nc/conversion_driver_esm1p5.py#L166

Converting a generic list of fields files: https://github.com/ACCESS-NRI/um2nc-standalone/blob/4616f430e9254f2c1c1c63ff58d8312848e2a3c2/src/um2nc/conversion_driver_esm1p5.py#L185

Reporting on conversion status: https://github.com/ACCESS-NRI/um2nc-standalone/blob/4616f430e9254f2c1c1c63ff58d8312848e2a3c2/src/um2nc/conversion_driver_esm1p5.py#L217

blimlim:
Closes #213 

This PR reorganises the ESM1.5 conversion driver, separating out functions which will be useful for other model drivers (ESM1.6, CM3), and putting them in a new `drivers/common.py` file. This PR only moves and renames functions with minimal functionality changes. I'll add details of each of the moved functions in separate comments.

For an example of how these changes would be used in other drivers, see the prototype branch https://github.com/ACCESS-NRI/um2nc-standalone/tree/esm1p6_driver

For the reviewers: The diff looks pretty big, and so it might be easier to click through the commits individually where it's clearer that functions are just being removed around.

### DIFF
--- a/.conda/meta.yaml
+++ b/.conda/meta.yaml
@@ -30,6 +30,7 @@ requirements:
 test:
     imports:
         - um2nc
+        - um2nc.drivers
     commands:
     {% for name, script in project.get('scripts').items() %}
         - {{ name }} --help


src/um2nc/drivers/common.py
```diff
@@ -0,0 +1,237 @@
...
+import traceback
+import warnings
+
+import mule
+
+from pathlib import Path
+from um2nc import um2netcdf
+
+
+def get_fields_file_pattern(run_id: str):
```
blimlim: UM pp outputs will have the same naming scheme in ESM1.5 and 1.6. This function which creates a regex pattern for matching filenames can be used for both models

src/um2nc/drivers/common.py
```diff
@@ -0,0 +1,237 @@
...
+        raise ValueError(
+            f"Received run_id = {run_id} with length {len(run_id)}. run_id must be length 5"
+        )
+
+    fields_file_name_pattern = rf"^{run_id}a.p[a-z0-9]+$"
+
+    return fields_file_name_pattern
+
+
+def find_matching_files(dir_contents, fields_file_name_pattern):
```
blimlim: This is a fairly generic function which subsets a list of files based on a regex pattern. This will be used across models.

src/um2nc/drivers/common.py
```diff
@@ -0,0 +1,237 @@
...
+
+def _resolve_path(path):
+    """
+    Resolve path for use in comparison. Ensure that symlinks, relative paths,
+    and home directories are expanded.
+    """
+    return os.path.realpath(os.path.expanduser(path))
+
+
+def filter_name_collisions(input_output_pairs):
```
blimlim: This one is a safety check to make sure multiple input files aren't being sent to the same output location, and will be useful across models

atteggiani (approved):
Thank you @blimlim for working on this. 

It looks good! Grouping common functions will simplify future development.

The conda build action seems to be failing during the [driver script testing](https://github.com/ACCESS-NRI/um2nc-standalone/actions/runs/20117380514/job/57729947416?pr=214#step:5:432), even though the same test succeeded in the testing CI.
Do you understand why? It seems to complain about the `um2nc.drivers` module, but it exists so I"m not sure about that.

PR #216, opened 2026-01-13
Error and warning behaviour for missing heaviside variables

### LINKED ISSUE #204: Promote missing heavyside warning to error during model runs
When the required heavyside function (stash code 30301 or 30304) is missing, um2nc drops the dependent pressure level variables from the output. 

It only produces a warning that this is happening when the optional verbose flag `-v` is used. This makes it difficult to notice if or why output is missing, see [here](https://github.com/ACCESS-NRI/access-esm1.6-configs/issues/113#issuecomment-2870245114) and [here](https://forum.access-hive.org.au/t/convert-um-output-to-netcdf-um2nc-not-working/4557/3). 

I think it makes sense for the warning to always be produced regardless of the verbose level. @MartinDix also suggests promoting it to an error when called from a model run, as this will provide early feedback during model simulations.

blimlim:
Closes #204 
### Background:
Pressure level data that falls below the surface for some fraction of the time requires special handling. A heaviside mask output by the UM needs to be applied to ensure these points have valid data. 

When the required heaviside is missing, the current behaviour is to drop the pressure level variables, producing a warning only when the `verbose` flag is true. Silently dropping the variables isn't ideal, as it can take [some time](https://github.com/ACCESS-NRI/access-esm1.6-configs/issues/364) to notice the missing variables in the output.

Following discussion, it was suggested that 1. warnings always be produced when variables are dropped, and 2. warnings be promoted to an error when running through a model driver – this will hopefully make it faster to spot issues with the requested variables in the model configurations. 

### Implementation notes:
I've converted the `--no-mask` boolean option into a string option with three choices:
- `drop-missing`: The default option. Produce a warning and drop pressure level variables with missing associated heavisides
- `error-missing`: Produce an error if any required heavisides are missing. To be used with model drivers
- `no-mask`: Skip the masking step. Same as the current `--no-mask` flag.

I thought this was the cleanest way to implement it as the choices are mutually exclusive, and it's simple to pass a choice from a model driver to the `um2nc.process` function by using the existing arguments. See [here](https://github.com/ACCESS-NRI/um2nc-standalone/commit/5dec665215ccde391721d70bf786ff9c563c4d8b) for a much messier attempt which didn't make use of the existing arguments.

### Testing
The integration tests currently fail. I believe this is due to the earlier changes [here](https://github.com/ACCESS-NRI/um2nc-standalone/commit/b88551d4d92923e51833e5c82e446efc16e627b7) which changed the variable names while the integration test reference data hasn't been updated to match. I'll make a seperate issue to update the integration test data. Output from this PR and [this commit](https://github.com/ACCESS-NRI/um2nc-standalone/commit/b88551d4d92923e51833e5c82e446efc16e627b7) are identical.

Running `esm1p5_convert_nc` on model output which omitted the heaviside variables produced the following error:
```
RuntimeError: Heaviside_UV field (code 30301) required for masking fields [30201, 30202, 30204, 30205, 30206, 30207, 30208] is missing.See 'mask_option' argument for pressure level masking options.
```

### DIFF
--- a/src/um2nc/um2netcdf.py
+++ b/src/um2nc/um2netcdf.py
@@ -78,9 +78,9 @@ TIME = "time"
 DEFAULT_FILL_VAL_FLOAT = 1.0e20
 
 # Options for pressure level masking
-NO_MASK = "no_mask"
-ERROR_MISSING = "error_missing"
-DROP_MISSING = "drop_missing"
+NO_MASK = "no-mask"
+ERROR_MISSING = "error-missing"
+DROP_MISSING = "drop-missing"
 
 
 class PostProcessingError(Exception):
@@ -795,7 +795,7 @@ def is_heaviside_t(item_code):
 
 def non_masking_cubes(cubes, heaviside_uv, heaviside_t, mask_option):
     """
-    Yields cubes that:
+    Returns cubes that:
     * do not require pressure level masking
     * require pressure level masking & the relevant masking cube exists
 
--- a/test/test_um2netcdf.py
+++ b/test/test_um2netcdf.py
@@ -255,7 +255,7 @@ def test_process_cubes_no_heaviside_error(ta_plev_cube, precipitation_flux_cube,
                                                geo_potential_cube, mule_vars, std_args):
     """
     Attempt end-to-end process_cubes() test. Check that an error is raised when the
-    mask_option is set to error_missing and the relevant heaviside variables are missing.
+    mask_option is set to ERROR_MISSING and the relevant heaviside variables are missing.
     """
 
     # include cubes requiring both heaviside uv & t cubes to filter
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ Points on a pressure level grid may fall below ground-level in some fields of th
 By default, variables on pressure level grids that fall below-ground level will be masked with the appropriate Heaviside variable found in the input file. If the Heaviside variable cannot be found, these variables will be omitted from the output. This behaviour can be controlled by the following options:
 
 * `--hcrit HCRIT` Minimum fraction of the time spent above ground-level for a pressure grid data point to be considered valid.  Data points in pressure grid variables will be masked if they were above ground-level for less than the critical fraction `HCRIT` of the time. This option has no effect when used together with the `--nomask` option. Default `0.5`.
-* `--nomask` Don't mask variables on pressure level grids. When selected, unmasked pressure level variables will be written to the output file regardless of the presence of the Heaviside variable.
+* `--mask-option` Option for masking pressure level variables. Choose from `drop-missing` (drop pressure level variables which require masking if the required heaviside variables are missing), `error-missing` (produce an error if pressure level variables require masking but the heaviside variables are missing), `no-mask` (don't mask variables on pressure level grids. When selected, unmasked pressure level variables will be written to the output file regardless of the presence of the Heaviside variable). Default: `drop-missing`.
 
 
 _Metadata options:_
--- a/integration/INTEGRATION_README.md
+++ b/integration/INTEGRATION_README.md
@@ -60,7 +60,7 @@ fields file, and can be used for faster but less in-depth testing. It includes:
 For each of the above data choices, the following netCDF variants were produced:
 
 * `mask`: produced with the `--nohist` flag only.
-* `nomask`: produced with the `--nomask` and `--nohist` flags.
+* `nomask`: produced with the `--mask-option no-mask` and `--nohist` options.
 * `hist`: produced with no flags. These files will have a conversion datestamp
 in their history attribute.
 
--- a/integration/regression_tests.sh
+++ b/integration/regression_tests.sh
@@ -176,7 +176,7 @@ function diff_warn {
 # Test 1:
 # Execute nomask variant, pressure masking is turned OFF & all cubes are kept.
 run_um2nc    --nohist \
-             --nomask \
+             --mask-option no-mask \
              "$source_ff" \
              "$out_nomask_nc"
 


src/um2nc/um2netcdf.py
```diff
@@ -77,6 +77,11 @@
 
 DEFAULT_FILL_VAL_FLOAT = 1.0e20
 
+# Options for pressure level masking
+NO_MASK = "no-mask"
+ERROR_MISSING = "error-missing"
+DROP_MISSING = "drop-missing"
```
atteggiani: I think this logic would be more solid as an Enum:

import it:
```
from enum import Enum
```

and then
```suggestion
# Options for pressure level masking
class PressureLevelMaskingOption(Enum):
    NO_MASK = "no-mask"
    ERROR_MISSING = "error-missing"
    DROP_MISSING = "drop-missing"
```

I think some other logic in other files might change due to this

src/um2nc/um2netcdf.py
```diff
@@ -548,13 +553,13 @@ def process_cubes(cubes, mv, args):
...
 
-    do_masking = not args.nomask
+    do_masking = (args.mask_option != NO_MASK)
     heaviside_uv, heaviside_t = get_heaviside_cubes(cubes)
 
     if do_masking:
         # drop cubes which cannot be pressure masked if heaviside uv or t is missing
         # otherwise keep all cubes when masking is off
-        cubes = list(non_masking_cubes(cubes, heaviside_uv, heaviside_t, args.verbose))
+        cubes = non_masking_cubes(cubes, heaviside_uv, heaviside_t, args.mask_option)
```
atteggiani: I see the `non_masking_cubes` function has been changed to return a list already.
More comments on this in the related section.

src/um2nc/um2netcdf.py
```diff
@@ -799,22 +806,41 @@ def non_masking_cubes(cubes, heaviside_uv, heaviside_t, verbose: bool):
...
+        if mask_option == ERROR_MISSING:
+            msg += "See 'mask_option' argument for pressure level masking options."
+            raise RuntimeError(msg)
+        elif mask_option == DROP_MISSING:
+            msg += "These cubes will be dropped. See 'mask_option' argument for pressure level masking options."
+            warnings.warn(msg, category=RuntimeWarning)
+        else:
+            raise ValueError(f"Invalid masking option '{mask_option}'. Expected '{DROP_MISSING}' or '{ERROR_MISSING}'")
+
+    return keep_cubes
```
atteggiani: This changes the function from returning a generator to a list. 
This might become important when the number if the number of cubes is very large, because generators yield each cube as they are read, while list store all cubes in memory.
In general, this is a fundamental change that I don't see related to issue this PR tries to fix.

A similar logic to the proposed one can be achieved by keeping the function as a generator function.

src/um2nc/drivers/conversion_driver_esm1p5.py
```diff
@@ -20,6 +20,7 @@
 
 from pathlib import Path
 from um2nc.stashmasters import STASHmaster
+from um2nc.um2netcdf import ERROR_MISSING
```
atteggiani:
```suggestion
from um2nc.um2netcdf import PressureLevelMaskingOption
```

test/test_um2netcdf.py
```diff
@@ -192,7 +192,7 @@ def heaviside_t_cube(lat_standard_eg_coord, lon_standard_eg_coord):
 def std_args():
     # TODO: make args namedtuple?
     args = mock.Mock()
-    args.nomask = False  # perform masking if possible
+    args.mask_option = um2nc.DROP_MISSING  # perform masking if possible
```
atteggiani:
```
from um2nc.um2netcdf import PressureLevelMaskingOption
```
and then
```suggestion
    args.mask_option = PressureLevelMaskingOption.DROP_MISSING  # perform masking if possible
```

test/test_um2netcdf.py
```diff
@@ -223,6 +223,8 @@ def test_process_cubes_no_heaviside_drop_cubes(ta_plev_cube, precipitation_flux_
     # ensure both uv/t dependent cubes are dropped
     cubes = [ta_plev_cube, precipitation_flux_cube, geo_potential_cube]
 
+    std_args.mask_option = um2nc.DROP_MISSING
```
atteggiani:
```suggestion
    std_args.mask_option = PressureLevelMaskingOption.DROP_MISSING
```

test/test_um2netcdf.py
```diff
@@ -280,6 +308,9 @@ def test_process_mask_with_heaviside(ta_plev_cube, precipitation_flux_cube,
     cubes = [ta_plev_cube, precipitation_flux_cube, geo_potential_cube,
              heaviside_uv_cube, heaviside_t_cube]
 
+    # Ensure that the masking is applied
+    std_args.mask_option = um2nc.ERROR_MISSING
```
atteggiani:
```suggestion
    std_args.mask_option = PressureLevelMaskingOption.ERROR_MISSING
```

test/test_um2netcdf.py
```diff
@@ -314,7 +345,7 @@ def test_process_no_masking_keep_all_cubes(ta_plev_cube, precipitation_flux_cube
     ):
         m_time_coord.side_effect = mock_fix_time_no_time_dim
 
-        std_args.nomask = True
+        std_args.mask_option = um2nc.NO_MASK
```
atteggiani:
```suggestion
        std_args.mask_option = PressureLevelMaskingOption.NO_MASK
```

atteggiani (changes requested):
Thank you Spencer for working on this PR. 

I added some comments with some suggestions for the changed parts. Please refer to the specific ones for details.

Also, I see the `--no-mask` option was not replaced and it is still included within the `ARG_NAMES`. Was this done on purpose? How are the `--no-mask` and `--mask-option` options playing together? This sounds a bit confusing to me.

I then have a more general thought about this approach and the issue it is trying to address.

The main purpose of this PR is:
- changing a warning from appearing only in verbose mode (`-v` option) to always appearing 
- changing the behaviour for drivers script to treat warnings (only the warning mentioned above or all "internal" warnings???) as errors

To me these both sound like issues with the "logging" level. As a consequence, I am not sure a good approach is to change the functionality of flags and internal functions whose main logic is not related to logging, when what should really change is the logging logic. Having multiple different flags dealing with logging control (`--verbose` and `--mask-option`) might easily lead to confusion and makes the codebase more error-prone and more difficult to test.

um2nc doesn't use the `logging` module to control verbosity, which in my opinion would probably be the best solution to handle these type of problems. Also, I see warnings are not logged by default, but only in the "verbose" mode.

I think a solution to this issue that only modifies options and functions that concern "logging" could be:
- log warnings by default (not only with the `-v` option). This is often the case with many tools that have the default level set to WARNING so warnings are always logged. We could also implement a flag to forcefully prevent warnings from being logged (for example `--quiet/-q`).
- implement a flag (for example `--strict`) to convert all internal warnings (the ones raised directly by `um2nc`) to errors. This is the flag that would be used within drivers scripts.

With this solution, we could have the expected functionality: a warning is produced when variables are removed because the Heavyside function is missing (because warnings would be logged by default), and an error would be produced in the driver scripts (because we would run them using the `--strict` option). All this without touching other logic that doesn't involve logging. 

This solution can be implemented with the "custom" logging system that um2nc currently uses, but it might be more convoluted and hacky. A cleaner solution, even though it requires to change bigger parts of the codebase, is to update um2nc to be using "logging" for verbosity control.

blimlim:
Let's try the logging suggestion!

PR #217, opened 2026-01-14
Use logging to module to control output and warnings

### LINKED ISSUE #204: Promote missing heavyside warning to error during model runs

blimlim:
Closes #204 and follows from discussion in #216.

This PR adds in the logging module to simplify the control of verbosity, output and warnings. As suggested by @atteggiani in #216, this also provides a cleaner method for promoting the heaviside warnings to errors when running from a model driver. 

I've attempted to keep the same verbosity behaviour with these changes, i.e. so that the same messages are reported for the same verbosity settings.

I'll add some more detailed comments and questions in the comments.

### DIFF
--- a/src/um2nc/drivers/conversion_driver_esm1p5.py
+++ b/src/um2nc/drivers/conversion_driver_esm1p5.py
@@ -14,7 +14,6 @@ https://github.com/ACCESS-NRI/access-cm2-drivers/blob/main/src/run_um2netcdf.py
 import os
 import collections
 import f90nml
-import logging
 import warnings
 import argparse
 import errno
@@ -24,7 +23,6 @@ from um2nc.stashmasters import STASHmaster
 from um2nc.um2netcdf import setup_logging
 
 from um2nc.drivers.common import find_matching_files, get_ff_date
-from um2nc.drivers.common import format_successes, format_failures
 from um2nc.drivers.common import filter_name_collisions, safe_removal
 from um2nc.drivers.common import get_fields_file_pattern
 from um2nc.drivers.common import convert_fields_file_list
@@ -199,12 +197,6 @@ def main():
     setup_logging(ARG_VALS.verbose, ARG_VALS.quiet, ARG_VALS.strict)
 
     successes, failures = convert_esm1p5_output_dir(args.current_output_dir)
-
-    # Report results to user
-    for success_message in format_successes(successes):
-        logging.info(success_message)
-    for failure_report in format_failures(failures):
-        warnings.warn(failure_report, category=RuntimeWarning)
 
     if args.delete_ff:
         # Remove files that appear only as successful conversions
--- a/src/um2nc/drivers/common.py
+++ b/src/um2nc/drivers/common.py
@@ -4,6 +4,7 @@ Common functions used across model conversion drivers
 """
 
 import collections
+import logging
 import os
 import re
 import warnings
@@ -96,10 +97,9 @@ def convert_fields_file_list(input_output_paths, process_args):
 
     Returns
     -------
-    succeeded: list of tuples of (input, output) filepaths for successful
-               conversions.
-    failed: list of tuples of form (input path, exception) for files which
-            failed to convert due to an allowed exception.
+    succeeded: list of input filepaths which were successfully converted.
+    failed: list of input filepaths which could not be converted due to an
+            allowed exception.
     """
 
     succeeded = []
@@ -108,55 +108,21 @@ def convert_fields_file_list(input_output_paths, process_args):
     for ff_path, nc_path in input_output_paths:
         try:
             um2netcdf.process(ff_path, nc_path, process_args)
-            succeeded.append((ff_path, nc_path))
 
         except um2netcdf.UnsupportedTimeSeriesError as exc:
-            failed.append((ff_path, exc))
+            failed.append(ff_path)
+            warnings.warn(
+                f"Failed to convert {ff_path} with error:\n{repr(exc)}",
+                category=RuntimeWarning
+            )
+
+        else:
+            succeeded.append(ff_path)
+            logging.info(f"Successfully converted {ff_path} to {nc_path}")
 
         # Any unexpected errors will be raised
 
     return succeeded, failed
-
-
-def format_successes(succeeded):
-    """
-    Format reports of successful conversions to be shared with user.
-
-    Parameters
-    ----------
-    succeeded: list of (input, output) tuples of filepaths for successful
-               conversions.
-
-    Yields
-    -------
-    success_report: formatted report of successful conversion.
-    """
-
-    for input_path, output_path in succeeded:
-        success_report = f"Successfully converted {input_path} to {output_path}"
-        yield success_report
-
-
-def format_failures(failed):
-    """
-    Format reports of conversions which failed with permitted exceptions.
-
-    Parameters
-    ----------
-    failed: list of tuples of form (filepath, exception) for files which failed
-            to convert due to an allowable exception.
-    Yields
-    -------
-    failure_report: Formatted reports of failed conversions.
-    """
-
-    for fields_file_path, exception in failed:
-        failure_report = (
-                f"Failed to convert {fields_file_path}. Final reported error: \n"
-                f"{repr(exception)}"
-            )
-
-        yield failure_report
 
 
 def _resolve_path(path):
@@ -206,16 +172,15 @@ def safe_removal(succeeded, failed):
 
     Parameters
     ----------
-    succeeded: List of (input_file, output_file) tuples of filepaths from
-        successful conversions.
-    failed: List of (input_file, Exception) tuples from failed conversions.
+    succeeded: List of input filepaths from successful conversions.
+    failed: List of input filepaths from failed conversions.
 
     Returns
     -------
     successful_only: set of input filepaths which appear in succeeded but
         not failed.
     """
-    succeeded_inputs = {succeed_path for succeed_path, _ in succeeded}
-    failed_inputs = {fail_path for fail_path, _ in failed}
+    succeeded_inputs = set(succeeded)
+    failed_inputs = set(failed)
 
     return succeeded_inputs - failed_inputs
--- a/test/drivers/test_drivers_common.py
+++ b/test/drivers/test_drivers_common.py
@@ -1,5 +1,6 @@
 import um2nc.drivers.common as drivers_common
 
+import logging
 import pytest
 from pathlib import Path
 import unittest.mock as mock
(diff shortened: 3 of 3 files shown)


src/um2nc/drivers/conversion_driver_esm1p5.py
```diff
@@ -184,14 +186,6 @@ def parse_args():
...
         type=str
     )
-    parser.add_argument("--quiet", "-q", action="store_true",
-                        help=(
-                            "Report only final exception type and message for "
-                            "any expected `um2nc` exceptions raised during "
-                            "conversion. If absent, full stack traces "
-                            "are reported"
-                        )
-                        )
```
blimlim: The conversion driver had an additional `--quiet` option. This controlled the error reporting when the allowable [UnsupportedTimeseriesError](https://github.com/ACCESS-NRI/um2nc-standalone/blob/23e407029255d7676240fc009dcf1944ced412a0/src/um2nc/um2netcdf.py#L312) came up.

When `--quiet` was active, the conversion driver only reported the final error type and message. Otherwise (by default), it reported the full traceback.

I figured it's better to handle this with the same controls as the main um2nc program, and so have tried reimplement it using the introduced logging settings. I.e: with `verbose==False` we report the final exception type and message, and `verbose=True` we report the full traceback.

src/um2nc/drivers/conversion_driver_esm1p5.py
```diff
@@ -201,13 +195,16 @@ def parse_args():
 
 def main():
     args = parse_args()
+    setup_logging(ARG_VALS.verbose, ARG_VALS.quiet, ARG_VALS.strict)
```
blimlim: Set `strict==True` for the model driver

test/test_um2netcdf.py
```diff
@@ -506,25 +507,20 @@ def test_fix_standard_name_update_y_wind():
...
+def test_fix_standard_name_with_mismatch(x_wind_cube, caplog):
     # ensure mismatching standard names between cube & um uses the um std name
+    # and that the the mismatch is logged
     standard_name = "fake"
     assert x_wind_cube.standard_name != standard_name
-    um2nc.fix_standard_name(x_wind_cube, standard_name, verbose=False)
-    assert x_wind_cube.standard_name == standard_name
-
 
-def test_fix_standard_name_with_mismatch_warn(x_wind_cube):
```
blimlim: This test no longer needs to be seperate from `test_fix_standard_name_with_mismatch` above

src/um2nc/um2netcdf.py
```diff
@@ -93,6 +94,38 @@ class UnsupportedTimeSeriesError(PostProcessingError):
...
+    """
+    pass
+
+
+def setup_logging(verbose: bool, quiet: bool, strict: bool):
+    """Setup logging levels"""
+    level = logging.WARNING
+
+    if verbose and quiet:
+        raise RuntimeError("'verbose' and 'quiet' options cannot be simultaneously active")
```
atteggiani:
```suggestion
```

This might be better enforced during argument parsing by setting `--verbose` and `--quiet` options as [mutually exclusive](https://docs.python.org/3/library/argparse.html#mutual-exclusion)

src/um2nc/um2netcdf.py
```diff
@@ -93,6 +94,38 @@ class UnsupportedTimeSeriesError(PostProcessingError):
...
+
+    if verbose:
+        level = logging.INFO
+    if quiet:
```
atteggiani:
```suggestion
    elif quiet:
```
Since the options are mutually exclusive

src/um2nc/drivers/conversion_driver_esm1p5.py
blimlim: Admittedly the `verbose=True` option wouldn't currently be available when running from a model driver, as it's being hardcoded here https://github.com/ACCESS-NRI/um2nc-standalone/blob/04fdf47d6505799559d258c8e92416c42d0a7822/src/um2nc/drivers/conversion_driver_esm1p5.py#L43

I'm hoping when we swap to running the conversion drivers as subcommands that we'll also be able to make all the um2nc options available as arguments to the driver.

src/um2nc/um2netcdf.py
```diff
@@ -900,15 +931,14 @@ def fix_standard_name(cube, um_standard_name, verbose: bool):
...
-                )
-                warnings.warn(msg)
+            # TODO: show combined stash code instead?
+            # TODO: Should this be a warning?
+            msg = (
+                f"Standard name mismatch section={stash_code.section}"
+                f" item={stash_code.item} standard_name={cube.standard_name}"
+                f" UM var name={um_standard_name}"
+            )
+            logging.info(msg)
```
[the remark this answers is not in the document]
atteggiani: This is a mismatch, are we sure the message should be info and not an actual warning? 
Info in general is used more for informational messages about what is going on. This looks more like a warning message to me.

src/um2nc/um2netcdf.py
```diff
@@ -934,14 +964,14 @@ def fix_long_name(cube, um_long_name):
...
+def fix_units(cube, um_var_units):
     if cube.units and um_var_units:
         # Simple testing c.units == um_var_units doesn't catch format differences because
         # the Unit type works around them. repr is also unreliable
         if f"{cube.units}" != um_var_units:  # TODO: does str(cube.units) work?
-            if verbose:
-                msg = f"Units mismatch {cube.item_code} {cube.units} {um_var_units}"
-                warnings.warn(msg)
+            msg = f"Units mismatch {cube.item_code} {cube.units} {um_var_units}"
+            logging.info(msg)
```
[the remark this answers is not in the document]
atteggiani: Same as above, this is a mismatch. To me it sounds more like a warning message.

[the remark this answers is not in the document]
atteggiani: Same as above, this is a mismatch. To me it sounds more like a warning message.

src/um2nc/um2netcdf.py
```diff
@@ -1083,10 +1112,10 @@ def fix_time_coord(cube, verbose):
...
                 neworder.insert(0, tdim)
 
-                if verbose > 1:
-                    # TODO: fix I/O, is this better as a warning?
-                    print("Incorrect dimension order", cube)
-                    print("Transpose to", neworder)
+                logging.info(
+                    f"Incorrect dimension order on {cube.item_code}."
+                    f"Transposing to {neworder}"
+                )
```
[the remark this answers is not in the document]
atteggiani: I don't think a double verbosity is needed. I would keep this as info.

src/um2nc/um2netcdf.py
```diff
@@ -93,6 +94,38 @@ class UnsupportedTimeSeriesError(PostProcessingError):
...
+    )
+
+    # Route warnings → logging
+    logging.captureWarnings(True)
+
+    # Apply strict option
+    if strict:
+        warnings.filterwarnings("error", category=StrictWarning)
+
+
```
atteggiani: Would this whole logging setup be better suited in its own module? 

We could have a `um2nc/logging.py` file where this piece of code resides and we can import `from um2nc.logging import ...`

atteggiani (changes requested):
Thank you @blimlim for working on this PR and for incorporating logging into um2nc.
I think these updates make the logging structure much more solid and simplifies development, along with introducing good separation of concerns.

I left some comments and suggestions. Please refer to the specific messages for details.


One more general suggestion for the logging setup:

I see you added the `setup_logging` function in both um2netcdf and driver script, and call the logging messages (for example info) using `logging.info`.

This is fine as long as one between the driver script and um2netcdf code runs (as it should be the case right now).
However, with the future idea of restructuring driver scripts into q subcommand and in general as a good practice, it might be better to create a logger only once with its setup (using the setup_logging function) and then only import/use that logger everywhere else. This avoids setting up logging multiple times by mistake and getting confusing logging outputs.
An example could be adding after the logging_setup something like
```py
logger=logging.getLogger('um2nc')
```
and then in um2nc and driver scripts import it from `um2nc.logging` (see related comment about this within the review comments) and use it as:
```py
from um2nc.logging import logger
...
logger.info(...)
```

src/um2nc/um2netcdf.py
blimlim: Thanks, yeah I think that makes sense!

src/um2nc/um2netcdf.py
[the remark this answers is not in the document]
atteggiani: I think this would be better handled as a warning. 
Also, I'm not sure I would understand the message. Do you think we could make it clearer?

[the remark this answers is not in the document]
blimlim: I've tried making the message a bit clearer in 1b8bab8 and swapped this over to a warning

atteggiani (changes requested):
Thank you @spencer for addressing the suggestions. 

This looks pretty good!

I have another general suggestion to further simplify the logging for successes and failures. 
Since we are only capturing a specific error failure (TimeSeriesException) we decided that an extensive traceback was not needed so we only output a warning with it. 

For this reason, there should be no need for the format_successes/format_failures functions. 

I think it's much simpler to call the logging logic (`logging.info(<success_msg>)` for successes and `warnings.warn(<failure_msg>)` for failures) directly within the `convert_fields_file_list` function:

https://github.com/ACCESS-NRI/um2nc-standalone/blob/1b8bab8207d60724761a5e8fb9f9c5026d3957a4/src/um2nc/drivers/common.py#L87-L118

```diff
for ff_path, nc_path in input_output_paths: 
         try: 
             um2netcdf.process(ff_path, nc_path, process_args)
-            succeeded.append((ff_path, nc_path)) 
  
         except um2netcdf.UnsupportedTimeSeriesError as exc: 
             failed.append((ff_path, exc))
+            warnings.warn(f"Failed to convert {ff_path} with error:\n{repr(exc)}")
+        else:
+            succeeded.append((ff_path, nc_path))
+            logging.log(f"Successfully converted {ff_path} to {nc_path}")
```

**Note1**: I added the `else` clause just to make it clearer and to separate the error catch that can happen only during `um2netcdf.process(...)`

**Note2**: Using f-strings within logging messages is not advised due to performance reasons, but in this case f-strings are much more readable and the performance loss is highly negligible.

This will simplify the code and will also make testing easier: less functions to test and you can test logging/warnings directly when testing the `convert_fields_file_list` function for successes/failures cases and different options (verbose, quiet, etc.).

blimlim:
Thanks @atteggiani, I think that sounds like a good idea! I've added this change in 9fd5c0197d48d153303e375cab054622f68e48ed

With the `format_successes` and `format_failures` functions removed, I believe it's also unnecessary for the output paths and exceptions to be returned by `convert_fields_file_list`. These were only being used by the `format_` functions, and so I've changed `convert_fields_file_list` to also only return lists of input filepaths.

PR #229, opened 2026-03-04
Prototype ModelDriver class and esm1.6 driver

blimlim:
Closes #216. I've added this PR to start some discussion around the proposed implementation, and so will start it off as a draft.

## Background

So far, ESM1.6 simulations have been using the ESM1.5 conversion driver. However, we now wish to change the output file names and locations when running the conversion for ESM1.6, and hence need to add a separate driver. Previous work on the code organisation (#214) and command line interface (#221) laid the groundwork for this change.

However, the way that the ESM1.5 driver is constructed as a large function containing many steps, does not lend itself to adding a similar ESM1.6 driver with a few small differences:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/11306b42cf3847479b34e6eb38826dd5a94bbd2b/src/um2nc/drivers/esm1p5.py#L100-L161

The only approach I could think of which keeps the current structure would be to copy and paste the esm1.5 driver function, into a new function for esm1.6. This felt like it would be a bad idea, and following discussion with @atteggiani, we decided to investigate implementing a class based approach for the drivers.

## Proposed implementation

Based on the ESM1.5 and (embedded in rose/cylc suites) CM2, and CM3 conversion drivers, model conversion drivers seem to have the following general steps in common:
1. Find the directory containing model outputs to convert
2. Find/setup the directory to write netCDF files to
3. Find the fields files in the input directory to be converted
4. Set output netCDF filenames/paths for each input file
5. Run the conversion

Based on these common steps, I've added a generic `ModelDriver` class, with empty methods for steps 1-4. It then combines them into a conversion method for step 5. 
https://github.com/ACCESS-NRI/um2nc-standalone/blob/c3c68044ba78c2d4e3f33d26dce6d7a790800b90/src/um2nc/drivers/common.py#L18-L22

Specific model drivers are then created as subclasses of the `ModelDriver` class, and they can fill in the model-specific details for the methods in steps 1-4 (e.g. models may have different output directory structures, filenames etc).

This allows us to define a ESM1.5 driver, and an ESM1.6 driver as a subclass, with only the relevant details changed: https://github.com/ACCESS-NRI/um2nc-standalone/blob/c3c68044ba78c2d4e3f33d26dce6d7a790800b90/src/um2nc/drivers/esm1p5.py#L27-L39

https://g.ithub.com/ACCESS-NRI/um2nc-standalone/blob/c3c68044ba78c2d4e3f33d26dce6d7a790800b90/src/um2nc/drivers/esm1p6.py#L11-L22

This structure is largely based on [payu's Model class](https://github.com/payu-org/payu/blob/master/payu/models/model.py).

A potential limitation, is that I've assumed conversion drivers will loop through a sequence of input files, and will convert each into a separate netCDF. This wouldn't hold if in the future we want a driver to have more complicated behaviour such as combining monthly data into year length files, or splitting multi-variable files into single variable netCDFs.

@atteggiani any comments or ideas you have about the proposed changes are very welcome! @aidanheerdegen, I've pinged you just to check that the [ESM1.6 frequency names](https://github.com/ACCESS-NRI/um2nc-standalone/blob/c3c68044ba78c2d4e3f33d26dce6d7a790800b90/src/um2nc/drivers/esm1p6.py#L13-L19) which are added to the filenames are as required, however any other comments are welcome too.


aidanheerdegen:
This looks like a good design @blimlim.

As far as the need to do this goes, you could effectively deprecate the `1p5` driver, just rename it to `1p6` and put in the changes. Anyone who needs old behaviour has to use old versions. This only works if `um2nc` is separately deployed from the `payu` environment, which it isn't ... so is maybe not that useful a suggestion.

src/um2nc/drivers/esm1p5.py
```diff
@@ -16,57 +15,76 @@
...
+        Find atmosphere fields files for conversion in a given model output directory.
+        """
+        return find_esm1pX_fields_files(input_dir)
+
+    def get_output_dir(self, parent_dir):
+        """
+        Given a path to an experiment parent directory, set up a directory for writing
+        netCDF outputs and return its path.
+        """
+        nc_write_dir = parent_dir / "netCDF"
```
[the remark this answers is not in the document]
aidanheerdegen: Wrong driver.

src/um2nc/drivers/esm1p6.py
```diff
@@ -0,0 +1,28 @@
...
+
+class Esm1p6Driver(Esm1p5Driver):
+
+    # Output file suffix for each type of unit.
+    UNIT_SUFFIXES = {
+        "a": "1monthly",
+        "e": "1daily",
+        "j": "6hourly",
+        "i": "3hourly",
+        "c": "1hourly"
```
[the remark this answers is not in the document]
blimlim: Perfect thanks for clarifying those, I've added this b999a24

atteggiani:
Thank you @blimlim for starting to work on this!

I like the overall class-based design. 

I quickly had a look at the code and added a few comments. Some are questions to try and better understand your design idea.

On a more general level, I also have the following questions:

1. > ... (embedded in rose/cylc suites) CM2, and CM3 conversion drivers, ...
   
   What do you mean by "embedded in rose/cylc suites"? There is no currently no `um2nc driver cm2` command, so how is CM2 driver currently embedded in the rose/cylc suite?

2. When you mention "inputs" and "outputs" files, directories, etc., are they referring to the inputs and outputs of `um2nc`, or the models? 
   It was not super clear at times and I got a bit confused.
   I know that for model drivers `um2nc` conversion inputs are model outputs, but ultimately, we care about `um2nc` inputs and outputs, so for the implementation I don't think we care too much if they are model outputs, random files, or anything else. Therefore, I would suggest referring mostly to "inputs" and "outputs" files relative to `um2nc`, and using "model outputs" very rarely (Maybe only in a model driver general docstring) to avoid confusion.
   For example, instead of using `output_files` (referring to the model output files) we could use `input_files` or `conversion_target_files` to refer to the um2nc conversion target (input) files.

3. > 1. Find the directory containing model outputs to convert
    > 2. Find/setup the directory to write netCDF files to
    > 3. Find the fields files in the input directory to be converted
    > 4. Set output netCDF filenames/paths for each input file
    > 5. Run the conversion
    
    A model driver command, ultimately, is a set of multiple `um2nc` conversions for multiple files, to which there might also be pre/post-processing associated. For this reason, I would consider each `um2nc` file conversion independent, and for this reason they could also be easily parallelised in the future. 
    Therefore, I would prioritise the following steps:
    1. Get the list of targets (files) to be converted by `um2nc` coversion.
    2. Get the output directory where to write the converted files (this is basically your point 2)
    3. Perform the conversion (with any pre/post-processing steps)
    
    I think finding the directory containing the model outputs (which contain some of the `um2nc` conversion targets above) is a step included within my number 1 above. There might be drivers in which these targets are found in multiple directories, therefore I wouldn't list that as a "core" step for a driver conversion. Also related to [this comment](https://github.com/ACCESS-NRI/um2nc-standalone/pull/229#discussion_r2892786166).
    Also, your number 4 is part of the post-processing steps, which can be included within the conversion step (it is also partly included in getting the output dir).

These comments are mostly to get a better understanding and have a clearer view of the design model.
And if you think it might be useful, we could also have a chat about this. It would also help me to better align with your vision.

Thank you! :)

blimlim:
Thanks @atteggiani for your detailed feedback, those are all very useful points and questions. I'll try to respond to them here!

1. 
> What do you mean by "embedded in rose/cylc suites"? There is no currently no um2nc driver cm2 command, so how is CM2 driver currently embedded in the rose/cylc suite?

That's correct that there's no drivers for CM2/CM3 in um2nc. Instead there a separate python scripts within the cylc suites which perform essentially the same functions as a driver, i.e. finding inputs, setting output paths, and running the conversion:
- CM3: https://github.com/ACCESS-NRI/access-cm3-configs/blob/cm3_O25km/app/netcdf_conversion/bin/cm3_run_um2netcdf.py
- CM2: https://github.com/ACCESS-NRI/access-cm2-drivers/blob/main/src/run_um2netcdf.py

I used the common steps between these scripts and the ESM1.5 driver to help think of a general structure.

2.
> When you mention "inputs" and "outputs" files, directories, etc., are they referring to the inputs and outputs of um2nc, or the models?

Good point, I agree "inputs" and "outputs" aren't being used consistently right now which is a bit confusing. Your suggestion to consistently use "inputs" for inputs to um2nc, and outputs to refer to output netCDF files makes sense. I've tried to make this consistent in 6e7d5a6, and have also used "model history directory" in the docstrings to refer to the directories created by the model – model output files are often called "history files", and so this might be clearer than the previous use of "model output directories".


3. 
> I think finding the directory containing the model outputs (which contain some of the um2nc conversion targets above) is a step included within my number 1 above

This makes sense to me, I've updated the proposed implementation in 6e7d5a6 so that finding the input directory is included within step 1.

> number 4 is part of the post-processing steps, which can be included within the conversion step (it is also partly included in getting the output dir).

That makes sense to me, I agree it doesn't need to be considered as a core step. In the implementation I think it should be set up as a one of the general driver methods which can be redefined for specific drivers, which will allow each model to have a different naming scheme.

src/um2nc/drivers/common.py
```diff
@@ -15,6 +15,70 @@
...
+        raise NotImplementedError
+
+    def get_output_dir(self, parent_directory):
+        """
+        Given a path to an experiment parent directory, set up a directory for writing
+        netCDF outputs and return its path.
+        """
+        raise NotImplementedError
+
+    def get_input_files(self, input_directory):
```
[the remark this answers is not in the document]
blimlim: I think that's a good point, there likely isn't any need to have these as separate steps. I've combined them into a single `get_input_dir()` method in 6e7d5a6!

src/um2nc/drivers/common.py
```diff
@@ -15,6 +15,70 @@
...
+
+        input_dir = self.get_input_dir(parent_directory)
+
+        # Find fields file outputs to be converted
+        input_files = self.get_input_files(input_dir)
+
+        output_dir = self.get_output_dir(input_dir)
+
+        if len(input_files) == 0:
+            return [], []  # Don't try to run the conversion
```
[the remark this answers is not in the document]
blimlim: Good point, I've added this in 6e7d5a6!

test/drivers/test_drivers_esm1p5.py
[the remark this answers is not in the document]
blimlim: That makes sense to me, I've split these into separate tests in 6e7d5a6!

blimlim:
>  I suppose that the next step will be to convert these scripts into actual um2nc driver ... comands then, right?

Yeah definitely! For CM3, I believe we'd like to stop using the old um2netcdf script and add a driver here. For CM2 it should be viable to add a driver, but I suspect it might not be a priority

src/um2nc/drivers/common.py
```diff
@@ -15,6 +15,70 @@
 from um2nc import um2netcdf
 
 
+class ModelDriver:
+    """
+    Generic model conversion driver class. Defines a general sequence of steps
+    which are followed by the drivers.
+    """
+
+    def get_input_dir(self, parent_directory):
```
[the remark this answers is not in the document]
blimlim: Good question, just realised I missed it! Now that you mention it, "parent_directory" is a bit ambiguous.

When calling a um2nc driver for ESM1.5 or 1.6, payu doesn't directly provide the driver with the path to the directory containing the UM outputs. Instead it provides the path to a higher level model output directory (named like `output000`, `output001`, ..., `output934` etc). This then contains subdirectories with outputs for each submodel (e.g. atmosphere, ocean, ice).

The "parent_directory" is meant to refer to the directory that payu (or rose-cylc etc) would provide as an input to the driver, which the driver uses it as a starting point to locate the actual atmosphere outputs to convert. 

I've renamed it to `model_directory` – do you think this would be any clearer? Or would it be good to add comments in the driver class definition explaining what this directory refers to?

src/um2nc/drivers/common.py
```diff
@@ -15,9 +15,64 @@
...
+        Given a path to a model history directory, set up a directory for writing
+        netCDF outputs and return its path.
+        """
+        raise NotImplementedError
+
+    def set_output_path(self, input_file):
+        """
+        Given an input fields file, set the path to save its netCDF conversion.
+        """
+        raise NotImplementedError
```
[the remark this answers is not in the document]
atteggiani: This could then be easily parallelised if we want.

atteggiani (commented):
I left a few more comments.

src/um2nc/drivers/common.py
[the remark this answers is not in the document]
atteggiani: Thank you for addressing this!
I looked at the implementation, and it seems pretty good to me!

blimlim:
Thanks for the ideas and discussion @atteggiani! I've marked this one ready for review, let me know if you have any other ideas or suggestions!

.github/workflows/CI.yml
```diff
@@ -95,11 +95,19 @@ jobs:
...
+        run: um2nc driver --help
+      
+      - name: Entrypoint test of esm1p5 driver command
         shell: micromamba-shell {0}
         run: um2nc driver esm1p5 --help
 
+      - name: Entrypoint test of esm1p6 driver command
+        shell: micromamba-shell {0}
+        run: um2nc driver esm1p6 --help
+
```
atteggiani: This looks fine.
However, with more drivers appearing, each time adding a whole new step is overkill. 
I would change to one step performing tests for each driver, so when a new driver appears we'd only need to add the driver to the `drivers_list`.
```suggestion
      
- name: Entrypoint test of driver commands
        shell: micromamba-shell {0}
        run: |
          # We include '' in the drivers list to test for the `driver` command without a driver argument
          drivers_list=(
            ''
            esm1p5
            esm1p6
          )
          for driver in "${drivers_list[@]}"; do
            um2nc driver "$driver" --help
          done      
```

src/um2nc/drivers/common.py
```diff
@@ -11,13 +11,62 @@
...
+class ModelDriver(ABC):
+    """
+    Generic model conversion driver class. Defines a general sequence of steps
+    which are followed by the drivers.
+    """
+
+    @abstractmethod
+    def get_input_files(self, model_directory):
+        """
+        Find atmosphere fields files for conversion in a given model history directory.
```
atteggiani:
```suggestion
        Get a list of target paths for conversion.
```

src/um2nc/drivers/common.py
```diff
@@ -11,13 +11,62 @@
...
 
 
+class ModelDriver(ABC):
+    """
+    Generic model conversion driver class. Defines a general sequence of steps
+    which are followed by the drivers.
+    """
+
+    @abstractmethod
+    def get_input_files(self, model_directory):
```
atteggiani: To be consistent with the output one, I would change the name to:
```suggestion
    def get_input_paths(self, model_directory):
```

Also change all occurrences

src/um2nc/drivers/common.py
```diff
@@ -11,13 +11,62 @@
...
+        """
+        Find atmosphere fields files for conversion in a given model history directory.
+        """
+        ...
+
+    @abstractmethod
+    def get_output_paths(self, input_files):
```
atteggiani:
```suggestion
    def get_output_paths(self, input_paths):
```

src/um2nc/drivers/common.py
```diff
@@ -11,13 +11,62 @@
...
+        """
+        Given a list of input paths, produce a list of corresponding netCDF output paths.
+        """
+        ...
+
+    def run_conversion(self, model_directory, delete_ff, convert_args):
```
atteggiani: Why is `delete_ff` not part of `convert_args`?
Here I would pass a general `**kwargs` for keyword arguments that would be for example all conversion arguments.

src/um2nc/drivers/common.py
```diff
@@ -11,13 +11,62 @@
...
+    def get_input_files(self, model_directory):
+        """
+        Find atmosphere fields files for conversion in a given model history directory.
+        """
+        ...
+
+    @abstractmethod
+    def get_output_paths(self, input_files):
+        """
+        Given a list of input paths, produce a list of corresponding netCDF output paths.
```
atteggiani: This should be the general description.
You can be more specific in the actual driver heredoc
```suggestion
        Produce a list of output paths for conversion.
```

src/um2nc/drivers/common.py
```diff
@@ -11,13 +11,62 @@
 
 import mule
 
+from abc import ABC, abstractmethod
 from pathlib import Path
 from um2nc import um2netcdf
 
 
+class ModelDriver(ABC):
```
atteggiani, line 19: A general though about the `ModelDriver` class :

We could consider adding `model_directory`, `input_paths` and `output_paths` as properties (to make them read-only by default) since they don't change for the whole driver conversion process.

I'm not 100% sure if input_paths and output_paths could need changing after being computed for the first time, so depending on this the implementation might vary.

However, this would remove the need for them to be declared as inputs in the other methods, and substitute that with self.model_directory and self.input_paths when needed.

Also, instead of having the `filter_name_collisions` function, we could ensure input and output path characteristics at creation:
- ensure input_paths are unique
- ensure output paths are unique 
- ensure they are 1-1 correspondent

To do so we could structure them into a dictionary where input_paths are keys and output_paths are values, and compute them at the same time using a specific method.

Also, we can have a `convert` method that is defined by specific drivers, and is used by the `run_conversion` method to run the conversion of each input file to its respective output file.

All of the above could be implemented by changing the ModelDriver class to something like:

```suggestion
class ModelDriver(ABC):
      """
      Generic model conversion driver class. Defines a general sequence of steps
      which are followed by the drivers.
      """
      def __init__(self, model_directory):
          self._model_directory = model_directory
          self._input_paths = None
          self._output_paths = None
          self._input_output_mapping = None

      @property
      def model_directory(self):
          return self._model_directory

      @property
      def input_paths(self):
          return list(self.input_output_mapping.keys())
    
      @property
      def output_paths(self):
          return list(self.input_output_mapping.values())
      
      @property
      def input_output_mapping(self):
          if self._input_output_mapping is None:
              self._get_input_output_mapping()
          return self._input_output_mapping
        
      def _get_input_output_mapping(self):
          """Create an input output mapping with unique inputs and outputs paths."""
          mapping = {}
          for input_path in self._get_input_paths():
              output_path = self._get_output_path(input_path)
              if output_path not in mapping.values():
                  mapping[input_path] = output_path
          self._input_output_mapping = mapping
    
      @abstractmethod
      def _get_input_paths(self):
          """Returns a list of target input paths for conversion."""
          ...

      @abstractmethod
      def _get_output_path(self, input_path):
          """Returns the output path for a given input path."""
          ...

      @abstractmethod
      def _convert(self, input_path, output_path, **kwargs):
          """The core conversion logic."""
          ...
                              
      def run_conversion(self, delete_ff, **kwargs):
        """
        Run the conversion for each of the input and output pairs.
        """
        for input_path, output_path in self.input_output_mapping.items():
        try:
            self._convert(input_path, output_path, **kwargs)
            ... # Not sure what is needed here
        except Exception as e:
            ... # Not sure what is needed here
        else:
            if delete_ff: 
              os.remove(input_path)
```

This would require each specific model driver to define:
- `_get_input_paths(self)` that returns a list of input paths for the conversion
- `_get_output_path(self, input_path)` that returns the output path for a given input path
- `_convert(self, input_path, output_path, ...)` that has the core conversion logic (for a single input/output paths pair)

This decouples most of the processing and makes it easier to test each method, also decoupling the processing of multiple inputs at the same time. Everything should be developed for one single input and output, and then only one function does the looping through all of the inputs/outputs.

Note the `input_paths` and `output_paths` properties and related sunder (single underscore) attributes are not strictly needed for the implementation suggested. What we need is the input_output_mapping. 
However, if there are any functions in the future that don't need both of them but only need one, it might be easier to have them and keep the implementation as my suggested one. Removing them wouldn't matter that much for performance, because in any case they would be computed only once and they wouldn't occupy too much memory (unless we're talking about thousands of files).

atteggiani (changes requested):
Hi @blimlim, 

Thank you for the updates. 

I went through the whole architectural implementation and I only made some comments on the ModelDriver class (because those would influence the rest of the code). 

Please have a look at my suggestion, and feel free to ask for any clarification on it.
We might want to have a chat as I feel it would be easier to discuss these design choices.

src/um2nc/drivers/common.py
```diff
@@ -11,13 +10,100 @@
...
+
+    @abstractmethod
+    def convert(self, input_path, output_path, process_args):
+        """The core conversion logic."""
+        ...
+
+    @abstractmethod
+    def setup(self):
+        """General driver setup."""
+        ...
```
blimlim: I've added a general setup step to allow for any required model specific setup. For the ESM1.5 driver, this is used specifically to create the output directory: https://github.com/ACCESS-NRI/um2nc-standalone/blob/be1b470aa9e1441eeda358c954a69456dd34f177/src/um2nc/drivers/esm1p5.py#L55

src/um2nc/drivers/common.py
```diff
@@ -11,13 +10,100 @@
...
+        return self._input_output_mapping
+
+    def _get_input_output_mapping(self):
+        """Create an input output mapping with unique inputs and output paths."""
+        mapping = {}
+        for input_path in self.get_input_paths():
+            output_path = self.get_output_path(input_path)
+            mapping[input_path] = output_path
+
+        check_mapping_collisions(mapping)
```
blimlim: To check for collisions, I've decided to use a seperate function which works with the complete mapping. The reasons for this were that it allows for more information to be provided to the user, e.g: the error message can include all instances of collisions. I also found it simplifies the testing, as testing `_get_input_output_mapping` directly seemed like it would involve some awkward mocking.

test/drivers/test_drivers_common.py
```diff
@@ -95,154 +92,27 @@ def test_find_matching_fields_files():
...
-    # Test that critical unexpected exceptions
-    # are raised, and hence lead to the conversion crashing.
-    generic_error_message = "Test error"
-    mock_process_with_exception(Exception, generic_error_message)
-    with pytest.raises(Exception) as exc_info:
-        drivers_common.convert_fields_file_list(
-            [("fake_file", "fake_file.nc")], ARGS)
-
-    assert str(exc_info.value) == generic_error_message
-
```
blimlim, line 193: Equivalent tests are moved over to the tests of the ESM1.5 driver

test/drivers/test_drivers_esm1p5.py
```diff
@@ -0,0 +1,182 @@
...
+        assert not caplog.records
+
+    # --verbose
+    with caplog.at_level(logging.INFO):
+        esm1p5_driver.run_conversion(delete_ff=False, process_args=ARGS)
+        assert len(caplog.records) == 1
+
+        for input, output in io_map.items():
+            assert input in caplog.text
+            assert output in caplog.text
```
blimlim: From what I read online, it was generally suggested to test using concrete subclasses rather than trying to test the abstract base class directly. I've done this in the implementation, though am happy to move things around if preferred.

blimlim:
Thanks @atteggiani for the suggestions! I've had a go at implementing the suggested structure, and have added some comments around some of the choices made. Let me know if you have any further thoughts or suggestions!

src/um2nc/drivers/common.py
atteggiani: One problem of this implementation is that with
```py
for input_path in self.get_input_paths():
            output_path = self.get_output_path(input_path)
            mapping[input_path] = output_path
```
if `self.get_input_paths()` generates a non-unique list (so some input paths are the same), with `            mapping[input_path] = output_path` you will lose that information and simply override the old `input_path` key with the new `output_path`. 

So, for example, if `self.get_input_paths()` is `["input1", "input2", "input1", "input2"]`, and `self.get_output_path(input_path)` appends `_out` to the input, running the snippet above you get:

> mapping = {"input1":"input1_out", "input2":"input2_out"}

And you lose the information that some inputs were the same.
`check_mapping_collisions` is not going to handle that case (In this case it would succeed).

It's true that you should never have multiple inputs element that are exactly the same, but I think we should be as safe as possible also against mistakes.

So I think in general you want to raise errors in the following cases:
- inputs are not unique (should never happen but good to fail for mistakes)
- outputs are not unique (a.k.a, different inputs have the same output)

Slightly modifying my initial suggestion for the `_get_input_output_mapping` method we could use:
```py
from collections import Counter, defaultdict
...
def _get_input_output_mapping(self):
    """Create an input output mapping with unique inputs and outputs paths."""
    input_paths = list(self._get_input_paths())

    # Check for duplicate inputs
    duplicate_inputs = [item for item, count in Counter(input_paths).items() if count > 1]
    if duplicate_inputs:
        raise RuntimeError(f"Duplicate input paths found: {duplicate_inputs}")

    # Build mapping and check for duplicate outputs
    mapping = {}
    output_to_inputs = defaultdict(set)
    for input_path in input_paths:
        output_path = self._get_output_path(input_path)
        mapping[input_path] = output_path
        output_to_inputs[output_path].add(input_path)

    duplicate_outputs = {out: inps for out, inps in output_to_inputs.items() if len(inps) > 1}
    if duplicate_outputs:
        msg = "\n".join(f"{inps} --> {out}" for out, inps in duplicate_outputs.items())
        raise RuntimeError(
            f"Multiple input paths are mapped to the same output.\nCollisions (inputs --> output):\n{msg}"
        )

    self._input_output_mapping = mapping
```

which could be easily tested by testing the three cases (success, failure for multi-inputs, failure for same outputs)

The testing of the ModelDriver class itself requires some mocking (a mocked modelDriver subclass needs to be created (mocking an actual modelDriver), and a mocked instance of that class should also be created. But that is inevitable, and testing a specific method doesn't make it more or less onerous.

src/um2nc/drivers/common.py
atteggiani: I don't fully understand the need for this `setup` abstractmethod. 

Any model driver can define driver-specific methods that are useful to the driver (but that are specific only to some drivers).
If the ESM1.5 driver needs a way to compute the output directory a method like `create_output_directory` can be defined. 

What do you think all of the drivers would need the setup method for?

src/um2nc/drivers/common.py
atteggiani: Actually, looking at the code, we can simplify even more by substituting `_get_input_output_mapping` with `input_output_mapping` and turn this into a cached_property (so it doesn't get computed multiple times but instead the cached value is returned). And `_get_input_output_mapping` can be deleted:

```py
from functools import cached_property
from collections import Counter, defaultdict
...
@cached_property
def input_output_mapping(self):
    """Create an input output mapping with unique inputs and outputs paths."""
    input_paths = list(self._get_input_paths())

    # Check for duplicate inputs
    duplicate_inputs = [item for item, count in Counter(input_paths).items() if count > 1]
    if duplicate_inputs:
        raise RuntimeError(f"Duplicate input paths found: {duplicate_inputs}")

    # Build mapping and check for duplicate outputs
    mapping = {}
    output_to_inputs = defaultdict(set)
    for input_path in input_paths:
        output_path = self._get_output_path(input_path)
        mapping[input_path] = output_path
        output_to_inputs[output_path].add(input_path)

    duplicate_outputs = {out: inps for out, inps in output_to_inputs.items() if len(inps) > 1}
    if duplicate_outputs:
        msg = "\n".join(f"{inps} --> {out}" for out, inps in duplicate_outputs.items())
        raise RuntimeError(
            f"Multiple input paths are mapped to the same output.\nCollisions (inputs --> output):\n{msg}"
        )

    return mapping
```

src/um2nc/drivers/common.py
```diff
@@ -11,13 +10,100 @@
...
+    @abstractmethod
+    def setup(self):
+        """General driver setup."""
+        ...
+
+    def run_conversion(self, delete_ff, process_args):
+        """
+        Run the conversion for each of pair of input and output files.
+        """
+        self.setup()
```
atteggiani: As mentioned above, I don't think is needed
```suggestion
```

src/um2nc/drivers/common.py
```diff
@@ -11,13 +10,100 @@
...
+        """Returns a list of target input paths for conversion."""
+        ...
+
+    @abstractmethod
+    def get_output_path(self, input_path):
+        """Returns the output path for a given input path."""
+        ...
+
+    @abstractmethod
+    def convert(self, input_path, output_path, process_args):
```
atteggiani, line 81: I think process_args should be keyword arguments so we know exactly what arguments we are passing and it's easier to pass them through.
```suggestion
    def convert(self, input_path, output_path, **kwargs):
```

src/um2nc/drivers/common.py
```diff
@@ -11,13 +10,100 @@
...
+        """The core conversion logic."""
+        ...
+
+    @abstractmethod
+    def setup(self):
+        """General driver setup."""
+        ...
+
+    def run_conversion(self, delete_ff, process_args):
```
atteggiani, line 85: See comment above about `**kwargs`
```suggestion
    def run_conversion(self, delete_ff, **kwargs):
```

src/um2nc/drivers/common.py
```diff
@@ -11,13 +10,100 @@
...
+        Run the conversion for each of pair of input and output files.
+        """
+        self.setup()
+
+        if not self.input_output_mapping:
+            return
+
+        for input_path, output_path in self.input_output_mapping.items():
+            try:
+                self.convert(input_path, output_path, process_args)
```
atteggiani, line 95: See comment above about `**kwargs`
```suggestion
                self.convert(input_path, output_path, **kwargs)
```

src/um2nc/drivers/common.py
```diff
@@ -85,103 +171,36 @@ def get_ff_date(fields_file_path):
...
+def resolve_path(path):
     """
     Resolve path for use in comparison. Ensure that symlinks, relative paths,
     and home directories are expanded.
     """
     return os.path.realpath(os.path.expanduser(path))
 
 
-def filter_name_collisions(input_output_pairs):
+def check_mapping_collisions(mapping):
```
atteggiani: See the related message above.

src/um2nc/drivers/esm1p5.py
```diff
@@ -16,57 +14,107 @@
...
+    @property
+    def unit_suffixes(self):
+        return self._unit_suffixes
+
+    def setup(self):
+        """Check the input directory exists and create the output directory."""
+        if not self.atmosphere_dir.exists():
+            raise FileNotFoundError(
+                errno.ENOENT, os.strerror(errno.ENOENT), self.atmosphere_dir
+            )
```
atteggiani: This check should be carried out performed in the `atmosphere_dir` property

src/um2nc/drivers/esm1p5.py
```diff
@@ -16,57 +14,107 @@
...
+
+        self.output_dir.mkdir(exist_ok=True)
```
atteggiani: This step should be carried out in the `output_dir` property

atteggiani (changes requested):
Thank you @spencer for the updates. 

I added a few comments mainly on the common parts (because they affect the rest).

Feel free to start a discussion below any comments :)

test/drivers/test_drivers_esm1p5.py
atteggiani: I agree.
I haven't looked at the tests yet, but the concept makes sense to me

test/drivers/test_drivers_common.py
atteggiani: I think we should test each modelDriver separately. 
We should test the methods in the abstract ModelDriver class by creating a concrete mock of the modelDriver, and then in their own files, we should test the specific model drivers.
Of course we should not be testing methods that the specific model drivers don't implement (because they are inherited from the abstract ModelDriver class), as we would have already tested those by testing the ModelDriver class.

src/um2nc/drivers/common.py
blimlim: Thanks! I agree that it would make sense to use `**kwargs`, my one reservation with using right now is that the main `process` function is using an argparse namespace to cary most of its arguments: https://github.com/ACCESS-NRI/um2nc-standalone/blob/d2613eed6cabbe5759638e5533ca7db19081d3fe/src/um2nc/um2netcdf.py#L495

and so I think swapping to `**kwargs` here would involve converting the original parsed arguments into a dictionary, passing them to `convert` and converting them back right away to something like a `SimpleNamespace` before calling the `process` function. I wasn't sure whether this would be worth it, but happy to add something like this in.

src/um2nc/drivers/common.py
blimlim: I thought it would be helpful at the time, but yes probably not necessary as you mention

src/um2nc/drivers/common.py
blimlim: Good idea, thanks! I'll add this in

src/um2nc/drivers/common.py
atteggiani: Wouldn't the `process` function still use specific argument as:
```py
args.arg_name
```
??

In that case, the only difference would be replacing `args` with `kwargs` I guess (tests might be slightly different depending on the implementation, but I haven't checked it).

atteggiani: But you're right, at some point the common arguments should be separated into specific keyword arguments (but not at the `process` function level because that function can be called with `**kwargs` as well. The problem might be for internal functions).
This is something that should happen anyway in my opinion, but we could defer that to another PR if you prefer.

blimlim: Yeah true, it looks like it would be a fairly straightforward change! Just thinking as it would touch on a few bits of code less relevant to this PR, it could be a bit cleaner to defer it to another one.

test/drivers/test_drivers_common.py
blimlim: This sounds like a good way to go, thanks!

src/um2nc/drivers/common.py
atteggiani: Yes that's completely fine.

src/um2nc/drivers/common.py
blimlim: Added this in 6ec9e12!

src/um2nc/drivers/common.py
blimlim: Removed in 6ec9e12!

test/drivers/test_drivers_common.py
blimlim: I've moved the testing for the shared methods to `test_drivers_common.py`, using a concrete `ModelDriverSubclass`, `ModelDriverTesting` in 6ec9e12, and kept any driver specific tests in `test_drivers_esm(1p5/1p6).py`

src/um2nc/drivers/esm1p5.py
blimlim: Good idea, added in 6ec9e12

src/um2nc/drivers/esm1p5.py
blimlim: I've added this in 6ec9e12. The one part I'm not 100% sure about is if it's clear/explicit enough to developers that accessing the `.output_dir` attribute could cause a new directory to be created. E.g. if adding new tests in the future, I can imaging inadvertently creating directories and not noticing. Do you think this could be a risk, or not so much of a concern?

blimlim:
Thanks again @atteggiani for the helpful suggestions and feedback :) I've had a go adding them in, and if you get a chance to look at the latest changes that would be great!

src/um2nc/drivers/esm1p5.py
atteggiani: THat's a valid point, but I don't think is something we should be careful about. 
The only instance in which directories could be created by mistake is during unit-testing (because in integration tests we want output dirs to be created).
But in unit-tests we should always mock side-effects of functions/methods (like creation of a directory as in this case) anyway, so I wouldn't see that as an issue.

src/um2nc/drivers/esm1p5.py
```diff
@@ -16,57 +14,104 @@
...
+            raise FileNotFoundError(
+                errno.ENOENT, os.strerror(errno.ENOENT), self._atmosphere_dir
+            )
+        return self._atmosphere_dir
+
+    @property
+    def output_dir(self):
+        if not self._output_dir.exists():
+            self._output_dir.mkdir(exist_ok=True)
+        return self._output_dir
```
atteggiani, line 56: With `exist_ok=True` you don't need verification: it would create it only if it doesn't exist, otherwise nothing happens.
```suggestion
    def output_dir(self):
        self._output_dir.mkdir(exist_ok=True)
        return self._output_dir
```

src/um2nc/drivers/esm1p6.py
```diff
@@ -0,0 +1,33 @@
...
+        super().__init__(model_directory)
+        self._unit_suffixes = ESM1P6_UNIT_SUFFIXES
+        # Write netCDF directly to the atmosphere directory
+        self._output_dir = self.atmosphere_dir
+
+    def setup(self):
+        if not self.atmosphere_dir.exists():
+            raise FileNotFoundError(
+                errno.ENOENT, os.strerror(errno.ENOENT), self._atmosphere_dir
+            )
```
atteggiani:
```suggestion
```

test/drivers/test_drivers_common.py
```diff
@@ -95,154 +96,173 @@ def test_find_matching_fields_files():
     assert set(found_fields_files) == set(expected_fields_files)
 
 
-@pytest.fixture
-def base_mock_process():
-    # Create a patch of um2netcdf.process
-    patcher = mock.patch("um2nc.um2netcdf.process")
-    yield patcher.start()
-    patcher.stop()
+class ModelDriverTesting(drivers_common.ModelDriver):
```
atteggiani: This is minor, but since we're calling the driver classes `Esm1p5Driver`, `Esm1p6Driver`, etc., should we call this test Driver something like `TestDriver`?
```suggestion
class TestDriver(drivers_common.ModelDriver):
```

test/drivers/test_drivers_esm1p5.py
```diff
@@ -0,0 +1,83 @@
+
+import um2nc.drivers.esm1p5 as esm1p5_convert
```
atteggiani: I would leave it as `esm1p5` since it's the general driver
```suggestion
import um2nc.drivers.esm1p5
```

test/drivers/test_drivers_esm1p5.py
```diff
@@ -0,0 +1,83 @@
...
+                                (120, 1, 7),
+                                "aiihca.pj-012001_6hr.nc"
+                            ),
+                            (
+                                "aiihca.pjc0jan",
+                                None,
+                                "aiihca.pjc0jan.nc"
+                            ),
+                         ])
+def test_get_esm1p5_nc_filename(ff_name, ff_date, expected):
```
atteggiani:
```suggestion
def test_get_nc_filename(ff_name, ff_date, expected):
```

test/drivers/test_drivers_esm1p5.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    Check that netCDF file naming produces expected file paths for various
+    expected unit keys.
+    """
+    nc_name = esm1p5_convert.get_nc_filename(
+                        ff_name,
+                        esm1p5_convert.ESM1P5_UNIT_SUFFIXES,
+                        ff_date
+                    )
```
atteggiani: Consider importing `ESM1P5_UNIT_SUFFIXES` once instead of accessing every time from the module (valid for all instances below too). 

```py
from um2nc.drivers.esm1p5 import ESM1P5_UNIT_SUFFIXES
```

```suggestion
    nc_name = esm1p5_convert.get_nc_filename(
                        ff_name,
                        ESM1P5_UNIT_SUFFIXES,
                        ff_date
                    )
```

src/um2nc/drivers/common.py
atteggiani: This was not changed but I think it can be simplified and also some naming can be changed. 
This function doesn't relate to fieldsfiles, it simply returns a list of files inside a directory that match a regex, independent of the nature of those files.
Therefore, I would keep its name, description and input names pretty general. 
Also, why does it takes a list of contents of the directory as an input, instead of taking the directory itself and taking care of listing its content?

I would simplify to:
```suggestion
def find_matching_files(directory, pattern):
    """
    Returns a list of files in the input directory whose names match the input pattern.
    
    Parameters
    ----------
    directory : The directory where to find files
    pattern : Regex pattern for matching file names

    Returns
    -------
    List : list of files within the directory whose names match the pattern.
    """

    return [
        path for path in Path(directory).iterdir()
        if path.is_file() and re.match(pattern, path.name)
    ]
```

test/drivers/test_drivers_common.py
atteggiani, line 41: I think this list is too long for testing. 

I would only keep a few files that match the regex and a few that don't.

atteggiani:
```suggestion
def test_find_matching_files():
```

test/drivers/test_drivers_common.py
```diff
@@ -95,154 +96,173 @@ def test_find_matching_fields_files():
     assert set(found_fields_files) == set(expected_fields_files)
```
atteggiani: It's ok to check using `set`, but there is no need if we list the `expected_files` in the correct order we expect. 
The function keeps the order, so I think it would be more appropriate to test without `set`.

test/drivers/test_drivers_esm1p6.py
```diff
@@ -0,0 +1,46 @@
+import um2nc.drivers.esm1p5 as esm1p5_convert
```
atteggiani: Why is esm1p5 imported here?
```suggestion
```

test/drivers/test_drivers_esm1p6.py
```diff
@@ -0,0 +1,46 @@
...
+import um2nc.drivers.esm1p6 as esm1p6_convert
```
atteggiani: Same as above.
```suggestion
import um2nc.drivers.esm1p6
```

test/drivers/test_esm1p5.py
atteggiani, line 1: There are no tests for the `Esm1p5Driver`

test/drivers/test_drivers_esm1p6.py
```diff
@@ -0,0 +1,46 @@
...
+                                (200, 5, 1),
+                                "aiihca.pc-020005_1hr.nc"
+                            ),
+                         ])
+def test_get_esm1p6_nc_filename(ff_name, ff_date, expected):
+    """
+    Check that netCDF file naming produces expected file paths for various
+    expected unit keys.
+    """
+    nc_name = esm1p5_convert.get_nc_filename(
```
atteggiani:
```suggestion
    nc_name = esm1p6.get_nc_filename(
```

test/drivers/test_drivers_esm1p6.py
```diff
@@ -0,0 +1,46 @@
...
+                        ff_name,
+                        esm1p6_convert.ESM1P6_UNIT_SUFFIXES,
```
atteggiani: Same as above
```suggestion
                        ESM1P6_UNIT_SUFFIXES,
```

test/drivers/test_drivers_esm1p6.py
atteggiani, line 1: There are no tests for the `Esm1p6Driver`

atteggiani (changes requested):
Thank you @blimlim for addressing my review!

The updates overall look pretty good! I like the tests you implemented. 

I left a few more suggestions and I noticed the specific model drivers are missing tests.

Please refer to the specific comments for details

src/um2nc/drivers/esm1p5.py
blimlim: Very good point, added in 9d45124

src/um2nc/drivers/esm1p6.py
blimlim: Oh thanks for catching that. Removed in 9d45124

test/drivers/test_drivers_common.py
blimlim: Yeah that's a lot more consistent! I've changed the name in 9d45124

test/drivers/test_drivers_esm1p5.py
blimlim: Makes sense to me, I've made this change in 9d45124/7b7e69c

test/drivers/test_drivers_esm1p5.py
blimlim: Changed in 9d45124

test/drivers/test_drivers_esm1p5.py
blimlim: Added in 9d45124

test/drivers/test_drivers_common.py
blimlim: I've cut it down in 9d45124

blimlim: Added in 9d45124

test/drivers/test_drivers_common.py
blimlim: Removed `set` in 9d45124

test/drivers/test_drivers_esm1p6.py
blimlim: Changed in 9d45124

test/drivers/test_drivers_esm1p6.py
blimlim: Changed in 9d45124

test/drivers/test_esm1p5.py
blimlim: The differences between the Esm1p5Driver and the base ModelDriver are essentially the setting of the `atmosphere_dir`, `output_dir`, and `unit_suffixes` properties, and the specific implementations of the `get_input_paths` and `get_output_path` methods.

The `get_input_paths` method only calls the [`find_esm1px_fields_files`](https://github.com/ACCESS-NRI/um2nc-standalone/blob/7b7e69c0a3968c51d1c1033d56afc7b91369bbf9/src/um2nc/drivers/esm1p5.py#L84) function, which is could probably do with an additional test added.

The `get_output_path` method almost directly calls the [`get_nc_filename`](https://github.com/ACCESS-NRI/um2nc-standalone/blob/7b7e69c0a3968c51d1c1033d56afc7b91369bbf9/src/um2nc/drivers/esm1p5.py#L98) function which has [its own tests](https://github.com/ACCESS-NRI/um2nc-standalone/blob/7b7e69c0a3968c51d1c1033d56afc7b91369bbf9/test/drivers/test_drivers_esm1p5.py#L36-L71)

I'm thinking one option would be to remove these functions and implement them directly in `get_input_paths` and `get_output_path` methods, and then test them through tests of the drivers. I don't expect these functions to be required outside of the drivers and so it could be a bit cleaner. Would this be preferable?

test/drivers/test_drivers_esm1p6.py
blimlim: This one is imported as the `get_nc_filename` function is defined under `um2nc.drivers.esm1p5`. It was previously suggested to test this under `test_drivers_esm1p6.py` when using the esm1p6 `ESM1P6_UNIT_SUFFIXES`. One [option here](https://github.com/ACCESS-NRI/um2nc-standalone/pull/229#discussion_r3012997781) would be to incorporate the `get_nc_filename` function into the Esm1p5Driver, in which case this import wouldn't be needed!

src/um2nc/drivers/common.py
blimlim: I think this seems like a pretty good change, I've added it in 9d45124

> Also, why does it takes a list of contents of the directory as an input, instead of taking the directory itself and taking care of listing its content?

There was previously a preference for keeping filesystem interactions higher up, but I think the suggested change makes the function conceptually simpler so have no issues adding it in

test/drivers/test_drivers_esm1p6.py
atteggiani: If `get_nc_filename` is a function that should be reused by multiple drivers, then I think the solution would be to define it in `common.py` and test it only once in `test_common.py`.

Also, looking at the [get_nc_filename](https://github.com/ACCESS-NRI/um2nc-standalone/blob/219-esm1.6-driver/src/um2nc/drivers/esm1p5.py#L103-L141) function, I see heavily reliant on the model outputs naming scheme, which makes it very fragile as that could change also for the same model (for example if the user customises it). Therefore, I am wondering if we should modify it now to check for regex matching, and in case it doesn't match, simply append ".nc" to the output file for the conversion.

In addition, since the function doesn't properly "get" the name, but it in facts "creates" it (based on some inputs), I would rename it to something like `create_nc_filename`.

test/drivers/test_esm1p5.py
atteggiani: We should still test each of those properties.

For the methods, if they call a function that needs to be reused by multiple drivers (like in this case), I would still keep the function in the `common.py` and test it in `test_common.py`, but still test each specific method implemented. Even if the test would only be testing that the function gets actually called (depending on the specific method).
The same exact concept applies for Esm1p6Driver.

test/drivers/test_drivers_esm1p6.py
blimlim: Happy to add in the regex and name changes! I'm a bit reluctant to move this function to `common.py` as it likely won't be relevant for future model drivers like CM3 where a different naming scheme may be used. For ESM1.6 there are also discussions on bringing in a different naming scheme

atteggiani: In general, I don't think is a problem if other drivers don't use it. As long as it's a function reused in multiple driver files (in this case ESM 1.5 and 1.6) I think it should be in the `common.py` file. Then, if the CM3 driver (for example) has a completely different naming scheme, that can be defined specifically for the driver (even by integrating it directly within its `get_output_paths` method).

However, if you think that the naming scheme for ESM1.6 will likely change so that this function will not be reused anymore, then I would directly define the logics directly within the respective `get_output_paths` model driver methods (for now the ESM1.5 and ESM1.6 might be very similar, but that's not an issue).

blimlim: Thanks! Following discussions with @joshuatorrance, we will target single variable files for ESM1.6 with a changed naming scheme (aiming to be closer to the ocean output names). Given this, I'll have a go with your suggestion of incorporating the logic directly into the `get_output_paths` methods.

src/um2nc/drivers/esm1p5.py
```diff
@@ -57,6 +60,21 @@ def output_dir(self):
...
+            xhist_nml = f90nml.read(self.atmosphere_dir / "xhist")
+            self._runid = xhist_nml["nlchisto"]["run_id"]
+        return self._runid
+
+    @property
+    def input_name_pattern(self):
+        # regex pattern for matching input files
+        if self._input_name_pattern is None:
+            self._input_name_pattern = get_fields_file_pattern(self.runid)
+        return self._input_name_pattern
```
blimlim: Since we are now using the regex pattern to:
1. Find input files in the input directory https://github.com/ACCESS-NRI/um2nc-standalone/blob/d74acc8717cef08c5e894a68758c1d1d00df1d29/src/um2nc/drivers/esm1p5.py#L98
2. Check input filenames and extract the "unit" from them when creating output filenames: https://github.com/ACCESS-NRI/um2nc-standalone/blob/d74acc8717cef08c5e894a68758c1d1d00df1d29/src/um2nc/drivers/esm1p5.py#L141-L147

It made sense to me to generate the regex pattern once, and store it as a property to reuse each time it's needed.

src/um2nc/drivers/esm1p5.py
```diff
@@ -95,79 +118,42 @@ def get_output_path(self, input_path):
...
-        input_atm_dir, fields_file_name_pattern
-    )
-    if len(atm_dir_fields_files) == 0:
-        warnings.warn(
-            f"No files matching pattern '{fields_file_name_pattern}' "
-            f"found in {input_atm_dir.resolve()}. No files will be "
-            "converted to netCDF."
-        )
-
-    return atm_dir_fields_files
```
blimlim: This function originally performed:
1. Generate the regex pattern for matching files
2. Call `find_matching_files` with the regex pattern

Since the first step has been moved to a property, it didn't make sense to me to keep this function, the `find_matching_files` call has been moved directly to the `get_input_files` method.

src/um2nc/drivers/esm1p5.py
```diff
@@ -95,79 +118,42 @@ def get_output_path(self, input_path):
...
+            suffix = f"_{self.unit_suffixes[unit]}"
+        except KeyError:
+            warnings.warn(
+                f"Unit code '{unit}' from filename f{input_name} "
+                "not recognized. Frequency information will not be added "
+                "to the netCDF filename.", RuntimeWarning
+            )
+            suffix = ""
 
-def get_nc_filename(fields_file_name, unit_suffixes, date=None):
```
blimlim: This has been moved to the `_create_nc_filename` method, which is then called via `get_output_path`

src/um2nc/drivers/esm1p5.py
```diff
@@ -71,17 +89,22 @@ def convert(self, input_path, output_path, process_args):
...
+        if len(input_paths) == 0:
+            warnings.warn(
+                f"No files matching pattern '{self.input_name_pattern}' "
+                f"found in {self.atmosphere_dir.resolve()}. No files will be "
+                "converted to netCDF."
+            )
+
+        return input_paths
 
     def get_output_path(self, input_path):
```
blimlim, line 104: The ESM1.6 driver currently inherits this, and `_create_nc_name`, as the logic at the moment is still identical. This will change when we know in more detail the future ESM1.6 specific changes.

test/drivers/test_drivers_esm1p5.py
```diff
@@ -32,43 +46,48 @@
...
                                 "aiihca.pjc0jan.nc"
                             ),
+                            (
+                                "non_matching",
+                                None,
+                                "non_matching.nc"
+                            ),
                          ])
-def test_get_nc_filename(ff_name, ff_date, expected):
+def test_create_nc_filename(ff_name, ff_date, expected, mock_atmosphere_dir):
```
blimlim: With the `get_nc_filename` function moved to the `_create_nc_filename` method, these tests now test the driver, and require it to: 1. set the regex pattern for matching files and extracting information, and 2. create the output file names

test/drivers/test_esm1p5.py
blimlim: Now added tests of the `get_input_paths`, `get_ouptut_paths` and `convert` methods for each driver in d74acc8

test/drivers/test_drivers_esm1p6.py
blimlim: With the reorganisation in d74acc8, the esm1p6 driver tests no longer import from the esm1p5 driver

blimlim:
Thanks @atteggiani for those suggestions. I've done a little bit more reorganising based on them. I've added some comments to outline my reasoning for the additional reorganisation, but happy again for any feedback :)

src/um2nc/drivers/common.py
```diff
@@ -128,34 +128,30 @@ def get_fields_file_pattern(run_id: str):
             f"Received run_id = {run_id} with length {len(run_id)}. run_id must be length 5"
         )
```
atteggiani: What is the purpose of this check? It doesn't really matter for the result of the function: for any `run_id`, this function would still return the correct pattern.

If a check on the run_id is needed, I would compute it when the run_id is calculated (i.e. within each driver's `runid` property). See related [comment](https://github.com/ACCESS-NRI/um2nc-standalone/pull/229/changes#r3019593390)

src/um2nc/drivers/common.py
atteggiani: Since this function only returns a regex pattern (which is specific to every model), I would instead define the pattern within the specific modelDriver as a property. 

For details, refer to my [comment](https://github.com/ACCESS-NRI/um2nc-standalone/pull/229/changes#r3019606557) below.

src/um2nc/drivers/common.py
```diff
@@ -37,34 +128,30 @@ def get_fields_file_pattern(run_id: str):
...
 
     return fields_file_name_pattern
 
 
-def find_matching_files(dir_contents, fields_file_name_pattern):
+def find_matching_files(directory, pattern):
     """
-    Find files in list of paths with names matching fields_file_name_pattern.
-    Used to find UM outputs in a simulation output directory.
+    Returns a list of files in the input directory whose names match the input pattern.
```
atteggiani:
```suggestion
    Returns a list of files in the directory whose names match the pattern.
```

src/um2nc/drivers/esm1p5.py
```diff
@@ -1,161 +1,159 @@
...
+    def unit_suffixes(self):
+        return self._unit_suffixes
+
+    @property
+    def runid(self):
+        # run ID used in input file names
+        if self._runid is None:
+            xhist_nml = f90nml.read(self.atmosphere_dir / "xhist")
+            self._runid = xhist_nml["nlchisto"]["run_id"]
+        return self._runid
```
atteggiani: We can add a sanity check for runid here if needed:

```suggestion
    @property
    def runid(self):
        # run ID used in input file names
        if self._runid is None:
            xhist_nml = f90nml.read(self.atmosphere_dir / "xhist")
            self._runid = xhist_nml["nlchisto"]["run_id"]
            if (length:=len(self._runid)) != 5:
                raise ValueError(
                    f"Found run ID {run_id} with length {length}. Run ID must be of length 5."
                )
        return self._runid
```

src/um2nc/drivers/esm1p5.py
atteggiani: Defining the regex pattern once (compiled) makes sense. 
However, since it is specific to each model driver, I would define it directly here as a property.

```suggestion
    @cached_property
    def input_name_pattern(self):
        return re.compile(rf"^(?P<stem>{self.run_id}a.p(?P<unit>[a-z]))[a-z0-9]+$")
```

src/um2nc/drivers/esm1p5.py
atteggiani: Yes, I think that's completely fine. 
If the changes are minimal, it makes sense to still inherit from `Esm1p5Driver` and redefine the methods/properties that change

src/um2nc/drivers/esm1p5.py
```diff
@@ -96,81 +118,42 @@ def get_output_path(self, input_path):
         --------
         output_path: Path for writing output netCDF.
         """
-        output_filename = get_nc_filename(input_path.name, self.unit_suffixes, get_ff_date(input_path))
+
+        output_filename = self._create_nc_filename(input_path.name, get_ff_date(input_path))
 
         return self.output_dir / output_filename
 
+    def _create_nc_filename(self, input_name, date=None):
```
atteggiani: Is there any advantage in defining this additional method, rather than writing this directly to `get_output_path`?

atteggiani (changes requested):
Thank you @blimlim, 
I left a few more comments on the updates. 

I haven't checked the updated tests yet, I will do so once the main implementation is fixed.

src/um2nc/drivers/common.py
blimlim: It was originally making sure that a bad runid hadn't been passed into the `get_fields_file_pattern` as the UMv7 is restricted to using 5 character runids. However I think it's not completely necessary – we're reading the runid from a file written by the UM, and so the runid should always match what the model used. I've removed the check in  74d2abb

src/um2nc/drivers/common.py
blimlim: Added in 74d2abb

src/um2nc/drivers/esm1p5.py
blimlim: Added in 74d2abb, just without the runid check which I think is not completely necessary

src/um2nc/drivers/esm1p5.py
blimlim: I'd initially thought the code came out a bit more readable separating it out like this, however I've moved it directly into `get_output_path` in 74d2abb and it still reads ok to me

src/um2nc/drivers/common.py
blimlim: Good idea, thanks! Added in 74d2abb

src/um2nc/drivers/esm1p5.py
blimlim: Thanks good idea, I've added this in 74d2abb

test/drivers/test_common.py
atteggiani, line 1: It is good practice to call the test files `test_<name_of_file_they_are_testing>`.
However, these test files within the `drivers` folder have a `_drivers` portion added (e.g., `test_drivers_common` instead of `test_common`). Is there a reason to add that? 

Also, there are a few test missing:

- `run_conversion` method:
  - case when `not self.input_output_mapping` is `True`
  - case when `delete_ff` is `True` and there is removal

- `get_ff_date` function
- `ModelDriver` initialisation 
- `ModelDriver` properties

test/drivers/test_esm1p5.py
atteggiani, line 1: The following tests of the `Esm1p5Driver` class are missing:

- initialisation
- properties

test/drivers/test_drivers_esm1p6.py
atteggiani, line 1: These tests are not necessary, as we inherit from `Esm1p5Driver` which have been tested in its own file. 

The only test that is necessary (and missing) is the initialisation of the `Esm1p6Driver`

atteggiani (changes requested):
Thank you @blimlim for the extensive work on this issue! 

Everything looks pretty good to me!! 

Also tests look good. I left a couple more comments related to them, but they relate mostly to some tests/cases missing.

test/drivers/test_common.py
blimlim: > However, these test files within the drivers folder have a _drivers portion added (e.g.,

Good point I don't think there was a reason for it, I've removed the extra `drivers` portion in c251d9a

I've also added the additional tests for `run_conversion`, `get_ff_date` and the `ModelDriver`

test/drivers/test_esm1p5.py
blimlim: Good thinking, I've added tests for these in c251d9a

test/drivers/test_drivers_esm1p6.py
blimlim: Good point, I've added the initialisation test in c251d9a and removed the tests that are already covered by the `Esm1p5Driver` tests. 

I've kept int the `test_get_output_path` as it's expected to give different output names for the `Esm1p6Driver`, and added some tests of the `Esm1p6Driver` properties where they differ from the `Esm1p5Driver`

blimlim:
Thanks @atteggiani for the thorough reviewing! I've added in those additional tests and if you are happy to take a look, that would be much appreciated!

test/drivers/test_common.py
```diff
@@ -0,0 +1,307 @@
...
+    def get_output_path(self, input_path):
+        return
+
+    def convert(self, input_path, output_path, process_args):
+        return
+
+
+def test_driver_initialisation():
+    """Test that the TestDriver successfully initialises."""
+    TestDriver(Path("fake_model_dir"))
```
atteggiani:
```suggestion
def test_driver_initialisation():
    """Test that the TestDriver successfully initialises."""
    driver = TestDriver(Path("fake_model_dir"))
    assert driver._model_directory == "fake_model_dir"
    assert driver._input_paths is None
    assert driver._output_paths is None    
    assert driver._input_output_mapping is None    
```

atteggiani:
```suggestion
    directory=Path("fake_model_dir")
    instance=TestDriver(directory)
    assert instance._model_directory == directory
    assert instance._input_paths is None
    assert instance._output_paths is None
    assert instance._input_output_mapping is None
```

test/drivers/test_esm1p5.py
```diff
@@ -0,0 +1,236 @@
...
+    patcher.stop()
+
+
+def test_esm1p5_initialisation(tmp_path):
+    """Test initialisation of the Esm1p5Driver."""
+    model_dir = tmp_path / "model_dir"
+    atmosphere_dir = model_dir / "atmosphere"
+    atmosphere_dir.mkdir(parents=True)
+
+    Esm1p5Driver(model_dir)
```
atteggiani:
```suggestion
def test_esm1p5_initialisation(tmp_path, mock_atmosphere_dir):
    """Test initialisation of the Esm1p5Driver."""
    model_dir = tmp_path / "model_dir"
    atmosphere_dir = model_dir / "atmosphere"
    output_dir = atmosphere_dir / "netCDF"
    mock_atmosphere_dir.return_value = atmosphere_dir
    instance=Esm1p5Driver(model_dir)
    assert instance._atmsophere_dir == atmosphere_dir
    assert instance._output_dir == output_dir
    assert instance._unit_suffixes == ESM1P5_UNIT_SUFFIXES
```

test/drivers/test_esm1p5.py
```diff
@@ -0,0 +1,236 @@
...
+def test_driver_directories(tmp_path):
+    """Test the atmosphere_dir and output_dir properties are correctly set."""
+    model_dir = tmp_path / "output000"
+    atmosphere_dir = model_dir / "atmosphere"
+    atmosphere_dir.mkdir(parents=True)
+
+    driver = Esm1p5Driver(model_dir)
+
+    assert driver.atmosphere_dir == atmosphere_dir
+    assert driver.output_dir == driver.atmosphere_dir / "netCDF"
```
atteggiani:
```suggestion
def test_atmosphere_dir():
    """Test the atmosphere_dir property is correctly set."""
    model_dir = Path("output000")
    atmosphere_dir = Path("output000/atmosphere")
    with patch("pathlib.Path.exists", return_value=True):
        driver = Esm1p5Driver(model_dir)
    assert driver.atmosphere_dir == atmosphere_dir

@pytest.mark.parametrize("output_dir_exists", [True,False])
def test_output_dir(tmp_path, output_dir_exists):
    """Test the output_dir property is correctly set."""
    model_dir = tmp_path / "output000"
    atmosphere_dir = model_dir / "atmosphere"
    atmosphere_dir.mkdir(parents=True)
    output_dir = model_dir / "netCDF"
    if output_dir_exists:
        output_dir.mkdir(parents=True)
    else
        assert not output_dir.exists()
    driver = Esm1p5Driver(model_dir)
    assert driver.output_dir == output_dir
    assert output_dir.exists()
```

test/drivers/test_esm1p6.py
```diff
@@ -0,0 +1,124 @@
...
+    patcher.stop()
+
+
+def test_esm1p6_initialisation(tmp_path):
+    """Test initialisation of the Esm1p6Driver."""
+    model_dir = tmp_path / "model_dir"
+    atmosphere_dir = model_dir / "atmosphere"
+    atmosphere_dir.mkdir(parents=True)
+
+    Esm1p6Driver(model_dir)
```
atteggiani:
```suggestion
def test_esm1p6_initialisation(mock_atmosphere_dir):
    """Test initialisation of the Esm1p6Driver."""
    model_dir = Path("model_dir")
    atmosphere_dir = model_dir / "atmosphere"
    mock_atmosphere_dir.return_value = atmosphere_dir
    driver=Esm1p6Driver(model_dir)
    assert driver._unit_suffixes == ESM1P6_UNIT_SUFFIXES
    assert driver._output_dir == atmosphere_dir
```

test/drivers/test_esm1p6.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    driver = Esm1p6Driver(Path("fake_model_dir"))
+    mock_runid.return_value = "aiihc"
+
+    mock_output_dir.return_value = Path("output_dir")
+    mock_get_ff_date.return_value = ff_date
+
+    output_path = driver.get_output_path(input_path)
+
+    assert output_path == expected_output
```
atteggiani: This has all been tested within the esm1p5 driver, and it's not needed

```suggestion
```

atteggiani (changes requested):
Thank you @blimlim. 

A few last comments on the driver tests, mainly on their initialisation. 

Also, by testing the esm1p6 driver initialisation, no other test is required.

blimlim:
Thanks @atteggiani for taking another look! I've added these changes to the tests in 6c861fe and 3cad911 :)

atteggiani (approved):
Thank you @blimlim for working on this and for addressing my many reviews!!

This looks pretty good to me :D Great job!

Just a note for merging: might be preferable to use squash merge and refactor the message, as there have been many commits added to this PR :)

blimlim:
Awesome thanks @atteggiani for the thorough reviewing, definitely ended up with a much improved result because of it!

PR #231, opened 2026-03-10
Reorder fix_time_coord call to reduce memory use

MartinDix:
Move the `fix_time_coord` call to reduce memory use, particularly with large restart files.

See https://github.com/ACCESS-NRI/um2nc-standalone/issues/225#issuecomment-4009568589 for memory statistics.

### DIFF
--- a/src/um2nc/um2netcdf.py
+++ b/src/um2nc/um2netcdf.py
@@ -514,7 +514,8 @@ def process(infile, outfile, args):
             #     print(c.name(), c.item_code)
 
             sman.write(c, zlib=True, complevel=args.compression, unlimited_dimensions=dims, fill_value=fill)
-
+            # Save memory by setting this to None after use
+            c.data = None
 
 def process_cubes(cubes, mv, args):
     set_item_codes(cubes)
@@ -557,6 +558,9 @@ def process_cubes(cubes, mv, args):
             if require_heaviside_t(c.item_code) and heaviside_t:
                 apply_mask(c, heaviside_t, args.hcrit)
 
+        # TODO: can item code get removed here when new cubes returned?
+        c, unlimited_dimensions = fix_time_coord(c)
+
         # TODO: some cubes lose item_code when replaced with new cubes
         c = fix_pressure_levels(c) or c  # NB: use new cube if pressure points are modified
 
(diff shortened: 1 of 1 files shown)


atteggiani (approved):
Thank you @MartinDix. 

I approved this, as a temporary solution that reduces memory usage. 

However, I think this solution is still not the final one, because we should be able to leverage iris lazy loading and process each file using only slightly more than the memory of each field (in cases when masking or other processing that require multiple fields together are not performed), and a multiple of that (roughly equal to MEM_FIELD * N_FIELDS_NEEDED) when masking or other processing that require multiple fields together are performed.

blimlim:
@atteggiani @MartinDix, when working on #234 @joshuatorrance found that this change causes an error when using python 3.10:
```
  File "/scratch/tm70/sw6175/micromamba/envs/310_test/bin/um2nc", line 6, in <module>
    sys.exit(main())
  File "/g/data/tm70/sw6175/development/um2nc/um2nc-standalone/src/um2nc/cli.py", line 262, in main
    process(args.infile, args.outfile, args)
  File "/g/data/tm70/sw6175/development/um2nc/um2nc-standalone/src/um2nc/um2netcdf.py", line 518, in process
    c.data = None
  File "/scratch/tm70/sw6175/micromamba/envs/310_test/lib/python3.10/site-packages/iris/cube.py", line 2874, in data
    self._data_manager.data = data
  File "/scratch/tm70/sw6175/micromamba/envs/310_test/lib/python3.10/site-packages/iris/_data_manager.py", line 236, in data
    raise ValueError(emsg.format(self.shape, data.shape))
ValueError: Require data with shape (1, 38, 145, 192), got ().
```

It appears that support for setting `data=None` added in [iris v3.12.x](https://github.com/SciTools/iris/blob/v3.12.x/lib/iris/_data_manager.py) which only supports python >= 3.11

Would there be any issues with dropping um2nc support for python 3.10?

atteggiani:
> It appears that support for setting `data=None` added in [iris v3.12.x](https://github.com/SciTools/iris/blob/v3.12.x/lib/iris/_data_manager.py) which only supports python >= 3.11
> 
> Would there be any issues with dropping um2nc support for python 3.10?

Oooops, good catch! That's the importance to have a testing suite with 100% coverage :D

I don't think there should be any issue dropping support for Python 3.10

PR #232, opened 2026-03-12
Ensure native byteorder

MartinDix:
Fix the problem with restart files described in #230.

Test results at https://github.com/ACCESS-NRI/um2nc-standalone/issues/230#issuecomment-4043361852

Integration tests show no change.

### DIFF
--- a/src/um2nc/um2netcdf.py
+++ b/src/um2nc/um2netcdf.py
@@ -561,6 +561,13 @@ def process_cubes(cubes, mv, args):
         # TODO: can item code get removed here when new cubes returned?
         c, unlimited_dimensions = fix_time_coord(c)
 
+        # Ensure the data is in the native byte order
+        # (iris problem with single level fields packed to land points)
+        if not c.data.dtype.isnative:
+            logging.info(f"Swapping byteorder for {c.var_name} {c.name()}")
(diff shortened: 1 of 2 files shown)


atteggiani (dismissed):
Thank you @MartinDix for adding this fix!

A minor comment is that the portion added could be turned into a function (with testing) and the function could be run instead of leaving the code as it is. 

However, I'm happy for the code to be merged as it, because we will do these types of restructuring anyway once we add tests to bring coverage to 100%

MartinDix:
I've now added some test changes so that it passes.

Checking that `cube.data is None` at the end was pointless because the masking could never have been applied w/o a heaviside array anyway.

Thoughts @atteggiani

atteggiani (commented):
The changes in the tests are not related to these changes in the code. 

Therefore, I would not change tests like you did in this PR. 

I didn't look at the test suite carefully, but it should be restructured and completed, therefore I would not worry about changing it now (as long as the unchanged tests pass).

MartinDix:
The dummy cube needed to be changed else there's a test error from the new code because `cube.data.dtype` doesn't exist.

atteggiani (approved):
Looks good, thank you @MartinDix!

Issue #196, opened 2025-03-06
Create regression tests with pytest and embed in CI testing


atteggiani:
## Overview
The regression tests currently implemented are based on a [bash script](https://github.com/ACCESS-NRI/um2nc-standalone/blob/main/integration/regression_tests.sh) run. 
It would be preferable if all tests could be run using `pytest` and embedded within the [CI workflow](https://github.com/ACCESS-NRI/um2nc-standalone/blob/main/.github/workflows/CI.yml)

An example of integration tests implemented using `pytest`, along with their inclusion within a CI workflow that runs them on `Gadi`, can be found in the [replace_landsurface](https://github.com/ACCESS-NRI/replace_landsurface) repo.

## Solution

- [ ] Rewrite regression tests using `pytest`
- [ ] Add regression tests to CI workflow

PR #234, opened 2026-04-10
Split output into single variable files

joshuatorrance:
- [x] Update code
- [x] Update tests
  - Only simple test at the moment...
- [x] Update docs
  - ~~Main branch's README is not up to date itself~~


blimlim:
Thanks @joshuatorrance for getting this working, I'll take a look at the unit tests.

I'm wondering whether the number of files produced would be a concern if we split each month's output separately.  In the esm1.6 run at, `/g/data/p73/archive/CMIP7/ACCESS-ESM1-6/production/PI-CNP-concentrations/output000/atmosphere/netCDF`, each month's data has 299 variables, which would result in 3588 files in a year, and a long 500 year control run would produce nearly 1.8m files for the atmosphere.

If this is a concern, we could try to group each 12 months of files together in the driver and pass them together as a list to the `process` function. I'll think about how the drivers could be adapted to do this if you think it could be worthwhile. I'm hoping the upcoming changes to the drivers could simplify a change like this.

joshuatorrance:
> Thanks @joshuatorrance for getting this working, I'll take a look at the unit tests.
> 
> I'm wondering whether the number of files produced would be a concern if we split each month's output separately. In the esm1.6 run at, `/g/data/p73/archive/CMIP7/ACCESS-ESM1-6/production/PI-CNP-concentrations/output000/atmosphere/netCDF`, each month's data has 299 variables, which would result in 3588 files in a year, and a long 500 year control run would produce nearly 1.8m files for the atmosphere.
> 
> If this is a concern, we could try to group each 12 months of files together in the driver and pass them together as a list to the `process` function. I'll think about how the drivers could be adapted to do this if you think it could be worthwhile. I'm hoping the upcoming changes to the drivers could simplify a change like this.

#### Number of Files
Good point, yearly files sound like a good idea. Although there is a risk that higher frequency data files (e.g. a whole year of hourly data) would get unmanageably large. So there's a bit of tension between keeping the number of files under control, keeping the file sizes reasonable, and having all frequencies treated the same.

Keeping the frequencies consistent feels important... perhaps we should keep things configurable then for long runs, which are unlikely to have higher-frequency data, the output can be yearly, and for shorter runs it can be monthly or less?

#### File sizes
From a look at my test directory, the single-var monthly files are mostly less than 300kB, there's a few files that are a few MB, the biggest is 3.1MB. So for the biggest, a year of monthly data would be 37MB which should be fine. If that was hourly and the compression ratio is the same then it would be 24GB.

The increase in disk usage for the separate-variable files looks fine by the way:
- unsplit file : 69MB
- split files : 79MB

#### Organising files
When we get up to redoing the filenames would may want to consider nesting the files under variables names and/or frequencies rather than dumping them all into the same directory. Something like `outputX/atmosphere/{freq}/{var}/*.nc` might work.

src/um2nc/um2netcdf.py
```diff
@@ -502,20 +502,43 @@ def process(infile, outfile, args):
...
     cubes = iris.load(infile)
 
-    with iris.fileformats.netcdf.Saver(outfile, NC_FORMATS[args.ncformat]) as sman:
-        # Add global attributes
-        if not args.nohist:
-            add_global_history(infile, sman)
+    delayed_writes = []
 
-        sman.update_global_attributes({"Conventions": "CF-1.6"})
+    if args.single_field_files:
```
blimlim: The `process` function is called in two places, through the driver: https://github.com/ACCESS-NRI/um2nc-standalone/blob/0522a476beb0e96512d2fa2eec097541c87c2112/src/um2nc/drivers/esm1p5.py#L83

and through the `um2nc convert infile outfile` command (used to convert a single UM file to netCDF through the command line):
https://github.com/ACCESS-NRI/um2nc-standalone/blob/0522a476beb0e96512d2fa2eec097541c87c2112/src/um2nc/cli.py#L262

The `um2nc convert infile outfile` command currently fails as it doesn't have the `single_field_files` argument set. 

```
  File "/g/data/tm70/sw6175/development/um2nc/um2nc-standalone/src/um2nc/um2netcdf.py", line 507, in process
    if args.single_field_files:
       ^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Namespace' object has no attribute 'single_field_files'
``` 

A couple quick options might be:

1. Force `args.single_fields_files = False` for the `um2nc convert` command [around here](https://github.com/ACCESS-NRI/um2nc-standalone/blob/0522a476beb0e96512d2fa2eec097541c87c2112/src/um2nc/cli.py#L260-L262)
```diff
    # Run selected command
    if args.command == "convert":
+      args.single_fields_files = False
        process(args.infile, args.outfile, args)
```

2. Allow `--single-fields-files` to be an argument for both the `um2nc driver ...` and `um2nc convert ...` commands by adding it to the [`common_args`](https://github.com/ACCESS-NRI/um2nc-standalone/blob/0522a476beb0e96512d2fa2eec097541c87c2112/src/um2nc/cli.py#L53). Admittedly I can't think of a use case for running something like:
```
um2nc convert --single-fields-files infile output.nc
```

src/um2nc/um2netcdf.py
```diff
@@ -502,20 +502,43 @@ def process(infile, outfile, args):
...
+
+            with iris.fileformats.netcdf.Saver(filename, NC_FORMATS[args.ncformat], compute=False) as sman:
+                print(c.name(), c.var_name, nf)
+                # Add global attributes
+                if not args.nohist:
+                    add_global_history(infile, sman)
+  
+                sman.update_global_attributes({"Conventions": "CF-1.6"})
+                delayed = sman.write(c, zlib=True, complevel=args.compression, unlimited_dimensions=dims, fill_value=fill)
+                delayed_writes.append(delayed)
```
blimlim: With Martin's changes in #237 I think these changes can be skipped

src/um2nc/um2netcdf.py
```diff
@@ -502,20 +502,43 @@ def process(infile, outfile, args):
...
         for c, fill, dims in process_cubes(cubes, mv, args):
-            # if args.verbose:
-            #     print(c.name(), c.item_code)
+            nf += 1
+
+            # Prefix the filename with the field's name and _
+            filename = f'{outfile.parent}/{c.var_name}_{outfile.name}'
+
+            with iris.fileformats.netcdf.Saver(filename, NC_FORMATS[args.ncformat], compute=False) as sman:
+                print(c.name(), c.var_name, nf)
```
blimlim:
```suggestion
                logging.info(f"{c.name}, {c.var_name}, {nf}")
```

blimlim (commented):
Thanks @joshuatorrance for your work on this. I'm glad that the change to implement this didn't end up being too involved. I've just added a couple of comments.

src/um2nc/um2netcdf.py
atteggiani: I vote for number 2, having `--single-fields-files` as an option for both `driver` and `convert` commands."
Even though there might not be many cases of users needing this with the `convert` command, I don't see a reason not to include it since it makes development easier and has no downsides.

atteggiani: I think in general "driver-only options" should be used only when the option relates specifically to a driver processing (like in the case of the current `--delete-ff` option)

src/um2nc/cli.py
```diff
@@ -259,7 +266,8 @@ def main():
 
     # Run selected command
     if args.command == "convert":
-        process(args.infile, args.outfile, args)
+        delayed_writes = process(args.infile, args.outfile, args)
+        dask.compute(*delayed_writes)
```
atteggiani: Related to [comment below](https://github.com/ACCESS-NRI/um2nc-standalone/pull/234/changes#r3150871989), this change could be reverted.

src/um2nc/um2netcdf.py
atteggiani: > The `um2nc convert infile outfile` command currently fails as it doesn't have the single_field_files argument set.

Also, why no test is failing?

test/test_um2netcdf.py
```diff
@@ -379,7 +379,7 @@ def test_ancillary_files_no_support():
         mload.return_value = af
 
         with pytest.raises(NotImplementedError):
-            um2nc.process("fake_infile", "fake_outfile", args=None)
+            _delayed_writes = um2nc.process("fake_infile", "fake_outfile", args=None)
```
atteggiani: Why this change?

test/test_single_field.py
atteggiani, line 1: This test uses a "real" file, which has not been used so far for the other unit-tests.
There are some integration tests on Gadi, but those are mostly for reproducible outputs. We could think of adding tests for the single variable files option there if we want to test it with real files.
Can we please call the test file `test_cli.py` and test the new option without using a real files?

src/um2nc/cli.py
```diff
@@ -175,6 +176,12 @@ def __call__(self, parser, namespace, value, option_string=None):
     default=False,
     help="Delete fields files upon successful conversion"
 )
+driver_args.add_argument(
+    "--single-field-files",
```
atteggiani: I think `single-field-files` doesn't necessarily capture the outcome of this option, which is to separate the file into multiple variables, one variable per file (not necessarily one field, since a variable can have multiple fields if they are multi-time or multi-level).

We could maybe rename this option (and add more options) to something like:
```suggestion
    "--separate-variables", "--single-variables", "--one-file-per-variable",
```

atteggiani (changes requested):
Thanks @joshuatorrance!

Left a few comments as well :)

src/um2nc/um2netcdf.py
blimlim: > Also, why no test is failing?

This appears to be a big gap in the code coverage

test/test_um2netcdf.py
joshuatorrance: `process` now returns the delayed_writes list. I don't fully understand the test but I don't think the now-delayed writes matter but I wanted to highlight that process isn't writing files anymore.

test/test_single_field.py
joshuatorrance: I haven't been able to think of a meaningful test for the single-field portion of `process` without using a real file. The real file has been compressed to 100kB and only takes a minute or so to run so it's fairly lightweight.

Seems better to have more lightweight tests running in CI rather than having to run the integrations tests...

I also don't know how to run the integration tests yet either.

Adding more tests to the integration tests is a good idea though. In theory we could test with more realistic files, this one has had the fields zeroed, and perhaps more thorough checking.

src/um2nc/cli.py
joshuatorrance: I feel like `variables` in the netCDF context is more ambiguous. For example, in netCDF files `lat` and `lon` are typically `dimensions`, `coordinates`, and `variables`.

As far as I know, from a purely netCDF perspective, `field` doesn't have any innate meaning.

And to my mind, one can have multi-dimensional fields. E.g. 4D temperature is a field in t, lat, lon, height.

test/test_um2netcdf.py
joshuatorrance: This is no longer relevant since we're no longer using delayed dask writes.

src/um2nc/um2netcdf.py
joshuatorrance: I've added tests for `um2nc convert infile outfile` (currently failing as expected).

Next I'll moved the `--single-field-files` args to the common args. I think there's another bug related to strings/Paths hiding in there too.

joshuatorrance: This is done now.

Looks like there may be a separate issue as `um2nc convert` is creating a different number of files than `um2nc driver esm1p5/6` with `--single-field-files`.

atteggiani: `um2nc convert ...` only converts one file at a time whereas `um2nc driver ...` converts as many files as present in the output directory. 

Do you mean when the number of files in the output directory is only 1, converting that output file using `um2nc convert ...` or `um2nc driver ...` produces a different number of files?

test/test_single_field.py
atteggiani: I agree focusing on good unit-tests suite makes sense (and potentially also updating the integration tests to be using pytest as well - see #196 ), and I think the unit tests should not use real-data.

One way for testing this could be:
- separate the function that produces the output file(s) (which currently is embedded in the `process` function) by creating a new testable function (e.g. `save_output_file`)
- test the `save_output_file` function
- with a mock of the `save_output_file`, test that when the multi-file option is active, the mocked function gets called as many times as the number of files expected to produce
- in a similar fashion, test that when the multi-file option is NOT active, the mocked function gets called only once

This is only an example design for a possible test for this option that doesn't require real data

src/um2nc/cli.py
atteggiani: I get your point, and `variables` might not be a good choice for the reasons you outlined. 

However, from the UM perspective, fields files are the name of its native outputs (that `um2nc` converts), making `single-field-files` ambiguous and not very clear to me. 

Is there any other term we could use that doesn't include neither `field(s)` not `variable(s)`?

What about `--multiple-files` or `--multiple-output-files`?

joshuatorrance: Good point!

The fields in fieldsfile is referring to the entities we're trying to separate but "field-files" in `--single-field-files` is rather similar to "fieldsfile"...

I think "field" may be okay if we can avoid having "file" to avoid confusion with "fieldsfile".

`multiple` isn't confusing but doesn't describe what is happening as explicitly.

`--one-nc-per-field`? `--per-field-output`?

atteggiani: > The fields in fieldsfile is referring to the entities we're trying to separate

Not exactly unfortunately.
The problem is that a "field" in a fieldsfile is always 2D (lat x lon), and 3D or 4D output would have multiple fields (from a fieldsfile point of view) even if they are constituted by one single variable (i.e., only one STASH code).
In that sense the `per-field` is still confusing. 
Since it's a tool that involves UM files and netCDF, I would avoid both `field` and `variable` in the option name for the reasons mentioned above.

atteggiani: Maybe we could use `STASH`? Like `--one-nc-per-stash` or `--one-nc-per-stash-variable`?

joshuatorrance: > Not exactly unfortunately.

Ah I didn't realise. I made some incorrect assumptions!

> Maybe we could use `STASH`? Like `--one-nc-per-stash` or `--one-nc-per-stash-variable`?

Good idea, let's go with `--one-nc-per-stash` then for the sake brevity

src/um2nc/cli.py
```diff
@@ -123,6 +123,13 @@ def __call__(self, parser, namespace, value, option_string=None):
     action="store_true",
     help="Write output using simple variable names of format 'fld_s<section number>i<item number>'."
 )
+common_args.add_argument(
+    "--one-nc-per-stash",
+    action="store_true",
+    help="Create a separate netCDF file for eash STASH code. The name for each "
```
atteggiani: Minor typo
```suggestion
    help="Create a separate netCDF file for each STASH code. The name for each "
```

src/um2nc/cli.py
blimlim: Not to complicate things even further, it is possible to have multiple entries for the same stash code, i.e. it is possible to save the same stash code could be saved at different vertical levels or with different time processing into the same file. (we don't do this by default though)

atteggiani: That's a good point! 

How does it work for the conversion when there are multiple separate STASH codes in the file? How would they be differentiated and what's the output naming going to be? 

If currently we're only prepending the STASH variable name to the output (e.g. `air_temperature_output_name.nc`), if multiple separate STASH code exist in the input file, we would overwrite some of the outputs for the same STASH.

atteggiani: Would `--one-nc-per-stash-variable` be clearer? As even with the same STASH code, they would be different STASH variables 😕

blimlim: > How does it work for the conversion when there are multiple separate STASH codes in the file? How would they be differentiated and what's the output naming going to be?

Good question, the approach added here is to add an increment to the variable name and the output name: https://github.com/ACCESS-NRI/um2nc-standalone/blob/bb2fbddf4e6a3d52c1002603e59213bbfcaed2d6/src/um2nc/um2netcdf.py#L552-L557

This essentially matches how iris handles writing the same stash code multiple times to a single netcdf: https://github.com/SciTools/iris/blob/3f1aa5d394e087ccb3d08bbf3f1ff78395c0685d/lib/iris/fileformats/netcdf/saver.py#L2391-L2393

blimlim: `--one-nc-per-stash-variable`  this sounds ok to me!

test/test_single_field.py
joshuatorrance: @atteggiani

I've made some improvements to the structure of saving the cubes (e.g. see [_write_cube](https://github.com/ACCESS-NRI/um2nc-standalone/blob/592f6051aa6bae7ec467f7e713ab693a84fecc81/src/um2nc/um2netcdf.py#L523))

I've replaced the test with one that mocks the output file writing. It still uses the "real" input fieldsfile though, attempting to mock something as complex as a ESM1.6 fieldsfile in a meaningful way would probably require dramatic refactoring of `um2nc` or complicated and fragile tests. It's not a unit test but is it a simple way to test if ESM1.6 files are being handled correctly.

atteggiani: Looks good, thank you!

atteggiani:
Just a note for future PRs:
I'm noticing `main` has been merged to this PR head several times creating merge commits.
Instead of merging, it is preferrable to `git rebase ...` this PR head onto main so as to get a linear history instead of merge commits.

For that purpose, @joshuatorrance could you please rebase to `main` and resolve the conflicts? 
I think it will be better to do that before getting the last review.

src/um2nc/um2netcdf.py
```diff
@@ -503,21 +542,45 @@ def process(infile, outfile, args):
...
-            # c.data = None # Requires iris >= 3.12
-            c.data = dask.array.zeros(c.data.shape)
+            field_name = c.var_name
+
+            # Check for field name collisions
+            # Check here rather than at file output so that old files can be overwritten
+            if field_name in field_name_list:
+                new_field_name = field_name
+                while new_field_name in field_name_list:
+                    new_field_name = increment_name(field_name)
```
blimlim: Does it need to be:
```suggestion
                    new_field_name = increment_name(new_field_name)
```

blimlim (commented):
Thanks @joshuatorrance, just noticed one small thing but otherwise this is looking good to me. @atteggiani, did you have any further feedback on this one?

src/um2nc/um2netcdf.py
joshuatorrance: Nice catch, thanks! This causes an infinite loop if there's a second name collision.

Fixed with tests in https://github.com/ACCESS-NRI/um2nc-standalone/pull/234/commits/8be08245ff43f6113f969c9608e6633fdff4d5d7

test/test_single_field.py
```diff
@@ -0,0 +1,119 @@
+import os
+from pathlib import Path
+import pytest
+import shlex
+import subprocess
```
blimlim:
```suggestion
```

test/test_single_field.py
```diff
@@ -0,0 +1,119 @@
...
+
+            var_num = varname.split('_')[-1]
+
+            # var_X_file.nc where X is an integer gets sorted before var_file.nc
+            if i==len(filepaths_list) - 1:
+                # In this case file should be var_file.nc so var_num is just var
+                assert var_num == varname
+            else:
+                # Since the un-numbered var_file.nc is sorted last need to add 1
+                assert var_num == str(i+1)
```
blimlim: Could this be slightly simplified with something like:

```python
expected_paths = {f"{variable_name}_{base_filename}"} 
if n>1:
    expected_paths = expected_paths |  {f"{variable_name}_{i}_{base_filename}" for i in np.arange(1, n)}

assert filepaths_set == expected_filepaths
```

test/test_single_field.py
```diff
@@ -0,0 +1,119 @@
+import os
+from pathlib import Path
+import pytest
+import shlex
+import subprocess
+import tarfile
+from unittest import mock
+import um2nc
+from um2nc.cli import run_command
```
blimlim:
```suggestion
from pathlib import Path
import pytest
import tarfile
from unittest import mock
import um2nc
from um2nc.cli import run_command
```
I think these ones can be removed!

blimlim (commented):
Thanks @joshuatorrance, just two tiny comments on the last look through

test/test_single_field.py
joshuatorrance: https://github.com/ACCESS-NRI/um2nc-standalone/pull/234/commits/77b46c3a03dfd3f8b4db5b3d74b4ed4219b28614

blimlim (approved):
Thanks @joshuatorrance, looks good from my end. @atteggiani, just pinging you in case you had any further comments :)

atteggiani (approved):

PR #237, opened 2026-04-27
Support older iris versions

MartinDix:
See https://github.com/ACCESS-NRI/um2nc-standalone/issues/225#issuecomment-4330824243

Now works with older iris versions


atteggiani (approved):
Looks good, thanks @MartinDix

PR #241, opened 2026-05-21
Adopt ESM1.6's new filenames

joshuatorrance:
## Overview
It has been decided (see [Zulip topic](https://access-nri.zulipchat.com/#narrow/channel/470334-access-esm-release/topic/ESM1.2E6.20Filenames/with/593658029)) that ESM1.6's output files (ocean, ice and atmosphere files) will follow the [naming scheme for OM3](https://access-om3-configs.access-hive.org.au/configurations/Ocean_diagnostics/).

`um2nc` needs to use the new filename pattern for single-variable output files for the ESM1.6 driver.
For ESM1.6 we have:
`<model>.<component>.<dimension>.<field>.<frequency>.<time_cell_method>.<datestamp>.nc`
- `<model>`: `access-esm1p6`
- `<component>`: `umX.Y` where `X.Y` is the UM version used
- `<dimension>`: `2D` or `3D`, i.e. the number of dimensions for the field, excluding time
- `<field>`: The variable/field name
- `<frequency>`: The frequency of the data
- `<time_cell_method>`: Function applied along the time axis, e.g. `mean`, `max` (optional)
- `<datestamp>`: Date for the data (optional)

The optional portions are omitted where they are not applicable, e.g. variables without time aggregation, or `fx` files with no `time` component.

## Implementation
Certain portions of the filename need information from a couple of places:
- From the input filename: `frequency`
- From the source cube : `um_version`, `dimension`, `field`, `frequency`, `time_cell_method`, `datestamp`

This means that the output file name cannot be fully defined until the iris cube for the data is available, i.e. in the `process` function of `um2netcdf.py`.

### DelayedCubePath

Prior to these changes the output paths were defined well before `process` was called in the Driver's `get_output_path` method. For single variable output files, the variable name crudely prepended to the output filename during process - things were implemented that way as a placeholder for a change such as this.

To allow the output file name to be determined later get_output_path now returns a `Path` (for multi-variable output files) or a `DelayedCubePath` (for single-variable output files) which has a `resolve_cube` method to allow `process` to resolve the remaining unknowns in the filename without having to know anything about which scenario called `process`.
- For ESM1.6 a subclass, `Esm1p6DelayedCubePath`, contains the functionality to resolved the filename as described above
- For non-ESM1.6 workflows `DelayedCubePath` replicates the previous behaviour of just prepending the output filename with the variable name.

### Changes
This PR messes with some of the nuts and bolts, I've attempted to isolate changes into separate commits.
- The custom Exceptions needed to be moved to another file to avoid circular dependencies
- ~~`get_ff_date` was changed to use a datetime instead of a tuple to support refactoring I later decided not to do but a datetime seems neater so I left that change in~~ reverted
- Added `DelayedCubePath` which allows abstraction of not-fully-resolved paths that need info from Iris Cubes

### Still to do
- [x] Improve handling of instance attributes in Esm1p6DelayedCubePath
- [x] Unit tests for DelayedCubePaths
- [ ] Fix use of `unpack_fieldsfile` in `test_esm1p6_filenames.py` - should be shared with `test_single_field.py` but I was struggling to import it.


src/um2nc/drivers/common.py
```diff
@@ -142,4 +162,4 @@ def get_ff_date(fields_file_path):
     header = mule.FixedLengthHeader.from_file(
                                             str(fields_file_path))
 
-    return header.t2_year, header.t2_month, header.t2_day
+    return datetime(header.t2_year, header.t2_month, header.t2_day)
```
atteggiani: The `year`, `month` and `day` outputs from this function are used separately (i.e., the `datetime` object is never used as it is, but it's always used as `datetime.year`, `datetime.month`, etc.). 
Therefore, I would keep this as it is. The datetime object only adds overhead that I don't think is useful in our case.

src/um2nc/drivers/esm1p5.py
```diff
@@ -135,5 +135,5 @@ def get_output_path(self, input_path):
             )
             suffix = ""
 
-        year, month, _ = get_ff_date(input_path)
-        return self.output_dir / f"{stem}-{year:04d}{month:02d}{suffix}.nc"
+        dt = get_ff_date(input_path)
+        return DelayedCubePath(self.output_dir / f"{stem}-{dt.year:04d}{dt.month:02d}{suffix}.nc")
```
atteggiani: I would revert this (see my datetime-related [comment](https://github.com/ACCESS-NRI/um2nc-standalone/pull/241/changes#r3296336425))

atteggiani (commented):
Also some small initial comments

src/um2nc/drivers/common.py
joshuatorrance: I figured we were returning a custom data object, i.e. a tuple of (year, month, day), and using a standard python data object would be an improvement.

No a big deal though, I abandoned the changes that I originally tweaked this for. I've reverted the change. https://github.com/ACCESS-NRI/um2nc-standalone/pull/241/commits/fb0116a4183b7f79cbe2bc8d1eaf66715ee6a2be

src/um2nc/drivers/esm1p6.py
```diff
@@ -18,8 +22,97 @@
...
+            if 'time' in cell_method.coord_names:
+                method = f".{cell_method.method}"
+                break
+        else:
+            method = ""
+        return method
+
+    def _get_freq(self):
+        # Determine the freq from the input filename
+        if match := re.match(r"\w+\.p(\w)", self.input_filename):
```
joshuatorrance: @blimlim - Is this regex going to be a reliable way to get the freq character out of the filename to use with `ESM1P6_UNIT_SUFFIXES`?

blimlim: I just tested it out, and it's technically possible to produce filenames that will fail with the above regex. E.g. `ai]hca.par2feb`.

It's fairly unlikely anyone will change the filenames though. 

It could be slightly safer to use the regex pattern from the driver: https://github.com/ACCESS-NRI/um2nc-standalone/blob/a89d88f4ab79d3b1446ed3db5cfc01c09e8ce959/src/um2nc/drivers/esm1p5.py#L68-L71

as this reads some of the model control files to check how the output names are constructed.

src/um2nc/drivers/common.py
```diff
@@ -106,6 +113,28 @@ def run_conversion(self, delete_ff, process_args):
                     os.remove(input_path)
 
 
+class DelayedCubePath:
```
blimlim: I think it would be good to define `DelayedCubePath` under `um2nc/common.py` rather than `um2nc/drivers/common.py`, as it's used both in the drivers and the command line `convert` command.

src/um2nc/drivers/esm1p5.py
```diff
@@ -70,7 +70,7 @@ def input_name_pattern(self):
         # regex pattern for matching input files
         return re.compile(rf"^(?P<stem>{self.runid}a.p(?P<unit>[a-z]))[a-z0-9]+$")
 
-    def convert(self, input_path, output_path, process_args):
+    def convert(self, input_path, output_path: DelayedCubePath, process_args):
```
blimlim:
```suggestion
    def convert(self, input_path, output_path: Path | DelayedCubePath, process_args):
```

src/um2nc/drivers/esm1p5.py
blimlim, line 124: This early return will also need to be handled for the single variable case.

src/um2nc/drivers/esm1p6.py
```diff
@@ -18,8 +22,97 @@
...
+            unit_key = match[1]
+
+            if unit_key in ESM1P6_UNIT_SUFFIXES:
+                return ESM1P6_UNIT_SUFFIXES[unit_key]
+
+        raise ValueError(f"Unable to deduce frequency from filename while building output filename for {self.input_filename}")
```
blimlim: It's possible to add new output streams from the model which use different unit keys.

When the unit key isn't recognize, I think it would be good to produce a warning and set the frequency to `unknown`.

src/um2nc/drivers/esm1p6.py
```diff
@@ -18,8 +22,97 @@
...
+        # e.g. for datestamp 1234-01-01 00:00:00
+        #   For file containing a whole year - 1234
+        #   For file containing a month - 1234-01
+        #   For file containing a day - 1234-01-01
+        if re.match(r'\d+(yr|dec)', self.output_file_freq):
+            fmt = '%4Y'
+        elif re.match(r'\d+mon', self.output_file_freq):
+            fmt = '%4Y-%m'
+        else:
+            fmt = '%4Y-%m-%d'
```
blimlim: As a 1 year output file frequency is hardcoded here:
https://github.com/ACCESS-NRI/um2nc-standalone/blob/b1ff4b3f4cf9e1710e78bf16fdf2e70527d2569f/src/um2nc/drivers/esm1p6.py#L44

Is the handling of the other options worth keeping or not?

test/drivers/test_common.py
```diff
@@ -301,3 +303,49 @@ def test_input_output_mapping_no_duplicates(input_output, monkeypatch):
...
+@pytest.mark.parametrize(
+    "p",
+    [
+        "input_path/file.ext",
+        Path("input_path/file.ext"),
+    ]
+)
+def test_DelayedCubePath__init__(p):
+    delayed_path = drivers_common.DelayedCubePath(p)
+    delayed_path.output_path == Path(p)
```
blimlim:
```suggestion
    assert delayed_path.output_path == Path(p)
```

test/test_um2netcdf.py
```diff
@@ -379,7 +381,7 @@ def test_ancillary_files_no_support():
         mload.return_value = af
 
         with pytest.raises(NotImplementedError):
-            um2nc.process("fake_infile", "fake_outfile", args=None)
+            um2nc.process("fake_infile", DelayedCubePath("fake_outfile"), args=None)
```
blimlim:
```suggestion
            um2nc.process("fake_infile", "fake_outfile", args=None)
```

blimlim (commented):
Thanks @joshuatorrance, I think this is looking good. I'm still thinking a little about the tests but wanted to put up some small comments in the meantime.

src/um2nc/drivers/esm1p6.py
joshuatorrance: Thanks for pointing that out!

Now using `input_name_pattern` from the driver - https://github.com/ACCESS-NRI/um2nc-standalone/pull/241/commits/c43c6f7b9254f3873b1352225414505f694db9fd

src/um2nc/drivers/common.py
joshuatorrance: https://github.com/ACCESS-NRI/um2nc-standalone/pull/241/commits/4ee8b9db864bd5e90d54ed12d2fbee05ac4e116a

src/um2nc/drivers/esm1p5.py
joshuatorrance: Good point!

I've been very inconsistent with using type hints though (I think this is the only one left) so I've just removed it and updated the docstring instead.

https://github.com/ACCESS-NRI/um2nc-standalone/pull/241/changes/5d72018d1a4dfc008d7489852527aa93c9ccef84

test/drivers/test_common.py
joshuatorrance: Can't just commit the suggestion since I've moved that test.

Fixed in https://github.com/ACCESS-NRI/um2nc-standalone/pull/241/commits/71f8f2e0ac39f154ae71dc4ef27d71ff418e327d

src/um2nc/drivers/esm1p6.py
joshuatorrance: Good point. I was thinking about future proofing but this code is specific to ESM1.6 and not particularly complicated.

I've removed this in https://github.com/ACCESS-NRI/um2nc-standalone/pull/241/commits/4f46f0ae1a7207909c2b5aae3722abb020abe3bd

src/um2nc/drivers/esm1p5.py
joshuatorrance: Currently for the ESM1.6 single-variable case a ValueError is raised if the regex fails when resolving the path.
https://github.com/ACCESS-NRI/um2nc-standalone/blob/c50b9f64e546e32cf234524e2808580cd8af5bf3/src/um2nc/drivers/esm1p6.py#L87

I've added a test case for this in https://github.com/ACCESS-NRI/um2nc-standalone/pull/241/changes/c50b9f64e546e32cf234524e2808580cd8af5bf3

Would we rather return `unknown` or similar for the freq portion of the filename?

joshuatorrance: Using `unknown` is exactly what you suggested in the next comment!

I've used `unknown_freq` so it obvious what portion is unknown without knowing filename pattern
https://github.com/ACCESS-NRI/um2nc-standalone/pull/241/commits/58ba9de1d077e5e28192199e471ad003f8702ff5

src/um2nc/drivers/esm1p6.py
joshuatorrance: I've used `unknown_freq` so it obvious what portion is unknown without knowing filename pattern
https://github.com/ACCESS-NRI/um2nc-standalone/pull/241/commits/58ba9de1d077e5e28192199e471ad003f8702ff5

src/um2nc/common.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class StrictWarning(UserWarning):
+    """
+    Warnings which should be promoted to errors when the strict flag is True.
+    """
+    pass
+
+
+class DelayedCubePath:
```
atteggiani, line 27: Could this be made to inherit from `Path`? That way we wouldn't really need to differentiate the DelayedCubePath from Path, and we could return the `cli::run_command()` function to always call `outfile = DelayedCubePath(args.outfile)`.
In the same fashion, with the `get_output_path` method of the ESM1.5 drivers we could always return a DelayedCubePath.

Sorry if you explained this already in our chat, I can't remember the reason.

test/drivers/test_esm1p6_filenames.py
```diff
@@ -0,0 +1,425 @@
...
+import re
+import tarfile
+import iris
+from iris.cube import Cube
+
+from um2nc.drivers.esm1p6 import Esm1p6Driver, Esm1p6DelayedCubePath
+from test_esm1p6 import mock_atmosphere_dir
+
+
+# FIXME: This should be imported from test_single_field not duplicated
```
atteggiani: Can we put this in the [`conftest.py`](https://docs.pytest.org/en/stable/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files)?

test/test_common.py
```diff
@@ -0,0 +1,178 @@
...
+def test_DelayedCubePath_resolve_cube_multiple_paths(cleanup_DelayedCubePath, output_path_list, var_list):
+    """
+    DelayedCubePath checks for repeated output paths, including between multiple
+    input/output pairs
+    """
+    paths = []
+    for output_path in output_path_list:
+        output_path = Path(output_path)
+
+        # Iterate over the ouput paths, simuluating multiple input/output pairs
```
atteggiani:
```suggestion
        # Iterate over the ouput paths, simulating multiple input/output pairs
```

atteggiani (commented):
Thanks @joshuatorrance for updating this.

I haven't looked at the tests too carefully, but overall this looks good.

I left a few minor comments.

src/um2nc/common.py
joshuatorrance: Previous [iterations](https://github.com/ACCESS-NRI/um2nc-standalone/tree/5e7bd42895837aa2b06f8e9e169d36d1d433f4bd/src/um2nc) did inherit from Path but
- `resolve_cube` was always being called before using the path at that point, so there was no real value in inheriting Path's features
- `Path`s are a little difficult to deal with (seemed to need to extend `PosixPath` rather than `Path`, their use of `__init__` & `__new__` is a bit complicated so extending __init__ wasn't working as I'd wanted) and it would have taken me some time to understand how to extend them properly.

So all in all the positives of having a neat use of `Path`/`DelayedCubePath` in our code didn't outweigh the negatives of having to spend time figuring out how to extend `Path` properly or the risk of not doing it correctly.

test/drivers/test_esm1p6_filenames.py
joshuatorrance: Perfect! Thanks!

https://github.com/ACCESS-NRI/um2nc-standalone/pull/241/commits/62ed0246b90c0c656f13108b1a88d6e0f39ac9e1

src/um2nc/common.py
atteggiani: Oh alright, I was just thinking not to have an `__init__` or `__new__` override with it because based on how you're currently defining it, I don't think it would be necessary anymore: 
currently calling `DelayedCubePath(<path>).output_path()` returns `<path>` as a `Path` instance.

Making `DelayedCubePath` inherit from `Path` (or `PosixPath` if you need to be specific) without overriding the `init` or `new` methods, would make `DelayedCubePath(<path>)` still return `<path>` as a `Path` instance. 

So you could still use all the overridden methods if needed, but you could also treat it as a "normal" `Path` in the case where the `DelayedCubePath` functionality is not really needed.


In any case, that's not an issue, I'm fine with keeping it the way you implemented it.

test/drivers/test_common.py
```diff
@@ -5,9 +5,11 @@
 from types import SimpleNamespace
 from unittest import mock
 
+from iris.cube import Cube
+
```
blimlim:
```suggestion

```

src/um2nc/drivers/common.py
```diff
@@ -8,12 +8,14 @@
 import re
 import warnings
 
+from iris.cube import Cube
```
blimlim:
```suggestion
```

src/um2nc/drivers/esm1p6.py
```diff
@@ -5,6 +5,9 @@
 directories
 """
 
+import iris
```
blimlim:
```suggestion
```

src/um2nc/drivers/esm1p5.py
blimlim:
```suggestion
        output_path: Path or DelayedCubePath for writing output netCDF.
```
not 100% sure that this suggestion is necessary

src/um2nc/drivers/esm1p6.py
```diff
@@ -18,8 +21,89 @@
...
+        """
+        Return the output path for a given input path.
+
+        Parameters:
+        -----------
+        input_path: Path to input fields file.
+
+        Returns:
+        --------
+        output_path: Path for writing output netCDF.
```
blimlim:
```suggestion
        output_path: Path or DelayedCubePath for writing output netCDF.
```

test/test_common.py
```diff
@@ -0,0 +1,178 @@
...
+        (["a", "a", "b"], ["a", "a_1", "b"]),
+        (["a", "a", "a"], ["a", "a_1", "a_2"]),
+        (["a", "a", "b", "b"], ["a", "a_1", "b", "b_1"]),
+    ]
+)
+def test_DelayedCubePath_check_filename_collisions(name_list, expected_final_name_list):
+    final_name_list = []
+
+    # Need to clear the name list for each set of calls
+    DelayedCubePath.clear_filename_list()
```
blimlim: Could the `cleanup_DelayedCubePath` fixture be used instead of manually calling the cleanup here?

test/test_common.py
```diff
@@ -0,0 +1,178 @@
...
+        ("filename", "."),
+    ]
+)
+def test_DelayedCubePath__get_output_dir(path, expected_output_dir):
+    delayed_path = DelayedCubePath(path)
+
+    assert str(delayed_path._get_output_dir()) == expected_output_dir
+
+
+def test_DelayedCubePath_resolve_cube():
```
blimlim: I think it will be good to request the `cleanup_DelayedCubePath` fixture for this test to ensure the cleanup

test/drivers/test_esm1p6_filenames.py
```diff
@@ -0,0 +1,402 @@
...
+        # Set the var name in the cube here, it's usually set elsewhere in um2nc
+        cube.var_name = "varname"
+
+        # Resolve the filename
+        if isinstance(expected_filename_single, str):
+            resolved_path_single_field = output_path.resolve_cube(cube_list[0])
+            assert resolved_path_single_field.name == expected_filename_single
+        else:
+            with pytest.raises(expected_filename_single):
+                output_path.resolve_cube(cube_list[0])
```
blimlim: Just wanted to double check what this `if` block is doing. It looks like in all cases `expected_filename_single` will be a string

test/drivers/test_esm1p6_filenames.py
```diff
@@ -0,0 +1,402 @@
...
+            "this_file_doesnt_match_regex",
+            "access-esm1p6.um7p3.3d.varname.unknown_freq.mean.0001.nc",
+            "this_file_doesnt_match_regex.nc"
+        ),
+    ],
+)
+@pytest.mark.parametrize("one_nc", [True, False])
+@pytest.mark.filterwarnings("ignore:Input filename this_file_doesnt_match_regex does not match pattern")
+def test_esm1p6_filenames(unpack_fieldsfile, input_filename,
+    expected_filename_single, expected_filename_multi, one_nc):
```
blimlim, line 34: Should we use the `cleanup_DelayedCubePath` fixture in this one just for safety?

test/drivers/test_esm1p6_filenames.py
```diff
@@ -0,0 +1,402 @@
...
+
+    units = cf_units.Unit("days since 0001-01-01", calendar="proleptic_gregorian")
+    time = iris.coords.DimCoord(
+        points=[n_days],
+        var_name="time",
+        units=units,
+    )
+    cube.add_dim_coord(time, data_dim=0)
+
+    # Call resolve_cube
```
blimlim:
```suggestion
    # Call _build_filename
```

test/drivers/test_esm1p6_filenames.py
```diff
@@ -0,0 +1,402 @@
...
+    filename = delayed_path._build_filename(cube)
+
+    assert isinstance(filename, str)
+    assert filename == "access-esm1p6.um7p3.0d.var.1mon.2026.nc"
+
+
+def test_resolve_cube(mock_atmosphere_dir):
```
blimlim: I think this test could also use a `cleanup_DelayedCubePath` fixture

test/drivers/test_esm1p6_filenames.py
```diff
@@ -0,0 +1,402 @@
...
+        (["parentdir"], ["aiihca.pa01apr"], ["var", "var"]),
+        # Test duplicate output dirs
+        (["parentdir", "parentdir"], ["aiihca.pa01apr", "aiihca.pa01apr"], ["var"]),
+        # Test duplicate filenames
+        (["parentdir"], ["aiihca.pa01apr", "aiihca.pa01apr"], ["var"]),
+        # Test duplicate combo
+        (["parentdir", "parentdir"], ["aiihca.pa01apr", "aiihca.pa01apr"], ["var", "var"]),
+    ]
+)
+def test_resolve_cube_multiple_paths(mock_atmosphere_dir, output_dir_list, filename_list, var_list):
```
blimlim: Should this one also use the `cleanup_DelayedCubePath` to clean up the file list?

blimlim (commented):
Thanks @joshuatorrance, I think this is looking great and the tests look thorough. I've just added a few very small comments!

test/test_common.py
joshuatorrance: https://github.com/ACCESS-NRI/um2nc-standalone/pull/241/changes/ce5a25c28259cc12c8206fd04321776cc8429574

test/test_common.py
joshuatorrance: https://github.com/ACCESS-NRI/um2nc-standalone/pull/241/changes/ce5a25c28259cc12c8206fd04321776cc8429574

test/drivers/test_esm1p6_filenames.py
joshuatorrance: I must have had an iteration of this test where some cases expected failures but ended up removing those cases.

I've gotten rid of the check in https://github.com/ACCESS-NRI/um2nc-standalone/pull/241/changes/5be37dc884b6dfc32c2d05140d7e3533cfcd0f14

test/drivers/test_esm1p6_filenames.py
joshuatorrance: This one doesn't need it since it uses a real file unpacked to a `tmp_path` so each test case will have a unique path.

I've added it anyway though as a precaution!

https://github.com/ACCESS-NRI/um2nc-standalone/pull/241/changes/6fa9d87719dcd7cde087597f3824efe2c9ee6000

test/drivers/test_esm1p6_filenames.py
joshuatorrance: https://github.com/ACCESS-NRI/um2nc-standalone/pull/241/changes/6fa9d87719dcd7cde087597f3824efe2c9ee6000

test/drivers/test_esm1p6_filenames.py
joshuatorrance: https://github.com/ACCESS-NRI/um2nc-standalone/pull/241/changes/6fa9d87719dcd7cde087597f3824efe2c9ee6000

blimlim (dismissed):
Thanks @joshuatorrance, this looks great. One point we should consider is the instantaneous cell methods for each component of ESM1.6, as discussed here: https://github.com/ACCESS-NRI/esm1.6-scripts/pull/6#discussion_r3410679664

If you prefer to separate that discussion out into a separate PR, that's ok with me too!

blimlim (approved):
Thanks @joshuatorrance for all your work on this, this is looking good to me!

